PR #113, opened 2021-07-09
Set up Storybook Documentation

hreineck:
Made inital setup for writing Storybook docs and
wrote up docs for components related to OSCAL Controls.

Added .stories.js files to the src/stories directory for
each component. These files render the components in the
Storybook stories.

Created a markdown page at src/stories/OSCALDocs.stories.mdx
that can serve as an initial framework for documenting the
Storybook page.
This will include a short description of each control
and the required and optional props for each one.


hreineck:
To view the Storybook page, run:
`npm install
npm run storybook`

The .stories.js files in src/stories are responsible for rendering the docs.
They follow the format described in Storybook 's [documentation](https://storybook.js.org/docs/react/writing-stories/introduction).

kylelaker:
It looks like maybe this is based off an old version of `develop` which is causing some of these pipeline failures. Could you try doing:

```
git pull --rebase origin develop
git push --force-with-lease origin EGRC-275
```

That might pull the commit that fixed these issues to this branch

kylelaker:
hmm it looks like maybe that wasn't the issue... we've caused this issue to re-appear another way. interesting.

tuckerzp:
Does running tests locally work @hreineck?

hreineck:
> Does running tests locally work @hreineck?

No, i get this error that we had encountered before:
`Cannot find module '@testing-library/dom' from 'node_modules/@testing-library/user-event/dist/click.js'`

tuckerzp:
@hreineck I get the same error. I have tried reinstalling `package-lock` & the `node-modules`. From what I have seen, the changes to `package.json` wouldn't cause this and I assume the `.story` files wouldn't either.

rgauss:
Initial feedback after a quick look here:
- Overall, this is a great first start!!
- We should look to extend the existing granular test data to be able to support the import of it into stories so we have as little duplication as possible and are testing with data that then surfaces in the docs
- For components used across multiple consuming parent components, we should consider the OSCAL 'stack' from bottom up and describe the first story variant as something like `Default` or `ControlOnly` (rather than `No*`) then build with others that add modifications, constraints, etc. Simple cases at the top, increased complexity as we go down the variants (I keep using that term, blame it on the Loki series I guess)
- We should think about whether or not we want separate component stories for each component actually exposed in a file.
  - As an example, the `OSCALControlProse.js` file actually exposes two components `OSCALReplacedProseWithByComponentParameterValue` and `OSCALReplacedProseWithParameterLabel` because they share significant code, but are for different purposes.  Should those two components be separate stories?
 - Some of the examples use a story `title` with a prefix of `Components/...` instead of just the component which Storybook then renders under a Components group in the interface.  Perhaps we should investigate what's common and/or which method will lead to a better developer experience.
 - The Getting Started section should probably focus on how to build/access (npm registry) the library and how to use Storybook, rather than any repeat of the READMEs or components themselves

hreineck:
> * The Getting Started section should probably focus on how to build/access (npm registry) the library and how to use Storybook, rather than any repeat of the READMEs or components themselves

So should we keep the markdown page listing the required/optional props for each component? Or is that
information covered by the stories themselves?

tuckerzp:
I wonder if linting error is due to not committing the changes to `example/package-lock.json`?

zclarkEDC:
> I wonder if linting error is due to not committing the changes to `example/package-lock.json`?

This is correct.
If you change the `package.json` , you need to commit the updated `package-lock.json` with it.

src/stories/OSCALControlImplementation.stories.js
```diff
@@ -0,0 +1,113 @@
...
+const exampleImplStatements = {
+  description: "This is the control implementation for the system.",
+  "implemented-requirements": [
+    {
+      uuid: "implemented-requirements-1",
+      "control-id": "control-1",
+      statements: [
+        {
+          "statement-id": "control-1_smt.a",
+          uuid: "f3887a91-9ed3-425c-b305-21e4634a1c34",
```
Bronstrom, line 24: Overall, this PR looks pretty good! Although, maybe we could avoid some code duplication in a couple parts that come directly from the test mock data. For example, this code here looks like it could be imported from `OSCALControlImplementationImpReq.test.js`.

package.json
```diff
@@ -37,12 +38,24 @@
...
+      {
+        "files": [
+          "**/*.stories.*"
+        ],
+        "rules": {
+          "import/no-anonymous-default-export": "off"
+        }
+      }
     ]
   },
```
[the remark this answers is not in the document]
hreineck: I think that got put there by the Storybook init script. Seems like it works fine just removing those lines.

src/stories/OSCALBackMatter.stories.js
```diff
@@ -0,0 +1,132 @@
...
+import OSCALBackMatter from "../components/OSCALBackMatter";
+
+export default {
+  title: "Components/Back Matter",
+  component: OSCALBackMatter,
+};
+
+const Template = (args) => <OSCALBackMatter {...args} />;
+
+export const Default = Template.bind({});
```
rgauss, line 11: Can we change these names/implementations to:
* Default (includes title but no citation, description, or media type)
* With Description
* With Citation
* With Media Type
* Without Title

src/stories/OSCALBackMatter.stories.js
```diff
@@ -0,0 +1,132 @@
...
+      rlinks: [
+        {
+          href: "NIST_SP-800-53_rev4_catalog.json",
+        },
+      ],
+    },
+  ],
+};
+
+const exampleBackMatterDescription = {
```
rgauss, line 51: Rather than repeating most of the data in each example, should the extended examples just add to the default?

We could do something like:
```javascript
const exampleBackMatterDescription = () => {
  const example = exampleBackMatter;
  example.resources[0].description = "some description";
  return example;
};
```
There may be a more elegant way to represent that.

src/stories/OSCALControl.stories.js
```diff
@@ -0,0 +1,118 @@
+import React from "react";
+import OSCALControl from "../components/OSCALControl";
+
+export default {
+  title: "Components/Control Display",
```
rgauss, line 5: This should probably reflect the component name, so just `Control` rather than `Control Display`.

src/stories/OSCALControlImplementation.stories.js
rgauss: Agreed, that what I was suggesting in my comment:
> We should look to extend the existing granular test data to be able to support the import of it into stories so we have as little duplication as possible and are testing with data that then surfaces in the docs.

src/stories/OSCALControl.stories.js
```diff
@@ -0,0 +1,118 @@
...
+          description: "another constraint",
+        },
+      ],
+    },
+  ],
+};
+
+Default.args = {
+  control: exampleControl,
+  modifications: exampleModificationsConstraints,
```
rgauss, line 112: Can we have the default without constraints and add another of `WithParameterConstraints`?

src/stories/OSCALControl.stories.js
```diff
@@ -0,0 +1,118 @@
...
+export default {
+  title: "Components/Control Display",
+  component: OSCALControl,
+};
+
+const Template = (args) => <OSCALControl {...args} />;
+
+export const Default = Template.bind({});
+
+export const AddsModifications = Template.bind({});
```
rgauss, line 13: Should we call this `WithModifications`?

src/stories/OSCALControlPart.stories.js
```diff
@@ -0,0 +1,115 @@
...
+export default {
+  title: "Components/Control Part",
+  component: OSCALControlPart,
+};
+
+const Template = (args) => <OSCALControlPart {...args} />;
+
+export const Default = Template.bind({});
+
+export const ParameterConstraints = Template.bind({});
```
rgauss, line 13: Should this be `With ParameterConstraints`?

src/stories/OSCALControlPart.stories.js
```diff
@@ -0,0 +1,115 @@
...
+
+export const AddsModifications = Template.bind({});
```
rgauss, line 15: Should this be `WithModifications`?

src/stories/OSCALControlProse.stories.js
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+import { OSCALReplacedProseWithParameterLabel } from "../components/OSCALControlProse";
+
+export default {
+  title: "Components/Replaced Prose with Parameter Label",
```
rgauss, line 5: I know I suggested we should consider separate stories for each component within a file, but I think in this case it's more confusing.

I'd combine into a single `Control Prose` story with no `Default` but:
* Parameters Replaced With Labels
* Parameters Replaced With By-Component Value
* With Parameter Constraints

src/stories/OSCALMetadata.stories.js
```diff
@@ -0,0 +1,74 @@
...
+  component: OSCALMetadata,
+};
+
+const Template = (args) => <OSCALMetadata {...args} />;
+
+export const Default = Template.bind({});
+
+export const MetadataParties = Template.bind({});
+
+export const MetadataRoles = Template.bind({});
```
rgauss, line 15: I think we can combine `MetadataParties` and `MetadataRoles` into a single `WithPartiesAndRoles`.

src/stories/OSCALModification.stories.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
```
rgauss, line 1: I think this file needs to be renamed to `OSCALControlModification.stories.js`?

rgauss (commented):
I think we're still waiting on consolidating with the test data.

We'll also want to update the root README with a link to the live Storybook deployment.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,38 @@
+name: Publish Storybook
+on:
+  push:
+    branches:
+      - develop
+      # TODO: Remove this prior to merging into `develop`
+      - EGRC-275
```
kylelaker, line 7: The *last* change before merging into develop needs to be to remove this line; I am going to request changes to make sure we keep a little red icon on the PR until this is done. But I also don't want to resolve this too soon because I want to see future changes to this branch still get built/published.

kylelaker (changes requested):

src/stories/OSCALBackMatter.stories.js
hreineck: Something like this would affect all instances of `exampleBackMatter` because it's passed by reference. I'm not sure of an easy way to do deep copies here, but the back matter test data itself seems to avoid duplication pretty well already.

rgauss: Yes, that's why I mentioned just the "extended" examples, i.e. additive.  We could do deep copies for simple objects with JSON.parse and stringify though that can break down quickly with certain property types, or use a library like lodash, but I'm not sure either of those approaches would be worth it in this case.

I'm also fine with the current duplication, was just a thought.

tuckerzp:
Is there a reason that linting and tests are not running?

src/stories/OSCALBackMatter.stories.js
```diff
@@ -0,0 +1,52 @@
...
+export default {
+  title: "Components/Back Matter",
+  component: OSCALBackMatter,
+};
+
+const Template = (args) => <OSCALBackMatter {...args} />;
+
+export const Default = Template.bind({});
+
+export const BackMatterWithDescription = Template.bind({});
```
rgauss, line 21: These variants are all within the Back Matter story so we can remove the `BackMatter` prefix from all of them.

src/stories/OSCALControlImplementation.stories.js
```diff
@@ -0,0 +1,33 @@
...
+
+const Template = (args) => <OSCALControlImplementation {...args} />;
+
+export const Default = Template.bind({});
+
+export const ImplementedStatements = Template.bind({});
+
+const exampleControls = [exampleControl];
+
+ImplementedStatements.args = {
```
rgauss, line 23: For consistency with the other components, lets rename this to `WithImplementedStatements`.

src/stories/OSCALControlPart.stories.js
rgauss: Looks like this wasn't addressed?

src/stories/OSCALControlPart.stories.js
rgauss: Looks like this wasn't addressed?

src/stories/OSCALControlProseByComponent.stories.js
```diff
@@ -0,0 +1,26 @@
+// Storybook docs for Replaced Prose with By-Component Parameter Value
```
rgauss, line 1: I think we can delete this file now that we've consolidated in `OSCALControlProse.stories.js`?

src/stories/OSCALMetadata.stories.js
```diff
@@ -0,0 +1,25 @@
...
+export default {
+  title: "Components/Metadata",
+  component: OSCALMetadata,
+};
+
+const Template = (args) => <OSCALMetadata {...args} />;
+
+export const Default = Template.bind({});
+
+export const MetadataWithPartiesAndRoles = Template.bind({});
```
rgauss, line 17: We don't need the `Metadata` prefix here, just `WithPartiesAndRoles`.

src/stories/OSCALSystemImplementation.stories.js
```diff
@@ -0,0 +1,36 @@
...
+export const SystemImplementationLastModified = Template.bind({});
+
+export const SystemImplementationVersion = Template.bind({});
+
+Default.args = {
+  systemImplementation: exampleSystemImplementation,
+  parties: exampleParties,
+};
+
+SystemImplementationLastModified.args = {
```
rgauss, line 28: I think these can all be consolidated into the single `Default`.

src/test-data/ModificationsData.js
```diff
@@ -0,0 +1,62 @@
+const props = [
+  {
+    name: "Name",
```
rgauss, line 3: Can we make these values something like `Some Prop Name` and `Some Prop Value` so it's a bit more obvious in the rending of the component?

README.md
```diff
@@ -23,6 +23,10 @@ in the manner that best fits their application.
 
 The [example](./example) folder shows how those components can be assembled to create an OSCAL Viewer application.
 
+### Component Documentation
```
rgauss, line 26: The README might flow better if we:
* Add the Storybook reference into what is currently Introduction
* Move what is currently Introduction into its own Components second level section before Setting Up

Something like:
# OSCAL React Library
A library of React components...
...
Take a tour ...

## Components
The library provides OSCAL UI developers...
Those components directly visualize...
**Documentation for the React components...**
The example folder...

## Setting Up
### Building and Running Locally
...

rgauss (changes requested):

src/test-data/Urls.js
```diff
@@ -0,0 +1,13 @@
+/* The following URLs are all of the URLs used in testing and in the stories. */
```
rgauss, line 1: Is it worth pulling these into their own file?

src/test-data/OtherData.js
```diff
@@ -0,0 +1,94 @@
+import { backMatterTestData } from "./BackMatterData";
+
+const adds = [
```
rgauss, line 3: This seems specific to modifications, should it be moved there?

src/test-data/OtherData.js
```diff
@@ -0,0 +1,94 @@
+import { backMatterTestData } from "./BackMatterData";
```
rgauss, line 1: Should we rename this file from `OtherData` to something like `CommonData`?

rgauss (commented):
A few more minor suggestions, but they don't have to hold up approval.

kylelaker (changes requested):
I am still going to keep my "Request changes" until all other approvals are complete and we have made our last push to the branch; this is in regards to `on.push.branches[1]` and the need to remove that value but needing to _not_ impact the ability to see live deployments while developing.

src/test-data/Urls.js
hreineck: I think it's a bit cleaner to keep them in their own file, but if we'd rather pull it into something else that's easy enough to do.

src/test-data/OtherData.js
mikeisen1: Are you referring to the adds const?

mikeisen1: adds is only used in that file and isn't referenced anywhere else, that is why I placed it in OtherData.js (or whatever we rename that file to)

src/test-data/Urls.js
tuckerzp: Would it make more sense to have the urls that go with specific data go in that data file? The other urls could go with other common data.

For example: `backMatterTestUrl` goes into `BackMatterData.js `

src/test-data/OtherData.js
tuckerzp: For #124 having the `modify` including the `adds` portion of a profile is used in places other than just `Profile` (`SSP` & `Component Definition`. Maybe having a `ModificationData` file makes sense? I am not sure if that is exactly what you mean @rgauss but it could prevent some duplication later.

hreineck: It is related to modifications, but the only place it's used is the profile test. 
I think it'd technically be cleanest to modify the profile test to use the existing
test data in `test-data/ModificationsData` but that might be more trouble than it's worth.

src/test-data/Urls.js
zclarkEDC: Having the urls in their own file will be a nice fix if/when the urls change in the future.

zclarkEDC (changes requested):
All local tests pass
```
Test Suites: 14 passed, 14 total
Tests:       66 passed, 66 total
Snapshots:   0 total
Time:        17.925 s
Ran all test suites.
```
Doing `npm run storybook` successfully builds all modules and opens the Storybook on localhost.

Console error when opening Control Prose -> Parameters Replaced With By-Component Value
console error
This issue is most likely caused in `OSCALControlProse.stories.js` line 51
```
ByComp.storyName = "Parameters Replaced With By-Component Value";
```

Removing this line caused the same error, however as you can see:
error 2
It changed the name to what looks like the correct naming scheme.

tuckerzp:
>  We're trying to then pass that in as the `title` prop of `StyledTooltip` by assigning `description` to be equal to the returned object. This makes it undefined, and passing undefined as a prop gives the error.

When `statementByComponent` is de-structured, is `description` always undefined? 

> What should this title prop actually be? I thought of passing in something like the componentId prop as a workaround, but maybe that's not what's intended.

If `description` is not `undefined`, I would think we could still use it? Maybe if it is `undefined`, using `componentId` or some other variable would be good.  Maybe something like this?
```javascript
<StyledTooltip title={description ?? props.componentId}>
```

zclarkEDC (approved):

tuckerzp (commented):
Everything looks good!

I was still curious about what everyone thought about Ray's recommendations of:
1. Putting urls into their relevant components
2. Creating a `modification` test data file and moving the `adds` & `modify` parameters to that file.

I was just hoping to get a final opinion on what to do for those conversations then I think we should be good to merge.

src/test-data/OtherData.js
hreineck: I moved this to a more accurately named 'ProfileData.js` - hope that makes this more readable.

src/test-data/Urls.js
hreineck: I agree with @zclarkEDC - plus I think the Urls file is justified because some of the Urls are used across a couple different files.

rgauss (approved):

tuckerzp (approved):

kylelaker (approved):

PR #120, opened 2021-07-16
Revert "Allow robots to crawl the site"

kylelaker:
Reverts EasyDynamics/oscal-react-library#119

PR #119, opened 2021-07-16
Allow robots to crawl the site

kylelaker:
Has the `robots.txt` changes from #115

PR #115, opened 2021-07-15
Additional branding fixes

kylelaker:
This fixes some color and wording in the branding. ~Not 100% on the `robots.txt` change but it's a separate commit and can be removed from this PR fairly trivially.~ Robots now moved to #119

PR #121, opened 2021-07-16
Revert "Revert "Allow robots to crawl the site""

kylelaker:
Reverts EasyDynamics/oscal-react-library#120 which reverted #119 which stemmed from #115. Let's have an actual conversation around whether we're ready to be crawled before moving forward with this PR.

### DIFF
--- a/example/public/robots.txt
+++ b/example/public/robots.txt
@@ -1,3 +1,3 @@
 # https://www.robotstxt.org/robotstxt.html
 User-agent: *
-Disallow: /
+Disallow:


tuckerzp (changes requested):
I thought since no one else has started a conversation, I would get the ball rolling. I have heard a bit on web crawling before but read a few articles to familiarize myself more. 

From what I have read, it would seem like allowing robots to crawl our site would not be a bad thing. It could potentially make our page show up more for searches, which would be a positive thing for our project. We also do not have a ton of different pages to go through, so it does not seem like a slowdown would be something to worry about. If there are certain pages we do not want to be crawled, we could add more specific rules for them. It made sense before we went public with our project, but we could consider allowing bots now. 

I will request changes for now in an attempt to prevent merging. 

---
For others who also were not super familiar I added a few links below:
- https://www.cloudflare.com/learning/bots/what-is-a-web-crawler/
- https://www.checkerboard.com/seo-services/what-is-website-crawling/

Bronstrom:
I have been reading through a few articles as well for this issue. It seems like the most benefit we would receive from the addition of bots would be their help in achieving a high rank in searches - making our project easier to be found and searched for. I was curious, is there anything we would want to consider limiting crawler bots from crawling on? I think for the most part our project emphasizes good crawlability (see below for more details) and I don't think we have too much content that would cause bots to overtax the environment. 

[What is Crawlability? - Codesquad](https://codesquad.cloud/wordpress-seo-in-5-minutes-what-is-crawlability/)

tuckerzp:
> is there anything we would want to consider limiting crawler bots from crawling on?

@rgauss would know better than us, but I can not think of something obvious that needs to be limited. Even if there were some pages we wanted to limit, adding those rules are trivial and wouldn't prevent us allowing bots to crawl the other pages.

zachclark-ccpo:
We could just allow the Google robots with :
```
User-agent: Google
Disallow:

User-agent: *
Disallow: /
```

If we do want to allow any sort of crawling, we need to add certain meta tags to our index.html file
https://developers.google.com/search/docs/advanced/crawling/special-tags

We are using the default create-react-app meta tags, we need to update these to reflect OSCAL Viewer

tuckerzp:
> If we do want to allow any sort of crawling, we need to add certain meta tags to our `index.html` file

Would that be within scope of this PR? If we agree on allowing bots, I can work on adding the meta tags.

tuckerzp:
I can no longer work on this issue. @rgauss or @kylelaker does allowing robots make sense?

rgauss:
I have no issue with allowing crawling of **our** sandbox deployment of the viewer app.

However, I'm not sure we want crawling enabled by default for **others** deploying the viewer app.

As such I think we can close this particular PR and if we have a strong desire to enable crawling for the sandbox environment we can revisit that in a PR to our infrastructure/pipeline repo.

PR #114, opened 2021-07-14
Improve Profile Viewer performance & user experience

Bronstrom:
Due to lengthy rendering, the Profile Viewer takes a long period of time to show elements on the page. In order to improve the user experience, a skeleton framework has been added to display on the page while rendering the resolved controls. To improve performance, a few areas of `OSCALProfile`, `OSCALControl`, & underlying components have been optimized. Most notably, the `modifications` data sent between components have been slimmed down.

- resolves #74

PR #124, opened 2021-07-19
Update SSP and Component Definition to Display Profile Modifications

### LINKED ISSUE #75: Update SSP and Component Definition to Display Profile Modifications
As user of the OSCAL Viewer or another app consuming the OSCAL React components, I want to be able to see the profile control modifications in the SSP’s implemented requirements and Component Definition’s control implementations so that I can fully understand the compliance requirements and how an SSP or Component Definition satisfies them.

### Acceptance Criteria
* The Component Definition view displays the control modifications from imported profiles
  * The example component definition [brings in the FedRAMP High profile](https://github.com/usnistgov/oscal-content/blob/master/examples/component-definition/json/example-component.json#L46) so should display that profile's AC-2.3 constraints and additions for that particular control implementation
* The SSP view displays the control modifications from imported profiles
  * The example SSP brings in the [800-53 rev4 Moderate profile](https://github.com/usnistgov/oscal-content/blob/master/examples/ssp/json/ssp-example.json#L44) so should display that profile's AU-1 additions.
* The same acceptance criteria from #74 will apply here

### Implementation Notes
* We should be able to leverage the work/components previously created for the Profile VIew

### Child Issues
- [x] Display Modifications in SSP 
- [x] Display Modifications in Component Definition

tuckerzp:
This adds the ability to view modifications on `SSP` & `Component` viewers. `OSCALProfileUtils` is used to fetch the `modify` attribute of a given `Profile`. This also implements some of changes to displaying modifications from #114, but with the `SSP` & `Component`. Tests were also added to check that the modifications only appear when they should. 

This will close #75

### DIFF
--- a/src/components/OSCALComponentDefinitionControlImplementation.js
+++ b/src/components/OSCALComponentDefinitionControlImplementation.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import Card from "@material-ui/core/Card";
 import CardContent from "@material-ui/core/CardContent";
@@ -6,6 +6,7 @@ import Grid from "@material-ui/core/Grid";
 import Typography from "@material-ui/core/Typography";
 import { List, ListItem, ListItemText } from "@material-ui/core";
 import OSCALControlImplementationImplReq from "./OSCALControlImplementationImplReq";
+import getProfileModifications from "./oscal-utils/OSCALProfileUtils";
 
 const useStyles = makeStyles((theme) => ({
   paper: {
@@ -28,6 +29,7 @@ const useStyles = makeStyles((theme) => ({
 
 export default function OSCALComponentDefinitionControlImplementation(props) {
   const classes = useStyles();
+  const [modifications, setModifications] = useState(props.modifications);
 
   return (
     <div className={classes.paper}>
@@ -43,32 +45,46 @@ export default function OSCALComponentDefinitionControlImplementation(props) {
             </Grid>
             <Grid item xs={12}>
               <List>
-                {props.controlImplementations.map((controlImpl) => (
-                  <ListItem key={controlImpl.uuid}>
-                    <ListItemText>
-                      {controlImpl.description}
-                      <Grid item xs={12}>
-                        <List
-                          className={
-                            classes.OSCALControlImplementationImplReqList
-                          }
-                        >
-                          {controlImpl["implemented-requirements"].map(
-                            (implementedRequirement) => (
-                              <OSCALControlImplementationImplReq
-                                implementedRequirement={implementedRequirement}
-                                components={props.components}
-                                controls={props.controls}
-                                childLevel={0}
-                                key={implementedRequirement.uuid}
-                              />
-                            )
-                          )}
-                        </List>
-                      </Grid>
-                    </ListItemText>
-                  </ListItem>
-                ))}
+                {props.controlImplementations.map((controlImpl) => {
+                  // Get modifications from source if it is profile
+                  if (!modifications) {
+                    getProfileModifications(
+                      controlImpl.source,
+                      props.parentUrl,
+                      setModifications
+                    );
+                  }
+
+                  return (
+                    <ListItem key={controlImpl.uuid}>
+                      <ListItemText>
+                        {controlImpl.description}
+                        <Grid item xs={12}>
+                          <List
+                            className={
+                              classes.OSCALControlImplementationImplReqList
+                            }
+                          >
+                            {controlImpl["implemented-requirements"].map(
+                              (implementedRequirement) => (
+                                <OSCALControlImplementationImplReq
+                                  implementedRequirement={
+                                    implementedRequirement
+                                  }
+                                  components={props.components}
+                                  controls={props.controls}
+                                  childLevel={0}
+                                  key={implementedRequirement.uuid}
+                                  modifications={modifications}
+                                />
+                              )
+                            )}
+                          </List>
+                        </Grid>
+                      </ListItemText>
+                    </ListItem>
+                  );
+                })}
               </List>
             </Grid>
           </Grid>
--- a/src/components/OSCALComponentDefinitionControlImplementation.js
+++ b/src/components/OSCALComponentDefinitionControlImplementation.js
@@ -6,7 +6,7 @@ import Grid from "@material-ui/core/Grid";
 import Typography from "@material-ui/core/Typography";
 import { List, ListItem, ListItemText } from "@material-ui/core";
 import OSCALControlImplementationImplReq from "./OSCALControlImplementationImplReq";
-import getProfileModifications from "./oscal-utils/OSCALProfileUtils";
+import fetchProfileModifications from "./oscal-utils/OSCALProfileUtils";
 
 const useStyles = makeStyles((theme) => ({
   paper: {
@@ -46,9 +46,10 @@ export default function OSCALComponentDefinitionControlImplementation(props) {
             <Grid item xs={12}>
               <List>
                 {props.controlImplementations.map((controlImpl) => {
-                  // Get modifications from source if it is profile
+                  // Fetch modifications from a source
+                  // Pass a set() function to set modifications
                   if (!modifications) {
-                    getProfileModifications(
+                    fetchProfileModifications(
                       controlImpl.source,
                       props.parentUrl,
                       setModifications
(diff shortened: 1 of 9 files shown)


kkennedy26 (commented):
So far, looks good to me! @tuckerzp are the tests for control modifications being added to `OSCALComponentDefinition.test.js` and `OSCALSsp.test.js` files?

tuckerzp:
>  are the tests for control modifications being added to `OSCALComponentDefinition.test.js`

I was thinking I would add it to `OSCALComponentDefinitionControlImplementation.test.js` as it is really the Control Implementations that are displaying modifications.

hreineck (commented):
Looks good! Everything in the viewer works fine on my end.

tuckerzp:
After working with @Bronstrom on tests, we have decided the code might be issue rather than the tests. Going with that assumption, we need to work on correctly displaying modifications and having the tests pass. The `SSP` & `Component Definition` oddly display the modifications sometimes, but more often fail to display.

src/components/OSCALComponentDefinitionControlImplementation.js
```diff
@@ -43,32 +45,46 @@ export default function OSCALComponentDefinitionControlImplementation(props) {
...
-                ))}
+                {props.controlImplementations.map((controlImpl) => {
+                  // Get modifications from source if it is profile
+                  if (!modifications) {
+                    getProfileModifications(
+                      controlImpl.source,
+                      props.parentUrl,
+                      setModifications
+                    );
+                  }
```
kylelaker: Maybe this is a larger architectural thing but I think this highlights some questions that I have more largely. So I don't want to hold up this PR but I want to ask to see if there are issues we can create to better get a feel for what's going on here.

So above, we make a call to `useState(props.modifications)` that gives us `modifications` and `setModifications`, both of which we use here. But `getProfileModifications` seems to asynchronously _set_ values on the profile using the passed in `setModifications` parameter (which appears to be a function?). 

We also seem to be passing all these URLs around; I don't see where an actual object gets set/returned. And maybe I don't know the code base well enough, but I feel like like I should get a vibe for where we end up actually getting an object with values somewhere.

I think we've got a few things here:
1. I am expecting this code to look more object-oriented and for a variety of reasons, it isn't and that's on me
2. We probably could a better job naming things. Is `setModifications` setting modifications? Or something else? Both? Is `getProfileModifications` purely getting things? It feels like quite a bit of a _write_ operation.
3. We could probably do a better job of documenting things; but I am also a big believer that names are an important part of documenting so maybe this is the same as 2.

I want to be really clear that I am not complaining about any changes/additions specifically in this code; I realize these are patterns we use in a few places. I _do_ think that there's an opportunity to make things a little better (or at least different) and maybe to identify how we can create better (and more reusable?) patterns elsewhere?

src/components/OSCALControlImplementationImplReq.js
```diff
@@ -111,9 +111,15 @@ export default function OSCALControlImplementationImplReq(props) {
   if (!implReqStatements) {
     implReqStatements = [];
   }
+
+  // Error check modifications
+  const modAlters = !props.modifications ? null : props.modifications.alters;
+  const modParams = !props.modifications
+    ? null
+    : props.modifications["set-parameters"];
```
kylelaker, line 119: Is this the same as

```suggestion
  const modAlters = props.modifications?.alters || null;
  const modParams = props.modifications?.["set-parameters"] || null;
```

Which will return `props.modifications.X` _if_ that chain of things exists and if the result is `truthy`. We could drop the `|| null` if this value being an empty array (or undefined) is acceptable.

kylelaker: Can we test the falsey path of this? Where these things don't exist? I don't see a test added for that.

kylelaker (changes requested):

tuckerzp:
Adding test data for the modification tests may conflict with the goals of #113. It a is minor issue, but something we will probably want to address in one of the PRs.

kylelaker:
you could avoid conflicts and issues by following the same style for test data that we're starting to follow as of #113 or we can see who gets merged to develop first. but i'd hope it's #113 for simplicity

kkennedy26 (commented):
The display of modifications and the tests look great, but may have to hold off on the approval until issue #125 is merged to get a full review

tuckerzp:
>  hold off on the approval until issue #125 is merged to get a full review

@kkennedy26 for now this `Component` works great
https://raw.githubusercontent.com/EasyDynamics/oscal-content/5ef5b3709045ba3aab1473813c13436e886d6185/examples/component-definition/json/example-component.json

kkennedy26:
> @kkennedy26 for now this `Component` works great
> https://raw.githubusercontent.com/EasyDynamics/oscal-content/5ef5b3709045ba3aab1473813c13436e886d6185/examples/component-definition/json/example-component.json
> 
> It is exactly how it will work once #125 is resolved.

Yep everything checks out! Looks good

kkennedy26 (approved):

hreineck (approved):
All tests pass locally and modifications render on the viewer. Awesome work!

src/components/OSCALComponentDefinitionControlImplementation.test.js
```diff
@@ -130,6 +130,43 @@ const controlsData = [
   },
 ];
 
+const profileModifyTestData = {
```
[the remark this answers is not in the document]
tuckerzp: I think as @kylelaker said it makes more sense for #113 to be merged into `develop` first.

kylelaker:
I will resolve my request as soon as #113 is merged. Thanks for your work on this @tuckerzp! Great stuff!

src/components/OSCALComponentDefinitionControlImplementation.test.js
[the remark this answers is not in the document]
kylelaker: @tuckerzp now that #113 is merged, please fix this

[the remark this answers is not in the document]
tuckerzp: I am working on it now @kylelaker

tuckerzp:
@rgauss now that #113 is merged into develop, would it be within scope of this PR to add these changes to the Storybook?

kylelaker:
Now that #113 is merged, I would say the new standard in the code base is to separate the test data and handle it the way we moved to doing in that change. Is there a specific concern about that? If it's not done here, then we need to identify when it will be done.

tuckerzp:
Test data has now been moved to `test-data` and Storybook successfully shows modifications. We should be good for a final review and to merge.

src/test-data/ModificationsData.js
```diff
@@ -49,3 +49,40 @@ export const exampleModificationAltersTopLevel = [
     ],
   },
 ];
+
+export const profileModifyTestData = {
```
hreineck, line 53: This doesn't seem significantly different from the existing `exampleModifcationAlters` and `exampleModificationSetParameters` objects in this file. Is it possible to modify this (or modify the tests) to use the existing data?

hreineck (changes requested):

src/test-data/ModificationsData.js
hreineck: If this seems out of the scope of this PR though this doesn't need to hold up approval.

hreineck (approved):

src/test-data/ModificationsData.js
tuckerzp: Oh I had missed that! Good catch @hreineck. Should be good now.

mikeisen1: I'm looking at the code and `exampleModificationAltersTopLevel` is almost the exact same as `exampleModificationAlters`. Would it make more sense to just use one of them, so as to get rid of the data duplication?

tuckerzp: @mikeisen1 I believe that the difference between a `ModificationAlter` & a `ModificationAltersTopLevel` might be significant enough for the duplication. The `by-id` is different for both and I think the amount of code redundancy fixed is very minimal.

mikeisen1 (approved):
Everything works as expected, nice job.

kylelaker (approved):

rgauss (commented):
Apologies, I was not able to review this earlier.

If I'm reading this correctly, we may have a few high-level issues to address.

The resolution of the controls and their associated modifications needs to be per-profile, not per SSP or Component Definition, and Profiles might import other Profiles.

We also want to avoid subsequent fetches of any external URL we've already processed, like reading the profile again.

Those factors would lead us to look to add the modification resolution somewhere like `OSCALResolveProfileOrCatalogUrlControls` in `OSCALProfileResolver` and pass those specific modifications to each specific control implementation.

When all modifications are lumped into a 'top-level' modifications array we can't discern which modification is for which implementation.  We can see this in the Component Definition viewer where the 800-53 rev4 catalog implementation shows modifications, but catalogs don't have modifications.

Can we rework this code with that in mind?

PR #130, opened 2021-07-20
Set Up GitHub Actions for Publishing Storybook to GitHub Pages

hreineck:
Created gh-pages.yml in workflows
This sets up the Github actions to publish the Storybook docs
to Github pages on each push to develop.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,36 @@
+name: Publish Storybook
+on:
+  push:
+    branches:
+      - develop
+      - EGRC-471
+
+jobs:
+  publish:
+    name: Publish Storybook to GitHub Pages
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup NodeJS
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          cache: "npm"
+      - name: Globally update npm
+        run: npm install -g npm@latest
+      - name: Install dependencies
+        run:
+          npm install
+      - name: Build Storybook
+        run:
+          npm run build-storybook
+      - name: Deploy to Pages
+        uses: JamesIves/github-pages-deploy-action@4.1.4 
+        with:
+          branch: storybook-docs
+          folder: storybook-static
+          clean: true
+
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - develop
-      - EGRC-471
+      - EGRC-275
 
 jobs:
   publish:
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -33,4 +33,6 @@ jobs:
           folder: storybook-static
           clean: true
           single-commit: true
+          git-config-name: Easy Dynamics Automation
+          git-config-email: noreply@easydynamics.com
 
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Storybook
+on:
+  push:
+    branches:
+      - develop
+      - EGRC-275
```
kylelaker, line 6:
```suggestion
      # TODO: This line must be removed prior to merging into `develop`
      - EGRC-275
```

Can we add a comment like this just so that we don't forget?

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  publish:
+    name: Publish Storybook to GitHub Pages
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
kylelaker, line 16:
```suggestion
        with:
          fetch-depth: 0
```

We actually don't need the whole history here; sorry for suggesting we copy/paste from the other workflow.

kylelaker (changes requested):
A few small changes would be good here

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,36 @@
...
+          npm install
+      - name: Build Storybook
+        run:
+          npm run build-storybook
+      - name: Deploy to Pages
+        uses: JamesIves/github-pages-deploy-action@4.1.4 
+        with:
+          branch: storybook-docs
+          folder: storybook-static
+          clean: true
```
kylelaker, line 35:
```suggestion
          clean: true
          git-config-name: Easy Dynamics Automation
          git-config-email: noreply@easydynamics.com
```

I think we can workshop these names; however, without something like this, each commit author will be set to the value in the `github` context in GitHub Actions, which will likely be the person who pushed to `develop` (so in our case, the person who merged the PR). That's _fine_ but it may not really be what we want. So we may want to fake a "service account" type person for this. If we let it use the merger's name/email we may have some commits marked as "Unverified" since some folks on the project use PGP keys; also the person merging may not be the person who wrote the change which may slightly obscure authorship info.

I am not 100% confident in this suggestion; there's quite a few pros and cons either way so maybe don't accept this suggestion until we have some sort of conversation around it?

kylelaker, line 35:
```suggestion
          clean: true
          single-commit: true
```

I would kind of like to see us keep a flat history on this branch (since it's not our `main`/`develop` branch) and is automatically generated. The actual history is in the source files. If we were doing something like pushing to the `docs/` folder on `main` then it'd make sense to keep all history; but this is basically just plain file storage.

kylelaker (commented):
Some things to talk about

kylelaker:
And as a last thing to discuss, the Action we're using has support for pushing to a different repo so that we could create a repo like `oscal-react-library-storybook` or something like that. We could then keep this repo's GitHub Pages clean/available for a different purpose (like deploying the sample app through GHA if we ever wanted to).

.github/workflows/gh-pages.yml
hreineck: Thanks for the feedback! 
I think this is a good change; it wouldn't make sense to have the merger be the author.

kylelaker: Ya know, since we've gone ahead and set the `single-commit: true` flag, let's go ahead and roll with this; if we decide to change our mind later (to use a different name/email), it's not we'll have an invalid history we've got to clean up (since we won't have a history at all). And we have one more chance to review this in #113

kylelaker (approved):

zclarkEDC:
Excellent job, thank you.

zclarkEDC (approved):

rgauss (approved):

PR #140, opened 2021-07-27
Adjust SSP Viewer for FedRAMP template

### LINKED ISSUE #133: Minor Adjustments for FedRAMP SSP Templates
As a user viewing the [FedRAMP SSP template](https://github.com/GSA/fedramp-automation/blob/master/dist/content/templates/ssp/json/FedRAMP-SSP-OSCAL-Template.json), I want the sections to render in a visually pleasing manner so that I can more easily read and comprehend the example.

### Acceptance Criteria
* The System Characteristics properties width is increased to fill white space and not word wrap
* The System Implementation Component Title width is increased slightly to lessen word wrap
* The System Implementation Component Properties width is decreased slightly to lessen word wrap of other component fields
* The controls can be resolved even though the `import-profile` references an XML URL

### Implementation Notes
* The existing [fixJsonUrls function](https://github.com/EasyDynamics/oscal-react-library/blob/develop/src/components/oscal-utils/OSCALBackMatterUtils.js#L1) can likely be used for profile resolution

Screen_Shot_2021-07-21_at_8_15_31_AM
Screen_Shot_2021-07-21_at_8_16_44_AM

Bronstrom:
Adjustments have been made in the visual display of the [FedRAMP SSP template]( https://raw.githubusercontent.com/GSA/fedramp-automation/master/dist/content/templates/ssp/json/FedRAMP-SSP-OSCAL-Template.json), importing a profile, and displaying statements.

Visually, columns in System Characteristics and System Implementation have been adjusted in placement and size to lessen word wrap and fill more of the negative space. While making visual adjustments, the placement of other elements on screen were maintained.

After resolving the `import-profile` xml link in `OSCALSspResolver`, there was a need to rework how statements are fetched for the implementation controls to display correctly. This FedRAMP SSP template provided the `set-parameters` in the first/overarching statement, rather than each individual statement for the controls. For example, the `set-parameters` are listed purely in "ac-1_smt" versus being listed in something like "ac-1_smt.a" or "ac-1_smt.b.1" in NIST's SSPs. `OSCALControl` now tries to grab these statements (which are null if provided like NIST's statements), then passes them to underlying components. With this added feature, the SSP Viewer has no problem handling controls from both NIST and FedRAMP.

- Resolves #133 
### Testing
Use the [FedRAMP SSP template]( https://raw.githubusercontent.com/GSA/fedramp-automation/master/dist/content/templates/ssp/json/FedRAMP-SSP-OSCAL-Template.json ) when running the SSP Viewer.

### DIFF
--- a/src/components/OSCALControl.js
+++ b/src/components/OSCALControl.js
@@ -61,6 +61,21 @@ export default function OSCALControl(props) {
     );
   }
 
+  // FedRAMP SSP - need to retrieve the set-parameters from the initial implementation request from each part
+  let firstImplReqStatement = null;
+  if (props.implReqStatements) {
+    firstImplReqStatement =
+      props.implReqStatements[Object.keys(props.implReqStatements)[0]];
+  }
+  let firstByComp = null;
+  if (firstImplReqStatement && firstImplReqStatement["by-components"]) {
+    firstByComp =
+      firstImplReqStatement["by-components"][
+        Object.keys(firstImplReqStatement["by-components"])[0]
+      ];
+  }
+  const setParams = firstByComp?.["set-parameters"] || null;
+
   return (
     <Card
       className={`${classes.OSCALControl} ${classes.OSCALControlStatus} ${classes.OSCALControlChildLevel}`}
@@ -81,6 +96,7 @@ export default function OSCALControl(props) {
               control={props.control}
               parameters={props.control.params}
               implReqStatements={props.implReqStatements}
+              implReqSetParameters={setParams}
               componentId={props.componentId}
               modificationAlters={props.modificationAlters}
               modificationSetParameters={props.modificationSetParameters}
--- a/src/components/OSCALControl.js
+++ b/src/components/OSCALControl.js
@@ -61,14 +61,14 @@ export default function OSCALControl(props) {
     );
   }
 
-  // FedRAMP SSP - need to retrieve the set-parameters from the initial implementation request from each part
+  // When FedRAMP, retrieve the set-parameters from the initial implementation request statement
   let firstImplReqStatement = null;
   if (props.implReqStatements) {
     firstImplReqStatement =
       props.implReqStatements[Object.keys(props.implReqStatements)[0]];
   }
   let firstByComp = null;
-  if (firstImplReqStatement && firstImplReqStatement["by-components"]) {
+  if (firstImplReqStatement?.["by-components"]) {
     firstByComp =
       firstImplReqStatement["by-components"][
         Object.keys(firstImplReqStatement["by-components"])[0]
--- a/src/components/OSCALControl.js
+++ b/src/components/OSCALControl.js
@@ -65,20 +65,16 @@ export default function OSCALControl(props) {
   let topLevelImplReqStatement = null;
   if (props.implReqStatements) {
     // NOTE: The top level statement ends with "_smt"
-    props.implReqStatements.forEach((statement, index) => {
-      if (statement["statement-id"].endsWith("_smt")) {
-        topLevelImplReqStatement =
-          props.implReqStatements[Object.keys(props.implReqStatements)[index]];
-      }
-    });
+    topLevelImplReqStatement = props.implReqStatements.find((statement) =>
+      statement["statement-id"].endsWith("_smt")
+    );
   }
   // NOTE: The top-level by-component is the first element in the by-components array
   let topLevelByComp = null;
   if (topLevelImplReqStatement?.["by-components"]) {
-    topLevelByComp =
-      topLevelImplReqStatement["by-components"][
-        Object.keys(topLevelImplReqStatement["by-components"])[0]
-      ];
+    topLevelByComp = topLevelImplReqStatement["by-components"].find(
+      (byComp) => byComp["component-uuid"] === props.componentId
+    );
   }
   const setParams = topLevelByComp?.["set-parameters"] || null;
 
--- a/src/components/OSCALControl.js
+++ b/src/components/OSCALControl.js
@@ -5,6 +5,7 @@ import Typography from "@material-ui/core/Typography";
 import { makeStyles } from "@material-ui/core/styles";
 import OSCALControlPart from "./OSCALControlPart";
 import OSCALControlModification from "./OSCALControlModification";
+import { getImplReqSetParameters } from "./OSCALControlProse";
 
 // TODO - This is probably 800-53 specific?
 function getControlStatusCss(props) {
@@ -40,23 +41,6 @@ const useStyles = makeStyles(() => ({
   // TODO - This is probably 800-53 specific?
   OSCALControlStatus: (props) => getControlStatusCss(props),
 }));
-
-/**
- * Retrieve the set-parameters when contained in the top-level implemented requirement statement
- *
- * @param {Object} implReqStatements statements on how the control is implemented
- * @param {String} componentId uuid of the component and desired by-component
- * @returns An array of set-parameters
- */
-function getImplReqSetParameters(implReqStatements, componentId) {
-  // Get the top-level implemented requirement statement
-  let topLevelByComp = implReqStatements
-    ?.find((statement) => statement["statement-id"].endsWith("_smt"))
-    ?.["by-components"]?.find(
-      (byComp) => byComp["component-uuid"] === componentId
-    );
-  return topLevelByComp?.["set-parameters"] || null;
-}
 
 export default function OSCALControl(props) {
   if (
--- a/src/components/OSCALControl.js
+++ b/src/components/OSCALControl.js
@@ -41,6 +41,32 @@ const useStyles = makeStyles(() => ({
   OSCALControlStatus: (props) => getControlStatusCss(props),
 }));
 
+/**
+ * Retrieve the set-parameters when contained in the top-level implemented requirement statement
+ *
+ * @param {Object} implReqStatements
+ * @param {String} componentId
+ * @returns An array of set-parameters
+ */
+function getImplReqSetParameters(implReqStatements, componentId) {
+  // Get the top-level implemented requirement statement
+  let topLevelImplReqStatement = null;
+  if (implReqStatements) {
+    // NOTE: The top level statement ends with "_smt"
+    topLevelImplReqStatement = implReqStatements.find((statement) =>
+      statement["statement-id"].endsWith("_smt")
+    );
+  }
+  // NOTE: The top-level by-component has a matching uuid to the componentId
+  let topLevelByComp = null;
+  if (topLevelImplReqStatement?.["by-components"]) {
+    topLevelByComp = topLevelImplReqStatement["by-components"].find(
+      (byComp) => byComp["component-uuid"] === componentId
+    );
+  }
+  return topLevelByComp?.["set-parameters"] || null;
+}
+
 export default function OSCALControl(props) {
   if (
     !props.control ||
@@ -61,22 +87,11 @@ export default function OSCALControl(props) {
     );
   }
 
-  // Retrieve the set-parameters when contained in the top-level implemented requirement statement
-  let topLevelImplReqStatement = null;
-  if (props.implReqStatements) {
-    // NOTE: The top level statement ends with "_smt"
-    topLevelImplReqStatement = props.implReqStatements.find((statement) =>
-      statement["statement-id"].endsWith("_smt")
-    );
-  }
-  // NOTE: The top-level by-component is the first element in the by-components array
-  let topLevelByComp = null;
-  if (topLevelImplReqStatement?.["by-components"]) {
-    topLevelByComp = topLevelImplReqStatement["by-components"].find(
-      (byComp) => byComp["component-uuid"] === props.componentId
-    );
-  }
-  const setParams = topLevelByComp?.["set-parameters"] || null;
+  // Retrieve implemented requirements set-parameters
+  const setParams = getImplReqSetParameters(
+    props.implReqStatements,
+    props.componentId
+  );
 
   return (
     <Card
--- a/src/components/OSCALControl.js
+++ b/src/components/OSCALControl.js
@@ -5,7 +5,6 @@ import Typography from "@material-ui/core/Typography";
 import { makeStyles } from "@material-ui/core/styles";
 import OSCALControlPart from "./OSCALControlPart";
 import OSCALControlModification from "./OSCALControlModification";
-import { getImplReqSetParameters } from "./OSCALControlProse";
 
 // TODO - This is probably 800-53 specific?
 function getControlStatusCss(props) {
@@ -62,12 +61,6 @@ export default function OSCALControl(props) {
     );
   }
 
-  // Retrieve implemented requirements set-parameters
-  const setParams = getImplReqSetParameters(
-    props.implReqStatements,
-    props.componentId
-  );
-
   return (
     <Card
       className={`${classes.OSCALControl} ${classes.OSCALControlStatus} ${classes.OSCALControlChildLevel}`}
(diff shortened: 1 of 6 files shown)


src/components/OSCALControl.js
```diff
@@ -61,6 +61,21 @@ export default function OSCALControl(props) {
     );
   }
 
+  // When FedRAMP, retrieve the set-parameters from the initial implementation request statement
```
rgauss, line 64: This isn't necessarily specific to NIST or FedRAMP any profile could be defined this way, and `implReqStatements` is actually short for 'implemented-requirements statements'.

src/components/OSCALControlProse.js
```diff
@@ -55,17 +55,30 @@ function getParameterLabel(parameters, parameterId) {
...
-function getParameterValue(statementByComponent, parameterId) {
+function getParameterValue(
+  statementByComponent,
+  implReqSetParameters,
+  parameterId
+) {
   // Locate matching parameter to parameterId
-  if (!statementByComponent["set-parameters"]) {
+  let foundParameterSetting = null;
+  // If NIST, locate by by-component
```
rgauss, line 69: Here as well, not necessarily specific to NIST vs FedRAMP.

src/components/OSCALControl.js
```diff
@@ -61,6 +61,21 @@ export default function OSCALControl(props) {
...
+  let firstImplReqStatement = null;
```
rgauss, line 65: Why are we just pulling the first of each here?

src/components/OSCALSystemCharacteristics.js
```diff
@@ -55,7 +55,7 @@ export default function OSCALSystemCharacteristics(props) {
             </Grid>
             <Grid
               item
-              xs={7}
+              xs={12}
```
rgauss, line 58: This ends up with a lot of white space to the right of most titles.  Can we just shrink the system name and expand the props?

src/components/OSCALSystemImplementation.js
```diff
@@ -73,7 +77,14 @@ export default function OSCALSystemImplementation(props) {
                     {Object.entries(props.systemImplementation.components).map(
                       ([key, component]) => (
                         <TableRow key={key}>
-                          <TableCell component="th" scope="row">
+                          <TableCell
+                            component="th"
+                            scope="row"
+                            align="left"
+                            style={{
```
rgauss, line 84: Any reason this one doesn't use a class?

rgauss (changes requested):

src/components/OSCALControl.js
Bronstrom: Yes. In this sample it contains the `set-parameters` only in the first (initial) implemented requirement statement

src/components/OSCALControlProse.js
Bronstrom: Gotcha, will note this as a case when the `set-parameters` are found in the `by-component`

src/components/OSCALControl.js
Bronstrom: I had been working with mostly NIST profiles, and assumed FedRAMP had there own method of laying out this information. Thanks for the clarification. I'll change this to something like "Retrieve the set-parameters when contained in the initial implemented requirement statement"

src/components/OSCALSystemImplementation.js
Bronstrom: It would probably make sense to have a class for this one with the style change. Will make adjustments for this.

src/components/OSCALControl.js
rgauss: Right, but we don't have guaranteed ordering here.

What I think we're really looking for are `set-parameters` at the 'top-level' of the control, so within `ac-1_smt`, which won't necessarily be the first statement in the array.

Bronstrom: Oh okay, that makes more sense. Maybe instead here, we should find the 'top-level' of the control with an ends with "_smt" check?

src/components/OSCALSystemCharacteristics.js
Bronstrom: Is this closer to your idea? : 

image

rgauss: Yes, that looks better IMO.  Maybe give the system name one more grid size and the props one less?  Or do the props start to wrap if you do that?

src/components/OSCALControl.js
rgauss: Something like that, we may already be doing something similar elsewhere?

src/components/OSCALSystemCharacteristics.js
Bronstrom: The props appear to still display well with this adjustment. I'll post this fix.

src/components/OSCALControl.js
Bronstrom: I've added an endsWith check when searching through the array of `implReqStatements` for now. I haven't found a place where we're doing something similar but will continue looking.

src/components/OSCALSystemCharacteristics.js
Bronstrom: The props and system name seem to both fit well in their grid with this change. Will post this fix.

src/components/OSCALControl.js
```diff
@@ -61,6 +61,27 @@ export default function OSCALControl(props) {
     );
   }
 
+  // Retrieve the set-parameters when contained in the top-level implemented requirement statement
+  let topLevelImplReqStatement = null;
+  if (props.implReqStatements) {
+    // NOTE: The top level statement ends with "_smt"
+    props.implReqStatements.forEach((statement, index) => {
```
rgauss, line 68: Can this just use Array.find?

src/components/OSCALControl.js
```diff
@@ -61,6 +61,27 @@ export default function OSCALControl(props) {
...
+      if (statement["statement-id"].endsWith("_smt")) {
+        topLevelImplReqStatement =
+          props.implReqStatements[Object.keys(props.implReqStatements)[index]];
+      }
+    });
+  }
+  // NOTE: The top-level by-component is the first element in the by-components array
```
rgauss, line 75: I'm not sure this is an accurate statement.

Each `component` in the `by-components` array represents how each component address the control.

I believe `OSCALContorlImplementationImplReq` sends in a `componentId` that we should check for?

src/components/OSCALSystemImplementation.js
```diff
@@ -100,7 +112,10 @@ export default function OSCALSystemImplementation(props) {
                                         >
                                           {property.name}
                                         </TableCell>
-                                        <TableCell align="right">
+                                        <TableCell
+                                          className={classes.SmallTabelCell}
+                                          align="right"
```
rgauss, line 117: Should we move `align` to the `SmallTableCell` class?

src/components/OSCALSystemImplementation.js
```diff
@@ -100,7 +112,10 @@ export default function OSCALSystemImplementation(props) {
                                         >
                                           {property.name}
                                         </TableCell>
-                                        <TableCell align="right">
+                                        <TableCell
+                                          className={classes.SmallTabelCell}
```
rgauss, line 116: Table misspelled?

rgauss (changes requested):

src/components/OSCALControl.js
Bronstrom: I was a little hesitant to include that part as well. Grabbing the first array is definitely not the best practice here as this doesn't seem to be a consistent pattern. Also, thanks for clarifying on this. Your right about `OSCALControlImplementationImplReq` passing `componentId` as a prop. I can make a `by-component` id match check as well - I'm not seeing it elsewhere.

src/components/OSCALControl.js
Bronstrom: Definitely a more efficient solution!

Bronstrom:
I chose to make the set-parameter fetch into a separate function in `OSCALControl`. I was wondering if we would even want to split this up further or place it elsewhere - like in `./oscal-utils`? At this point, the `implReqStatements` grab and `by-component` id checks aren't used outside the scope of `OSCALControl`, but could potentially be referenced by components in the future?

src/components/OSCALControl.js
```diff
@@ -41,6 +41,32 @@ const useStyles = makeStyles(() => ({
...
+    );
+  }
+  // NOTE: The top-level by-component has a matching uuid to the componentId
+  let topLevelByComp = null;
+  if (topLevelImplReqStatement?.["by-components"]) {
+    topLevelByComp = topLevelImplReqStatement["by-components"].find(
+      (byComp) => byComp["component-uuid"] === componentId
+    );
+  }
+  return topLevelByComp?.["set-parameters"] || null;
```
tuckerzp, line 67: I could be wrong here, but is it possible to make this into just a pipe of `find().find()`? It seems like there could be a cleaner way of doing this.

tuckerzp (commented):
Looking good so far! This does not have to prevent an approval. I just had a few potential things to look into.

hreineck (approved):
Looks good to me. Viewer looks as specified locally.
I would like to see @tuckerzp 's comments implemented, but other than that looks great.

src/components/OSCALControl.js
```diff
@@ -41,6 +41,32 @@ const useStyles = makeStyles(() => ({
...
 }));
 
+/**
+ * Retrieve the set-parameters when contained in the top-level implemented requirement statement
+ *
+ * @param {Object} implReqStatements
+ * @param {String} componentId
+ * @returns An array of set-parameters
+ */
+function getImplReqSetParameters(implReqStatements, componentId) {
```
rgauss, line 51: Should this entire function be pushed down to `OSCALControlProse.js` since it's already responsible for finding the parameters and has the `componentId`?

rgauss (changes requested):

src/components/OSCALControl.js
mikeisen1: I will push the function to `OSCALControlProse.js`

src/components/OSCALControl.js
```diff
@@ -61,6 +62,12 @@ export default function OSCALControl(props) {
     );
   }
 
+  // Retrieve implemented requirements set-parameters
+  const setParams = getImplReqSetParameters(
```
rgauss, line 66: Sorry, I meant the function and the execution of that function could be moved to `OSCALControlProse`.

There shouldn't be any need to evaluate here and pass the results as `implReqSetParameters ` all the way down to `OSCALControlProse` where it's actually used, I think we should be able to just call from within `OSCALControlProse`.

rgauss (changes requested):

src/components/OSCALControlProse.js
```diff
@@ -102,6 +115,23 @@ function getConstraintsDisplay(modificationSetParameters, parameterId) {
...
+ *
+ * @param {Object} implReqStatements statements on how the control is implemented
+ * @param {String} componentId uuid of the component and desired by-component
+ * @returns An array of set-parameters
+ */
+function getImplReqSetParameters(implReqStatements, componentId) {
+  // Get the top-level implemented requirement statement
+  const topLevelByComp = implReqStatements
+    ?.find((statement) => statement["statement-id"].endsWith("_smt"))
+    ?.["by-components"]?.find(
```
kylelaker, line 129:
```suggestion
    ?.["by-components"]
    ?.find(
```

Would the linter allow this? Makes the chain a little more "obvious"

src/components/OSCALControlProse.js
```diff
@@ -55,17 +55,30 @@ function getParameterLabel(parameters, parameterId) {
...
+  implReqSetParameters,
+  parameterId
+) {
   // Locate matching parameter to parameterId
-  if (!statementByComponent["set-parameters"]) {
+  let foundParameterSetting = null;
+  // Locate set-parameters when found in the by-component
+  if (statementByComponent?.["set-parameters"]) {
+    foundParameterSetting = statementByComponent["set-parameters"].find(
+      (parameterSetting) => parameterSetting["param-id"] === parameterId
```
kylelaker, line 72: If we're using this twice (once per branch), we should move it to it's own function within `getParameterValue`.

```javascript
// This assumes that `parameterId` from `getParameterValue` is in scope
function parameterHasGivenId(parameterSetting) {
  return parameterSetting["param-id"] === parameterId;
}
```

And then you can use it as:

```javascript
if (statementByComponent?.["set-parameters"]) {
  foundParameterSetting = statementByComponent["set-parameters"].find(parameterHasGivenId);
} else if (implReqSetParameters) {
  foundParameterSetting = implReqSetParameters.find(parameterHasGivenId);
} else {
  return null;
}
```

But I'd also suggest that we've got quite a bit of repetition there still (between the if and the actual clause). And we may want to consider also doing something like:

```
const setParameters = () => statementByComponents?.["set-parameters"] || implReqSetParameters;
const foundParameterSetting = setParameters()?.find(parameterHasGivenId);
```

And since the next check is `if (!foundParameterSetting?.values) { return null; }` we shouldn't need to modify any code. But we could do a `if (!foundParameterSetting) { return null; }` if we really wanted to.

This has the benefits of not only reducing repetition but also keeping more things `const`. It's also one less `null` that we really need to worry about.

src/components/OSCALControlProse.js
```diff
@@ -102,6 +115,23 @@ function getConstraintsDisplay(modificationSetParameters, parameterId) {
...
+ * @returns An array of set-parameters
+ */
+function getImplReqSetParameters(implReqStatements, componentId) {
+  // Get the top-level implemented requirement statement
+  const topLevelByComp = implReqStatements
+    ?.find((statement) => statement["statement-id"].endsWith("_smt"))
+    ?.["by-components"]?.find(
+      (byComp) => byComp["component-uuid"] === componentId
+    );
+  return topLevelByComp?.["set-parameters"] || null;
```
kylelaker, line 132:
```suggestion
  return implReqStatements
    ?.find((statement) => statement["statement-id"].endsWith("_smt"))
    ?.["by-components"]?.find(
      (byComp) => byComp["component-uuid"] === componentId
    )
    ?.["set-parameters"] || null;
```

Any reason to keep the temporary value variable there?

kylelaker (commented):
Some thoughts but not really required.

src/components/OSCALControlProse.js
mikeisen1: I've tried locally but whenever I run `npm run lint`, the suggested change is fixed to what it originally looked like so I don't think so.

src/components/OSCALControlProse.js
mikeisen1: I made all recommended changes except the change in the conditional because we do `foundParameterSetting.values.toString()` right after the check and if `foundParameterSetting.values` is null, we would get an error.

src/components/OSCALControlProse.js
```diff
@@ -64,21 +64,13 @@ function getParameterValue(
...
-    );
-  } else {
-    return null;
+  function parameterHasGivenId(parameterSetting) {
+    return parameterSetting["param-id"] === parameterId;
   }
+  // Locate set-parameters when found in the by-component
+  const setParameters = () =>
+    statementByComponent?.["set-parameters"] || implReqSetParameters;
+  const foundParameterSetting = setParameters()?.find(parameterHasGivenId);
```
kylelaker, line 73:
```suggestion
  // Locate set-parameters when found in the by-component
  const setParameters = statementByComponent?.["set-parameters"] || implReqSetParameters;
  const foundParameterSetting = setParameters?.find(parameterHasGivenId);
```

We actually don't need to make this a function since we never use it again. Previously it was a little more necessary but with the rest of the cleanup, we don't need it anymore

kylelaker (changes requested):

rgauss (approved):

kkennedy26 (approved):
Adjustments look good

PR #132, opened 2021-07-20
EGRC-469

mikeisen1:
Finished subtask EGRC-469 where I migrated the data into one directory and updated test
and storybook file imports as necessary.

PR #143, opened 2021-07-28
Permit OSCAL Object URL as URL Parameter

### LINKED ISSUE #134: OSCAL Object URL as URL Parameter
As a user of the OSCAL Viewer, I want to be able to specify the URL to load for an object viewer as a URL parameter so that I can create links that others can view without having to navigate, then paste a URL to load, then click reload.

### Acceptance Criteria

Passing in a `url=...` URL parameter to any viewer type (Catalog, Profile, Component, SSP) replaces the default URL in the loading form and immediately loads that URL

mikeisen1:
Adjustments were made to allow for someone to pass in a url parameter in the browser window for each of the OSCAL viewers, overriding the corresponding default url. This allows for any data contained in a valid url to be loaded into an OSCAL viewer, lending greater flexibility to what can be displayed.

To input an external url and display its data, add the following to the end of the browser url:

`?url=<external-url>`.

Listed below are example files which can be used as external URLs:

1. Catalog Viewer: https://raw.githubusercontent.com/usnistgov/oscal-content/master/nist.gov/SP800-53/rev4/json/NIST_SP-800-53_rev4_catalog.json
2. SSP Viewer: https://raw.githubusercontent.com/usnistgov/oscal-content/master/examples/ssp/json/oscal_leveraged-example_ssp.json
3. Component Viewer: https://raw.githubusercontent.com/EasyDynamics/oscal-content/manual-fix-of-component-paths/examples/component-definition/json/example-component-min.json
4. Profile Viewer: https://raw.githubusercontent.com/usnistgov/oscal-content/master/nist.gov/SP800-53/rev4/json/NIST_SP-800-53_rev4_HIGH-baseline_profile.json.

Note that the component viewer example does not load properly due to an issue with one of the sources in the data. However, there were no other examples, so just seeing whether the url in the OSCAL Component Viewer url box changed will have to do for now. Additionally, the above are just examples, as there are other urls for each of the Catalog, SSP, and Profile viewers that can be used.

This will resolve #134


src/components/OSCALLoader.js
```diff
@@ -6,6 +6,12 @@ import OSCALCatalog from "./OSCALCatalog";
 import OSCALComponentDefinition from "./OSCALComponentDefinition";
 import OSCALProfile from "./OSCALProfile";
 import OSCALLoaderForm from "./OSCALLoaderForm";
+import {
+  defaultOscalCatalogUrl,
+  defaultOSCALComponentUrl,
+  defaultOSCALProfileUrl,
+  defaultOscalSspUrl,
+} from "../test-data/Urls";
```
tuckerzp, line 14: Do we want the `OSCALLoader` to depend on `/test-data`?

src/components/OSCALProfile.test.js
```diff
@@ -49,8 +68,54 @@ function testOSCALProfile(parentElementName, renderer) {
...
+    fireEvent.click(modButton);
+    expect(await screen.findByText("Modifications")).toBeVisible();
+    expect(await screen.findByText("Adds")).toBeVisible();
+  });
+}
 
-testOSCALProfile("OSCALProfile", profileRenderer);
+function testUrlOSCALProfile(url) {
+  const externalUrl = url.searchParams.get("url");
+  if (!externalUrl) {
```
tuckerzp, line 95: Could this be done in one line?

src/components/OSCALLoader.js
```diff
@@ -100,6 +97,11 @@ export default function OSCALLoader(props) {
   );
 }
 
+function urlToLoad(defaultUrl) {
```
tuckerzp, line 100: Could this be named better? Maybe documentation on purpose would help?

src/test-data/ComponentsData.js
```diff
@@ -86,6 +85,53 @@ export const componentDefinitionTestData = {
   },
 };
 
+export const externalComponentDefinitionTestData = {
```
tuckerzp, line 88: This feels really similar to `componentDefinitionTestData`. It might be needed duplication, but I was wondering if there was a better way to do this?

tuckerzp (changes requested):
Looking good so far! I had a few comments.

I also wonder if we need to test for both cases when a `url` is provided vs when it is not? Could we primarily test that things load properly when we load a url and just have a few tests to make sure things work otherwise?

src/components/OSCALLoader.js
mikeisen1: Those imported Urls are simply the default Urls, which were still in their respective files, so I decided to move them to the directory containing the test data. If it makes more sense, I can just move those Urls back or create a separate directory/file for those Urls.

src/test-data/ComponentsData.js
mikeisen1: The actual data contained in `externalComponentDefinitionTestData` is different from the data contained in `componentDefinitionTestData`. I added more component test data for testing, where I would test against a specific set of data depending on whether there is an external or no external Url. The tests were done in a way to as closely mimic what would happen in the viewers as possible.

src/test-data/SystemData.js
```diff
@@ -1,4 +1,8 @@
-import { metadataTestData, responsibleRolesTestData } from "./OtherData";
```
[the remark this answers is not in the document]
mikeisen1: If it was already done, I think I pulled from the develop branch prior to the change being made, so that's why you see the renaming in my files.

src/components/OSCALLoader.js
tuckerzp: It is only used in the `OSCALLoader` from what I can tell. Is there a reason to move them?

src/components/OSCALProfile.test.js
mikeisen1: I don't think so, as `externalUrl` is used a few lines down in calling the `testExternalProfile` functions and `externalUrl` is used in the renderer in the tests. This just makes it so I don't have to call `url.searchParams.get(url)` multiple times.

src/components/OSCALLoader.js
mikeisen1: Looking at the function, it makes sense to rename it, will do that, comment on its purpose, and push a new commit.

src/components/OSCALLoader.js
mikeisen1: I moved them because I assumed we were migrating all data and urls to the test-data directory, as a standard.

kylelaker: I'd like to find a way to not import `test-data` in actual code. Our production code should not rely on tests or their data. Can we investigate another path forward here? Maybe we move these values into app code and the tests import from the app code?

src/test-data/SystemData.js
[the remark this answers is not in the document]
tuckerzp: Oh maybe do a `git pull rebase` to be sure?

src/components/OSCALLoader.js
hreineck: I vote to leave them in the OSCALLoader.js file itself. I think it'd be easier to read and develop if those are used as constants in that piece of code and not lumped in with the test-data urls.

src/test-data/Urls.js
```diff
@@ -1,13 +1,39 @@
...
+  "https://raw.githubusercontent.com/EasyDynamics/oscal-content/5ef5b3709045ba3aab1473813c13436e886d6185/examples/component-definition/json/example-component.json";
+
+export const defaultOscalCatalogUrl =
+  "https://raw.githubusercontent.com/usnistgov/oscal-content/master/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_catalog.json";
+
+export const defaultOSCALComponentUrl =
+  "https://raw.githubusercontent.com/EasyDynamics/oscal-content/manual-fix-of-component-paths/examples/component-definition/json/example-component.json";
+
+export const defaultOSCALProfileUrl =
+  "https://raw.githubusercontent.com/usnistgov/oscal-content/master/nist.gov/SP800-53/rev4/json/NIST_SP-800-53_rev4_MODERATE-baseline_profile.json";
```
kylelaker, line 15: Is this the same as `parentUrlTestData`? Will it always be?

kylelaker (changes requested):

src/test-data/ComponentsData.js
tuckerzp: Yeah and I think this goes back to my question above: is there a reason to use custom `componentDefinitionTestData` when we can just test that the `url` loads and matches `externalComponentDefinitionTestData`? It seems that loading a `url` vs loading `defaulturl` should not be that different. We could test that we are able to load something if no `url` is given but otherwise just provide a `url` and test that it loads properly.

src/test-data/Urls.js
mikeisen1: Yes, this is the same as parentUrlTestData

tuckerzp: It seems we do not use `parentUrlTestData` anywhere. Can we just get rid of it?

mikeisen1: That depends on how we handle the default viewer urls

src/components/OSCALLoader.js
mikeisen1: Ok, seems like it's best to move them back to the OSCALLoader.js file, so I'll go ahead and do that.

tuckerzp: Yeah I agree, minimizing the coupling of data is important if we want the components to be usable separately.

src/components/OSCALLoader.js
```diff
@@ -100,6 +97,15 @@ export default function OSCALLoader(props) {
...
 
+/*This function checks whether there is a url parameter provided
+  in the browser (window) url. If so, this returns the url defined
+  in the url parameter, and the provided defaultUrl otherwise. The
+  returned url and its data is then loaded into the viewer.*/
+function loadDefaultOrExternalUrl(defaultUrl) {
+  const url = new URLSearchParams(window.location.search).get("url");
+  return url || defaultUrl;
+}
+
```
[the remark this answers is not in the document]
rgauss: I'd add that a `load` prefix here might be confusing as it's only returning a URL string, not really loading anything.

Another approach would be to turn this in to a simpler `getRequestedUrl()` function with no arguments, then each `OSCALLoader` call could use something like:
```
oscalUrl={getRequestedUrl() || defaultOscal*Url}
```
Might be more readable, maybe not, interested in what others think.

src/components/OSCALCatalogGroup.test.js
```diff
@@ -1,15 +1,34 @@
...
+  test("ExternalOSCALCatalog displays control groups", () => {
+    render(<OSCALCatalogGroup group={externalCatalogTestData} />);
+    const result = screen.getByText("External Access Control");
+    expect(result).toBeVisible();
+  });
+}
+
+const url = new URL("https://www.oscal-catalog.com");
+testUrlOSCALCatalog(url);
+url.searchParams.set("url", revFourCatalog);
```
rgauss, line 33: I'm not sure this is really testing the functionality we're adding.

This looks like it would be the equivalent of someone pasting a URL of `https://www.oscal-catalog.com?url=https://raw.githubusercontent.com/usnistgov/...` into the URL input and click reload, not grabbing it from the browser's `window.location`.

Note that we don't need to test getting the URL from the browser at the component level though.

rgauss (changes requested):
At a high level, the majority of these test changes are really testing whether a different source URL loads correctly, not necessarily related to whether or not that source URL was provided as a URL parameter when making the request for the viewer in a browser, which we probably should not be testing with component tests.

While those additional tests may be valuable, we might want to consider whether we want to duplicate all tests with different expectations, or restructure the existing tests to follow a pattern of "for this URL, here are the expected values for the different elements", keeping in mind that some sources may not be expected to render certain elements at all.

That close tie of the source URL and expected rendering might also lend itself to the source URL being 'close to the expectations', i.e. in the test file, rather than in a separate URL file.

See a few other line-specific comments.

src/test-data/ComponentsData.js
tuckerzp: Another issue with testing for specific examples is that those examples change a lot. Which would mean breaking tests that might not really be broken.

src/components/OSCALCatalogGroup.test.js
tuckerzp: > Might be more readable, maybe not, interested in what others think.

When looking at the code I thought the same thing.

> While those additional tests may be valuable, we might want to consider whether we want to duplicate all tests with different expectations, or restructure the existing tests to follow a pattern of "for this URL, here are the expected values for the different elements", keeping in mind that some sources may not be expected to render certain elements at all.

I also wonder if it is a good idea to test for specific things to be found when loading external components that, as we have seen before, change pretty frequently.

tuckerzp:
> Might be more readable, maybe not, interested in what others think.

When looking at the code I thought the same thing.

> While those additional tests may be valuable, we might want to consider whether we want to duplicate all tests with different expectations, or restructure the existing tests to follow a pattern of "for this URL, here are the expected values for the different elements", keeping in mind that some sources may not be expected to render certain elements at all.

I also wonder if it is a good idea to test for specific things to be found when loading external components that, as we have seen before, change pretty frequently.

src/components/OSCALCatalogGroup.test.js
mikeisen1: So does it make sense to even add more tests with this issue?

example/package-lock.json
```diff
@@ -24,6 +24,7 @@
       }
     },
     "..": {
+      "name": "oscal-react-library",
```
kylelaker, line 27: Why is this being added with no corresponding change to `package.json`?

src/components/OSCALLoader.js
```diff
@@ -100,6 +103,13 @@ export default function OSCALLoader(props) {
   );
 }
 
+function getRequestedUrl(){
+  /*Returns url parameter provided by the browser url, if
+    it exists. If the url parameter exists, we want to
+    override the default viewer url.*/
```
kylelaker, line 109: This doesn't seem to match our conventions for comments. Consider:

```suggestion
  // Returns url parameter provided by the browser url, if
  //  it exists. If the url parameter exists, we want to
  //  override the default viewer url.
```

or 
```suggestion
  /* 
   * Returns url parameter provided by the browser url, if
   * it exists. If the url parameter exists, we want to
   * override the default viewer url.
   */
```

I will say that looking at the results of:
```
$ git grep -A2 '/\*[^*]' src/ # Find comments that start with `/*` but not `/**` and the two lines after
$ git grep -A2 '//' src/ # Find comments that start with `//` and the two lines after
```

The `//` style is far more common in the code base which makes sense because it makes it easier to comment out large blocks of code for testing (because `//` comments can be nested in `/*` comments but `/*` comments cannot be nested in `//` comments).

src/test-data/SystemData.js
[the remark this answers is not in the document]
kylelaker: Even a `git pull origin develop` that does a merge commit should help resolve that.

kylelaker (changes requested):
As @tuckerzp mentioned, there are a few changes here that I am wondering if they're a result of the merge conflict and the diff is off. Can you update with the `develop` branch and resolve the merge conflicts then push again?

src/test-data/Urls.js
kylelaker: Hey @mikeisen1 please try to let the thread authors resolve conversations. I think I still need an explanation for exactly what we plan to use this for, why the value is the same, and why it's a different variable. Are we expecting to eventually use different URLs? If the answer is "it depends" then I think we should talk about the assumptions and design and make sure we understand what it depends on and what we've chosen here. I am missing an understanding of that justification and it seems that @tuckerzp may be as well.

mikeisen1: I am not sure why the value is the same and they're in different variables. Both `parentUrlTestData` and `defaultOSCALProfileUrl` have been in the codebase since before I have been working on the project.

My "depends" reply was in response to Zach Tucker's reply. When I moved the default urls back to the OSCALLoader file, I kept their original scope (none of them had `export`). So, I would only delete `parentUrlTestData` if I moved `defaultOSCALProfileUrl` back to the OSCALLoader file and did not export it. As for whether they will both always be the same, I think they would only be different if we want the urls used in the stories and tests to be different from the default viewer urls.

example/package-lock.json
mikeisen1: I'm not really sure why, I've never manually touched either file.

kylelaker: Let's see if we can find that out. I am concerned that we're going to keep seeing PRs battle over changing this line in this file.

tuckerzp: Wasn't this issue looked into a bit before with #132 ?

src/components/OSCALLoader.js
tuckerzp: @mikeisen1 it was mentioned above but I think it is generally better to address the comment before resolving conversation. Just so it doesn't get lost in this large thread.

src/components/OSCALCatalogGroup.test.js
kylelaker: I think we want to make sure that we:
1. _do_ test documented and public functionality of the application
2. _do not_ test non-deterministic behavior of the application

I wonder if we should `export` `getRequestedUrl` and find a way to mock `window` or `window.location` (or set their attributes or whatever) so that we can at least make sure that we are parsing/retrieving it correctly. Maybe this test is pretty trivial but it's a good thing that could be a source of regressions in the future. Or someone may introduce a feature that causes the app to crash when `url` isn't present.

I'll defer to someone with stronger QA and testing knowledge than me. Perhaps @rgauss or @snorouzzadeh can provide some suggestions on the specifics here.

src/test-data/Urls.js
tuckerzp: > Both `parentUrlTestData` and `defaultOSCALProfileUrl` have been in the codebase since before I have been working on the project.

So that is an issue that I have had to deal with a few time. A lot of changes were made and thought out long before we started on project. Using `Git Blame` and checking the history for why something was done and when is super helpful to get an idea of why something may or may not be important.

From looking purely at the code you have added, `parentUrlTestData` is not used anywhere as of now. `defaultOSCALProfileUrl` seems to have replaced all those places where it was used. Going off of conversations above, `defaultOSCALProfileUrl` is not really needed in `Urls.js`, so `parentUrlTestData` could be needed again. 

With that said, I am curious why they are the same value and in different variables. Maybe look into why they might be similar?

src/components/OSCALCatalogGroup.test.js
tuckerzp: Having played around with our tests a bit, I think currently we have 3 tests that rely on fetching data outside of our mock test data prior to this PR. I think we should really try not to need external data when running tests as they are already pretty slow as is. Whether or not we load our own test data or data from an external source really should not make the page have any significant differences we would want to test for. 

I agree that it would make a lot more sense to check `getRequestedUrl` works for the loader and allow our other tests to stay the same.

mikeisen1: For testing, I planned on creating a separate folder for tests related to non-component code and creating tests in that folder. That would allow us to separate component tests with tests on other functionality such as grabbing the url from a browser.

tuckerzp: Could the test for grabbing a `url` just be in it's own `OSCALLoader.test.js` file? I wonder if we need a ton of tests to see the functionality of using a `url` parameter. 

I am curious what @rgauss and @kylelaker think about this. Adding a ton of testing functionality for a minor change seems unneeded here. This has definitely brought up some questions on how we could test external files, but I feel like deciding how to do that and if we should might be out of scope.

mikeisen1: I'm in agreement on putting it in its own file, that's what I have at the moment. However, I didn't create an `OSCALLoader.test.js` file because it seemed like from the original comment that we should not test this in a component test file.

mikeisen1: And there are not many tests to write, as we just need to test if `getRequestedUrl` works with no url parameter and when there is a url parameter.

package.json
```diff
@@ -12,11 +12,12 @@
     "node": ">=10"
   },
   "dependencies": {
+    "@jest/globals": "^26.6.2",
```
tuckerzp, line 15: Out of curiosity isn't `jest` another react testing library? Is this a needed import? Is there something like this already in `@testing-library/react`?

src/common-tests/UrlParameter.test.js
```diff
@@ -0,0 +1,18 @@
...
+    // The Url parameter should always initially be null.
+    expect(getRequestedUrl()).toBeNull();
+  });
+
+  test("Url Parameter Exists", () => {
+    const url = new URL(window.location);
+    url.searchParams.set("url", urlParameterTestUrl);
+    window.history.pushState({}, "", url);
+    expect(getRequestedUrl()).not.toBeNull();
+    expect(getRequestedUrl()).toEqual(urlParameterTestUrl);
```
tuckerzp, line 16: Is it needed to check that this is not `null` and that it equals `urlParameterTestUrl`?

tuckerzp (changes requested):

package.json
mikeisen1: Jest, while initially built to support react testing, is not another react testing library. The react testing library we are currently using only supports component testing and not unit tests like the ones I wrote (at least to my knowledge, correct me if I'm wrong).

src/common-tests/UrlParameter.test.js
mikeisen1: Yeah, I realize that now how the null check is unnecessary.

package.json
mikeisen1: I took a look at package.json and it looks like we already have Jest capabilities.

tuckerzp:
Just to keep history clean, would this be a good place to squash and force push the `Resolve Linter Errors` commits? Either way is fine, I was mostly just curious what everyone thought in this case.

example/public/robots.txt
```diff
@@ -1,3 +1,3 @@
-# https://www.robotstxt.org/robotstxt.html
-User-agent: *
-Disallow: /
+# https://www.robotstxt.org/robotstxt.html
+User-agent: *
+Disallow: /
```
tuckerzp, line 3: For example

tuckerzp (changes requested):
Has line ending issue been addressed?

example/public/robots.txt
mikeisen1: Line ending issue was addressed at least on my end.

tuckerzp: If it is not line ending issue, why is this change occurring? It seems like these files should not need to be deleted and then added in an identical way.

src/components/OSCALLoader.js
```diff
@@ -100,6 +103,13 @@ export default function OSCALLoader(props) {
   );
 }
 
+export function getRequestedUrl() {
+  // Returns url parameter provided by the browser url, if
+  //  it exists. If the url parameter exists, we want to
+  //  override the default viewer url.
+  return new URLSearchParams(window.location.search).get("url");
+}
+
```
tuckerzp, line 112: Now that this is being exported, is JSDoc needed here?

rgauss (approved):

kkennedy26 (approved):
Looks good

kylelaker (approved):
Thanks for your work on this! Looking really good!

PR #144, opened 2021-07-28
Fix Reload Button Crash

hreineck:
Pressing the Reload button in rapid succession would crash
the loader by calling `handleReloadClick` multiple times
very quickly and getting to a state where the loader was running
and the `isLoaded` state variable was set incorrectly.
Adding this conditional should prevent `handleReloadClick`
from setting this at an unintended time.


src/components/OSCALLoader.js
```diff
@@ -56,8 +56,10 @@ export default function OSCALLoader(props) {
   };
 
   const handleReloadClick = () => {
-    setIsLoaded(false);
-    loadOscalData(oscalUrl);
+    if (isLoaded) {
```
rgauss, line 59: Should we consider also visually disabling the reload button as part of this?

rgauss (changes requested):

src/components/OSCALLoader.js
hreineck: Good idea, I'll look into that.

rgauss: Should be able to just set the `disabled` flag: https://material-ui.com/components/buttons/

src/components/OSCALCatalog.js
```diff
@@ -17,6 +17,10 @@ const useStyles = makeStyles((theme) => ({
 export default function OSCALCatalog(props) {
   const classes = useStyles();
 
+  useEffect(() => {
+    if (props.setContentLoaded) props.setContentLoaded(true);
+  }, []);
+
```
tuckerzp, line 23:
```suggestion
  useEffect(() => {
    if (props.setContentLoaded) { 
       props.setContentLoaded(true);
    }
  }, []);
```
Kinda small, but I think that the braces add a lot to readability.

src/components/OSCALComponentDefinition.js
```diff
@@ -19,16 +19,21 @@ export default function OSCALComponentDefinition(props) {
   const [isLoaded, setIsLoaded] = useState(false);
   const classes = useStyles();
 
+  const setLoadedStates = () => {
+    setIsLoaded(true);
+    if (props.setContentLoaded) props.setContentLoaded(true);
```
tuckerzp, line 24:
```suggestion
    if (props.setContentLoaded) {
       props.setContentLoaded(true);
    }
```
Same thing here.

tuckerzp (changes requested):
Looks good to me so far. I had a few minor suggestions.

src/components/OSCALSsp.js
```diff
@@ -30,19 +30,24 @@ export default function OSCALSsp(props) {
     sspParties = ssp.metadata.parties;
   }
 
+  const setLoadedStates = () => {
+    setIsLoaded(true);
+    if (props.setContentLoaded) props.setContentLoaded(true);
```
tuckerzp, line 35:
```suggestion
    if (props.setContentLoaded) {
       props.setContentLoaded(true);
    }
```

src/components/OSCALLoader.js
```diff
@@ -56,8 +57,11 @@ export default function OSCALLoader(props) {
...
+      setIsLoaded(false);
+      setContentLoaded(false);
```
mikeisen1, line 62: Wouldn't it make more sense to call `setLoadedStates` instead and get rid of these two function calls?

mikeisen1 (changes requested):
Overall everything seems to work great, as the tests pass locally and I can see the added functionality. My only comment, besides my requested change, is that it seems we can still click the reload button multiple times, which increases the time it takes for the page to load. I'm not sure how (if you could) disable the reload button before a second click, but I just thought I'd mention this.

hreineck:
> Overall everything seems to work great, as the tests pass locally and I can see the added functionality. My only comment, besides my requested change, is that it seems we can still click the reload button multiple times, which increases the time it takes for the page to load. I'm not sure how (if you could) disable the reload button before a second click, but I just thought I'd mention this.

Agreed; it feels like it freezes on a mouse click and "buffers" the click even when the button is disabled. I'll look into this further, but has anyone worked with these types of React components and have any tips on making it run a bit smoother?

rgauss (commented):
Looking good overall.  We could consider a slightly different approach that lets `OSCALLoader` handle state changes, passes that on to the `OSCALLoaderForm`, and passes an `onResolutionComplete` function to the OSCAL objects which are responsible for calling it when their resolution is done.  I'll create a parallel draft PR to describe what I mean.

hreineck:
> Looking good overall. We could consider a slightly different approach that lets `OSCALLoader` handle state changes, passes that on to the `OSCALLoaderForm`, and passes an `onResolutionComplete` function to the OSCAL objects which are responsible for calling it when their resolution is done. I'll create a parallel draft PR to describe what I mean.

Thanks! This seems a bit cleaner. I'd like to see what others think, but I vote we go with your alternate approach.

mikeisen1:
I still see the issue, when on the profile viewer, of being able to click on the reload button multiple times (even when it's disabled). I initially wondered if this is an issue of simply displaying all the data on the screen. However, the background color of the reload button correctly reflects whether all the data has been rendered so that does not seem to be the issue. My guess is that we are incorrectly modifying  `isResolutionComplete`, as it is somehow being modified before the profile rendering is complete. Other than that, everything else looks good and works as expected.

rgauss:
> I still see the issue, when on the profile viewer, of being able to click on the reload button multiple times (even when it's disabled). I initially wondered if this is an issue of simply displaying all the data on the screen. However, the background color of the reload button correctly reflects whether all the data has been rendered so that does not seem to be the issue. My guess is that we are incorrectly modifying `isResolutionComplete`, as it is somehow being modified before the profile rendering is complete. Other than that, everything else looks good and works as expected.

@mikeisen1, I've noticed that there are certain states where the browser is pretty much frozen, and I'm not sure the additional clicks are actually being registered, which doesn't seem to have to do with the reload functionality as much as resolving and rendering controls.  Is that what you're seeing?

mikeisen1:
> @mikeisen1, I've noticed that there are certain states where the browser is pretty much frozen, and I'm not sure the additional clicks are actually being registered, which doesn't seem to have to do with the reload functionality as much as resolving and rendering controls. Is that what you're seeing?

I took a look at it a little more and I noticed that if I click the reload button while the profile viewer is reloading, I only see that exactly one more click is registered no matter how many times I click on the button.

rgauss:
> > @mikeisen1, I've noticed that there are certain states where the browser is pretty much frozen, and I'm not sure the additional clicks are actually being registered, which doesn't seem to have to do with the reload functionality as much as resolving and rendering controls. Is that what you're seeing?
> 
> I took a look at it a little more and yeah it does seem that it is an issue of the browser being frozen.

Ah, so only on the initial load?  I think we can fix that by just setting the initial default of `isResolutionComplete` to false [here](https://github.com/EasyDynamics/oscal-react-library/pull/144/files#diff-64f425e88729c3663f483e42f79e52e68757e8c9958eb4035b0b0b934ea5a0ebR32).

src/components/OSCALCatalog.js
```diff
@@ -17,6 +17,10 @@ const useStyles = makeStyles((theme) => ({
 export default function OSCALCatalog(props) {
   const classes = useStyles();
 
+  useEffect(() => {
```
rgauss, line 20: Is `useEffect` needed, or can we just call `props.onResolutionComplete()`?

src/components/OSCALLoader.js
```diff
@@ -29,6 +29,7 @@ const defaultOscalSspUrl =
 export default function OSCALLoader(props) {
   const [error, setError] = useState(null);
   const [isLoaded, setIsLoaded] = useState(false);
+  const [isResolutionComplete, setIsResolutionComplete] = useState(true);
```
rgauss, line 32: The may need to change to `false` based on PR comments.

rgauss (changes requested):

mikeisen1:
> > > @mikeisen1, I've noticed that there are certain states where the browser is pretty much frozen, and I'm not sure the additional clicks are actually being registered, which doesn't seem to have to do with the reload functionality as much as resolving and rendering controls. Is that what you're seeing?
> > 
> > 
> > I took a look at it a little more and yeah it does seem that it is an issue of the browser being frozen.
> 
> Ah, so only on the initial load? I think we can fix that by just setting the initial default of `isResolutionComplete` to false [here](https://github.com/EasyDynamics/oscal-react-library/pull/144/files#diff-64f425e88729c3663f483e42f79e52e68757e8c9958eb4035b0b0b934ea5a0ebR32).

At least on my end, I notice that happens not only on the initial load, but also sometimes when I click on reload.

rgauss (approved):

mikeisen1 (approved):

PR #145, opened 2021-07-28
Add withdrawn check for the Controls

snorouzzadeh:
Add a check to see if the Control is withdrawn.


kylelaker (approved):
Nice fix! Thank you!

I wondered whether it'd be worth it to write a test to prevent future regressions but since this seems to possibly be NIST SP 800-53-specific anyway and since it returns some CSS that we may choose to change, I think it's probably not worth testing.

tuckerzp (approved):
Looks good! Thank you for your work on this @snorouzzadeh

rgauss (approved):

PR #146, opened 2021-07-29
Fix Update SSP and Component Definition to Display Profile Modifications

### LINKED ISSUE #75: Update SSP and Component Definition to Display Profile Modifications

tuckerzp:
This PR is a fix to #124 and should hopefully closes #75

This reworks `ProfileResolver` to save `modifications` per imported profiles. `SSP` and `Component Definition` should now properly display `modifications` only when given an imported profile. 

Tests have also been changed. Changes in how we get `modifications` for each imported profile makes these tests no longer needed. Both the `SSP` and `Component Definition` rely on `OSCALControlImplementationImplReq` to display modifications. My reason for thinking these tests are unneeded is really due to how we pass modifications down to `Control` making it redundant to check on multiple levels that modifications are passed down. I would like if we can confirm that I am right in doing this as I still consider myself fairly new to the tests.


tuckerzp:
I am assigning @hreineck to hopefully address comments in my absence.

rgauss (changes requested):
The component definition viewer does not seem to show the parameter constraints for the FedRAMP High profile for the default component definition URL.

hreineck:
> The component definition viewer does not seem to be the parameter constraints for the FedRAMP High profile for the default component definition URL.

@rgauss could you clarify the intended behavior? seems this does not need to display any parameter constraints. It's currently calling `OSCALReplacedProseWithByComponentParameterValue`, but with no statement arguments; the default Component url doesn't have any `statements` props to pass in.

rgauss:
> > The component definition viewer does not seem to be the parameter constraints for the FedRAMP High profile for the default component definition URL.
> 
> @rgauss could you clarify the intended behavior? seems this does not need to display any parameter constraints. It's currently calling `OSCALReplacedProseWithByComponentParameterValue`, but with no statement arguments; the default Component url doesn't have any `statements` props to pass in.

In #75 acceptance criteria we see:
> ... example component definition ... should display that profile's AC-2.3 constraints

I'm not sure what you mean by "the default Component url doesn't have any `statements` props to pass in".

Both the SSP and the Component Definition viewers should display those parameter constraints so that a user can immediately see whether the parameter value they've entered is within those constraints rather than having to navigate to the Profile.

src/components/oscal-utils/OSCALComponentResolver.js
```diff
@@ -1,5 +1,13 @@
 import OSCALResolveProfileOrCatalogUrlControls from "./OSCALProfileResolver";
 
+function processesToRun(componentDefinition) {
+  let count = 0;
+  componentDefinition.components.forEach((component) => {
+    count += component["control-implementations"]?.length ?? 0;
+  });
+  return count;
```
kylelaker, line 11:
```suggestion
  return componentDefinition.components.reduce((acc, component) => acc + component?.length ?? 0, 0);
```

src/components/oscal-utils/OSCALComponentResolver.js
```diff
@@ -23,8 +32,16 @@ export default function OSCALComponentResolveSources(
         controlImplementation.source,
         parentUrl,
         componentDefinition["back-matter"], // not actually used for resolution in components
-        onSuccess,
-        onError,
+        () => {
+          pendingProcesses--;
```
kylelaker:
```suggestion
          pendingProcesses -= 1;
```

src/components/oscal-utils/OSCALComponentResolver.js
```diff
@@ -23,8 +32,16 @@ export default function OSCALComponentResolveSources(
...
+          if (pendingProcesses == 0) {
```
kylelaker:
```suggestion
          if (pendingProcesses === 0) {
```

or 
```suggestion
          if (!pendingProcesses) {
```

src/components/oscal-utils/OSCALComponentResolver.js
```diff
@@ -23,8 +32,16 @@ export default function OSCALComponentResolveSources(
...
+            onSuccess();
+          }
+        },
+        () => {
+          pendingProcesses--;
```
kylelaker:
```suggestion
          pendingProcesses -= 1;
```

kylelaker (changes requested):

src/components/oscal-utils/OSCALComponentResolver.js
```diff
@@ -1,5 +1,16 @@
 import OSCALResolveProfileOrCatalogUrlControls from "./OSCALProfileResolver";
 
+function processesToRun(componentDefinition) {
+  if (!componentDefinition.components.forEach){ 
+    return 0;
+  }
```
kylelaker, line 6:
```suggestion
  if (!componentDefinition.components.forEach){ 
    return 0;
  }
```

This doesn't make any sense? `forEach` is part of the `Array.prototype`? And it's a function -- it's not even being invoked here? If you apply my other suggestion a `?.reduce` followed by `?? 0` might accomplish what you want here?

src/components/oscal-utils/OSCALComponentResolver.js
kylelaker: Now that it's so much simpler, we can just inline this rather than keeping it as a separate function call

kylelaker: Or honestly even

```suggestion
  componentDefinition.components
  ?.map((component["control-implementations"]?.length ?? 0)
  ?.reduce((acc + len) => acc + len)
  || 0;
```

But that may be an extra iteration..

src/components/oscal-utils/OSCALComponentResolver.js
hreineck: This was added because I thought there were cases where `components` was an object, not an array.
Turns out the actual viewer resolves this as an array, but the test data passes in an object.
I believe the test data is incorrect and should be changed to pass in an array, but someone can correct me if that's wrong.

kylelaker: Looking at [the specification outline](https://pages.nist.gov/OSCAL/reference/latest/complete/json-outline/), `components` must be an array of 1 or more elements (if `components` is present`). If the tests are passing in an object, the tests are not compliant with the specification.

src/components/oscal-utils/OSCALComponentResolver.js
```diff
@@ -9,16 +9,35 @@ export default function OSCALComponentResolveSources(
   // Fixing linting error here would take significant change to codebase given how we use props.
   /* eslint no-param-reassign: "error" */
   componentDefinition.resolvedControls = [];
+
+  let pendingProcesses = componentDefinition.components
```
kylelaker, line 13:
```suggestion
  const pendingProcesses = componentDefinition.components
```

will this work?

src/components/oscal-utils/OSCALComponentResolver.js
```diff
@@ -9,16 +9,35 @@ export default function OSCALComponentResolveSources(
...
         componentDefinition.resolvedControls,
+        controlImplementation.modifications,
         controlImplementation.source,
         parentUrl,
         componentDefinition["back-matter"], // not actually used for resolution in components
-        onSuccess,
-        onError,
+        () => {
+          pendingProcesses -= 1;
+          if (pendingProcesses === 0) {
```
kylelaker, line 33:
```suggestion
          if (!pendingProcesses) {
```

will this work?

src/components/oscal-utils/OSCALComponentResolver.js
hreineck: No, because it needs to be decremented later.

hreineck: But maybe `var` instead? not sure what's preferred between `let` and `var` but this is function-scoped which works with `var`.

kylelaker: We want `let` over `var` in basically ever scenario. But we like `const` the most but I forgot this is mutable. So we need `let`. Thanks for explaining!

mikeisen1 (approved):
Everything works great and all requested changes have been resolved.

rgauss:
@hreineck, looks like this still has conflicts to be resolved.  Other than that looks good.

kkennedy26 (approved):
good work on the `pendingProcesses` workaround

rgauss (approved):

rgauss:
@kylelaker, this look OK now?

kylelaker (approved):
Phenomenal work here! Thanks for the iterations and good discussion. I really appreciate you being able to take over on this PR and get it delivered.

tuckerzp:
@hreineck Thanks again for taking this for me! Great job!

PR #152, opened 2021-08-11
Render Skeleton When Loading Profile Viewer

hreineck:
Modified the way the profile viewer loads to show only the skeleton
card when the page is loading in; this fixes the issues with the
skeleton not rendering at all, as well as making it a bit more clear
that the page is still loading content.

### DIFF
--- a/src/components/OSCALProfile.js
+++ b/src/components/OSCALProfile.js
@@ -89,18 +89,14 @@ export default function OSCALProfile(props) {
           />
         ))
       ) : (
-        <CardContent key="skeleton-card">
-          <span
-            style={{ marginTop: 5, display: "flex", gap: "1em" }}
-            key="controls load 0"
-          >
-            <Skeleton variant="text" width="25em" height="3em" />
-            <Skeleton variant="circle" width="3em" height="3em" />
-          </span>
-          <Skeleton variant="text" width="10em" height="2.5em" />
-          <Skeleton variant="rect" width="100%" height={115} />
-          <Skeleton variant="text" width="6.5em" height="3.5em" />
-        </CardContent>
+        <div>
+          <CardContent key="skeleton-card-0">
+            <Skeleton variant="rect" width="100%" height={115} />
+          </CardContent>
+          <CardContent key="skeleton-card-1">
+            <Skeleton variant="rect" width="100%" height={115} />
+          </CardContent>
+        </div>
       )}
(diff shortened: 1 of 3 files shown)


hreineck:
In the[previous commit](https://github.com/EasyDynamics/oscal-react-library/pull/114/commits/39b895773c3cd165b759756fbcd7ff1063215482) that initially sets up the skeleton,
there is an array of 3 skeleton cards rendered. I felt the single card looked a bit better, but if rendering multiple cards is desired that can be implemented instead.

src/components/OSCALProfile.js
```diff
@@ -100,18 +77,31 @@ export default function OSCALProfile(props) {
...
+        <CardContent key="skeleton-card">
+          <span
+            style={{ marginTop: 5, display: "flex", gap: "1em" }}
+            key="controls load 0"
+          >
+            <Skeleton variant="text" width="25em" height="3em" />
+            <Skeleton variant="circle" width="3em" height="3em" />
+          </span>
+          <Skeleton variant="text" width="10em" height="2.5em" />
+          <Skeleton variant="rect" width="100%" height={115} />
```
rgauss, line 101: Can we see what it looks like with 2 cards that only contain this 100% width `rect`?

rgauss (changes requested):

rgauss (changes requested):
Sorry, looked better before my terrible suggestion :worried:.

rgauss (approved):
Looks good, provided that the distribution (commands the pipeline execute) displays things properly.

kkennedy26 (approved):
Comparing the two different styles, it definitely looks better this way having the skeleton for loading imported controls instead of the whole page. Also, it isn't part of this task, but there is a console error for a key prop when running tests. I don't know if there is an issue for it already.

tuckerzp:
> Also, it isn't part of this task, but there is a console error for a key prop when running tests.

Out of curiosity, was console error there before this branch was created? Just don't want to unintentionally add errors to fix later.

kkennedy26:
> Out of curiosity, was console error there before this branch was created? Just don't want to unintentionally add errors to fix later.

It must have been there before this branch, but not exactly sure where or when this error occurred.

mikeisen1 (approved):
This looks so much better with the skeleton rendering than without. Great work!

PR #154, opened 2021-08-16
Modify Profile Test to Fix Intermittent Failures

hreineck:
Changed the profile test to use waitFor() instead of an await
on findByText().
Occasionally, certain conditions caused findByText() to throw an error,
likely due to the long load times for the imported controls in the
profile viewer. The errors caused the tests to fail even though the
code should pass.
This change should resolve these errors as waitFor() retries getByText()
until it does not throw an error, more resiliently handling the
conditions that caused findByText() to throw errors.

### DIFF
--- a/src/components/OSCALProfile.test.js
+++ b/src/components/OSCALProfile.test.js
@@ -28,12 +28,13 @@ function profileRenderer() {
 }
 
 function testOSCALProfile(parentElementName, renderer) {
+  jest.setTimeout(6000);
   test(`${parentElementName} displays controls`, async () => {
     act(() => {
       renderer();
     });
     const result = await waitFor(() => screen.getByText("ac-1"), {
-      timeout: 10000,
+      timeout: 5000,
     });
     expect(result).toBeVisible();
   });
@@ -43,7 +44,7 @@ function testOSCALProfile(parentElementName, renderer) {
     const modButton = await screen.findByRole(
       "button",
       { name: "ac-1 modifications" },
-      { timeout: 10000 }
+      { timeout: 5000 }
     );
     fireEvent.click(modButton);
     expect(await screen.findByText("Modifications")).toBeVisible();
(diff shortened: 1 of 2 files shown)


hreineck:
Might need some help testing this one - this change seems to work but I may just be getting really lucky (unlucky?) with the tests passing all the time. Before the fix, the test giving a false negative was pretty rare, so if anyone sees that behavior occur again even with this change please drop a comment in this PR.

snorouzzadeh:
I do not see any behavior occur during my testing. I think this should be good to go.

hreineck:
Turned up some other weird issues while testing; seems that the explicit timeout of 10000 ms doesn't actually do anything. A failing test will now return this output:
image
This is because jest sets a timeout for every test that takes precedence over the `timeout` field of `waitFor()`
As the error message says this can be remedied by using `jest.setTimeout()` to make that timeout higher than 10000ms, which gives this error instead:
image
I believe the second error message is preferable on a failed test because it's a bit more descriptive of the actual error.
For now I'm going to reduce the explicit timeouts to 5000ms and call `jest.setTimout(6000)`because that produces the more descriptive error and 5000ms seems to be enough time anyway. But if there's a better way of doing this, or a more preferable timeout, please let me know.
This may also be something to look into if it affects the other tests as well.

mikeisen1:
I noticed that you didn't use `waitFor` on the other tests in the profile test file. Why not use `waitFor` for `findByText` and `findAllByText` as well?

When testing the code for this branch, I did not see any profile test failures, but still am curious as to the above question.

hreineck:
> I noticed that you didn't use `waitFor` on the other tests in the profile test file. Why not use `waitFor` for `findByText` and `findAllByText` as well?
> 
> When testing the code for this branch, I did not see any profile test failures, but still am curious as to the above question.

Good point. I only changed what was needed to fix the tests, but it would be better if the whole test was consistent. I'll make that change now.

mikeisen1 (approved):
The profile test does not fail and everything seems good to go.

snorouzzadeh (approved):
No test is failing. This looks good to me. Great work!

PR #155, opened 2021-08-17
Pass Control Id into Control Part to Render Modifications

mikeisen1:
Modified the OSCALControl file to add a missing `controlId` prop to ControlPart. This was causing some control modifications to not be displayed and we want all control modifications to be visible to the viewer.

Additionally, test data and the test were modified to test these changes. Previously, the test for displaying control modifications in the `OSCALControlImplementationImplReq.test` test file was not testing anything different than the test for displaying control modifications in the `OSCALProfile.test` test file. There was that redundancy despite `OSCALControlImplementationImplReq.test` using `profileModifyTestData`, created for the purpose of testing whether top-level control modifications would be displayed. That test data, and subsequently the test file, were modified to test if a modifications for a control implementation ending in `_smt` were properly displayed. The distinction of `_smt` is important, as it is a top-level control modification and its appearance in a control's modifications, coupled with the missing `controlId` prop, caused the control modification to not be displayed.

### DIFF
--- a/src/components/OSCALControl.test.js
+++ b/src/components/OSCALControl.test.js
@@ -2,9 +2,23 @@ import React from "react";
 import { render, screen, fireEvent } from "@testing-library/react";
 import OSCALControl from "./OSCALControl";
 import OSCALControlImplementation from "./OSCALControlImplementation";
+import OSCALProfile from "./OSCALProfile";
 import { controlImplTestData, exampleControl } from "../test-data/ControlsData";
 import { exampleComponents } from "../test-data/ComponentsData";
 import { profileModifySmtTestData } from "../test-data/ModificationsData";
+import profileTestData from "../test-data/ProfileData";
+import { parentUrlTestData } from "../test-data/Urls";
+
+async function testIfModificationsDisplayed(modificationName) {
+  const modButton = await screen.findByRole(
+    "button",
+    { name: modificationName },
+    { timeout: 10000 }
+  );
+  fireEvent.click(modButton);
+  expect(await screen.findByText("Modifications")).toBeVisible();
+  expect(await screen.findByText("Adds")).toBeVisible();
+}
 
 test("OSCALCatalog displays controls", () => {
   const testControl = { id: "ac-1" };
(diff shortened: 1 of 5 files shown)


src/test-data/ModificationsData.js
```diff
@@ -42,7 +42,7 @@ export const exampleModificationAltersTopLevel = [
     "control-id": "control-1",
     adds: [
       {
-        "by-id": "control-1",
+        "by-id": "control-1_smt",
```
hreineck, line 45: This object shouldn't be changed. We've been using the term "top-level" to refer to modifications that apply to the control as a whole, not any of the individual control parts.

Changing this data also changes the Storybook; I believe the intention with the modifications in the Storybook page for the Control component was to show off the top-level modifications and how to render them; this change gets rid of that.

hreineck (changes requested):
Thanks for your work on this! The fix should work, however there's a couple issues with the new test:

The changes to the test and test data would get rid of tests and docs for the existing top-level modification feature; see my comment on our use of that term.
I think instead of swapping out our existing top-level modification tests for some different types of modifications, the tests could be expanded to test for all the different situations in which modifications can be rendered:
- Top Level, ie. on the level of the control itself
- Modifications that apply to the "first level" of control parts, ie. modifications that apply to control parts called from the control itself
- Modifications that apply to control parts that were called from other control parts.

Since all of these are called originating from the Control component, this test could probably be made a part of `OSCALControl.test.js`

src/test-data/ModificationsData.js
mikeisen1: That makes sense, will make those changes. I will create another test in `OSCALControl.test` and the appropriate test data in `ModificationsData`

src/components/OSCALControl.test.js
```diff
@@ -1,10 +1,34 @@
...
+import { profileModifySmtTestData } from "../test-data/ModificationsData";
 
 test("OSCALCatalog displays controls", () => {
   const testControl = { id: "ac-1" };
   render(<OSCALControl control={testControl} />);
   const result = screen.getByText("ac-1");
   expect(result).toBeVisible();
 });
+
+test("OSCALControl displays _smt modifications", async () => {
```
hreineck, line 16:
```suggestion
test("OSCALControl displays statement modifications", async () => {
```

Just one nitpick - this can be a tad more descriptive

hreineck (changes requested):
Looks good. Only one minor change.

Also, in reviewing this I came up with some notes about the modification tests.
Seems we're testing modifications in three different places now, and they're a bit redundant:
The Profile test and Control Implementation Impl. Req. test both have a test for top-level modifications. And now the new test that you've added properly tests the modifications that were fixed in this PR.

Since the only place modifications can actually be called or rendered is from the Control component, I think it'd make more sense for all the modification tests to live in the Control test file (like the new one you have added), and have the parent components simply test if they can properly render the controls.

Though cleaning up the tests is probably out of the scope of this issue. Up to you if you want to tack that onto this PR or get this merged first.

mikeisen1:
> Seems we're testing modifications in three different places now, and they're a bit redundant:

I somewhat alluded to this in the comment made when I created this PR. I'm going to reduce the redundancy as part of this PR, as it doesn't need to be it's own ticket. However, I wonder if we should keep one of the other control modifications tests, as the modification control ids are slightly different.

hreineck (approved):
Looks good to me! Thanks for making the changes to the tests.

kylelaker (approved):
Not seeing the message in the logs anymore and these new changes look great! Thank you!

PR #157, opened 2021-08-19
Check for Media Type in Profile Resolver

hreineck:
Modify the profile resolver to check for the `media-type` value
when grabbing a resource from back matter. This is a fix for
an issue where the first rlink in each resource was found even if there
were multiple links. This potentially could have lead to the wrong link
being grabbed and thus failing to load the resource.

### DIFF
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -120,7 +120,7 @@ export default function OSCALBackMatter(props) {
               {resource.rlinks &&
                 resource.rlinks.map((rlink) => (
                   <Chip
-                    key={resource.uuid}
+                    key={rlink.href}
                     label={getMediaType(rlink)}
                     component="a"
                     href={getAbsoluteUrl(rlink, props.parentUrl)}
--- a/src/components/OSCALProfile.test.js
+++ b/src/components/OSCALProfile.test.js
@@ -12,7 +12,6 @@ import testOSCALMetadata from "./OSCALMetadata.test";
 import testOSCALBackMatter from "./OSCALBackMatter.test";
 import { parentUrlTestData } from "../test-data/Urls";
 import profileTestData from "../test-data/ProfileData";
-import getUriFromBackMatterByHref from "./oscal-utils/OSCALBackMatterUtils";
 
 test("OSCALProfile loads", () => {
   render(<OSCALProfileLoader />);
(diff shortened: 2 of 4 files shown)


mikeisen1 (approved):
Everything looks good to me. Great work!

kylelaker (approved):
So we're going with the first `rlink` where the `media-type` is `application/oscal.catalog+json`. Sounds good to me. Will there always be one that meets these constraints? I guess so since we've validated the UUID.

kylelaker (commented):
Oh! Actually could we add a test for this? Since we're "fixing" a bug we want to codify that returning something with this media-type is expected. It also should hopefully be a pretty easy test if we have mock/sample data for this already?

If a test would be _way_ too much effort then we don't need to do it, but this feels like the kind of change we want to validate we don't break later.

hreineck:
> So we're going with the first `rlink` where the `media-type` is `application/oscal.catalog+json`. Sounds good to me. Will there always be one that meets these constraints? I guess so since we've validated the UUID.

I believe it's expected that there's always going to be one; we throw an error if a catalog can't be found for a given href.

hreineck:
> Oh! Actually could we add a test for this? Since we're "fixing" a bug we want to codify that returning something with this media-type is expected. It also should hopefully be a pretty easy test if we have mock/sample data for this already?
> 
> If a test would be _way_ too much effort then we don't need to do it, but this feels like the kind of change we want to validate we don't break later.

Good idea, I'll work on creating a test now.

mikeisen1:
Why are we testing it `getUriFromBackMatterByHref` returns a string? This PR is fixing the bug that just the first rlink is returned no matter what. Shouldn't we instead be testing whether the function returns a string that contains `application/oscal.catalog+json`?

kylelaker:
And I think that specifically we want to test a case where we have a back ref that has various media types with the same UUID and that the result is always the `oscal.catalog+json`

hreineck:
Made some more changes to the tests; now we simply add another rlink to the Back Matter data that is passed into the tests. I think this approach will sufficiently test that all the components work with multiple rlinks.

I also removed the profile resolver test that I added before - I think that test was redundant since any failure in the resolver will throw an error and show up in the tests. But if anyone thinks it should be added back that can be done pretty easily.

mikeisen1 (approved):
I like the changes to the test data and removing the added profile test, which isn't needed. Overall, great work!

kylelaker (approved):

PR #168, opened 2021-09-24
Configure Publishing OSCAL Viewer to GitHub Packages

hreineck:
Set up package.json and GitHub actions to publish to GitHub packages.

### DIFF
--- a/.github/workflows/gh-packages.yml
+++ b/.github/workflows/gh-packages.yml
@@ -1,0 +1,25 @@
+name: Publish to GitHub Packages
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  publish:
+    name: Publish OSCAL Viewer Build to GitHub Packages
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup NodeJS
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          registry-url: https://npm.pkg.github.com/
+      - name: Update Npm
+        run: npm install -g npm@latest
+      - name: Build and Publish
+        run:
+          cd example && npm ci && npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/gh-packages.yml
+++ b/.github/workflows/gh-packages.yml
@@ -8,6 +8,9 @@ jobs:
   publish:
     name: Publish OSCAL Viewer Build to GitHub Packages
     runs-on: ubuntu-20.04
+    permissions:
+      contents: read
+      packages: write
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/.github/workflows/gh-packages.yml
+++ b/.github/workflows/gh-packages.yml
@@ -22,7 +22,12 @@ jobs:
           cache: npm
       - name: Update Npm
         run: npm install -g npm@latest
-      - name: Build and Publish
+      - name: Build and Publish Library
+        run:
+          npm ci && npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      - name: Build and Publish OSCAL Viewer
         run:
           cd example && npm ci && npm publish
         env:
--- a/.github/workflows/gh-packages.yml
+++ b/.github/workflows/gh-packages.yml
@@ -19,6 +19,7 @@ jobs:
         with:
           node-version: 14
           registry-url: https://npm.pkg.github.com/
+          cache: npm
       - name: Update Npm
         run: npm install -g npm@latest
       - name: Build and Publish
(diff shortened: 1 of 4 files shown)


hreineck:
This and [the corresponding PR in oscal-rest-service](https://github.com/EasyDynamics/oscal-rest-service/pull/18) handle setting up GitHub packages as part of the all-in-one deployment.
Once the changes to `package.json` are made, the package can be published by simply running `npm publish` with necessary credentials - but I figured a formalized GitHub actions would be better and more reviewable.

example/package.json
```diff
@@ -2,7 +2,10 @@
   "name": "oscal-viewer",
   "homepage": ".",
   "version": "0.1.0",
-  "private": true,
```
hreineck, line 5: Not sure about this change. I thought it needed to be removed in order to publish - can someone clarify what this line was doing or if it's still necessary?

example/package.json
```diff
@@ -2,7 +2,10 @@
...
+  "repository": "git://github.com/EasyDynamics/oscal-react-library.git",
```
kylelaker, line 5: We probably don't want to use the `git://` protocol. So based on [`package.json` docs for `repository`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository) we probably also need to specify `directory`.
```suggestion
  "repository": {
    "type": "git",
    "url": "https://github.com/EasyDynamics/oscal-react-library.git",
    "directory": "example"
  },
```

The reason I'd recommend against using `git://` is that it's a little antiquated and is read-only (at least on GitHub). For more information about this protocol and others, see https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_git_protocol. `https` is a good compromise because for GitHub, no auth is required for reads (but is for pushes), it works in a web browser as well as the git CLI, and it supports encryption in transit. So it's cool that our

kylelaker (changes requested):
I have a few questions here since we're talking about the example application (not library). I think all of this would be absolutely phenomenal for the React Library at the root but we may want to think about the design a bit for the example app.

1. Is it really appropriate to deploy the example _app_ to GitHub Packages? It's more of a static asset after an `npm run build` and we can just deploy the `examples/build` folder directly. Do we really care about the full contents of the NPM package and do those provide us any benefit anyway? Perhaps we'd be better served by `actions/upload-artifact` rather than by publishing. Will it provide any benefit inside a `node_modules/` folder? I believe we're actually already pushing that as an artifact: https://github.com/EasyDynamics/oscal-react-library/actions/runs/1174249533.
2. If we're going to go to this effort to publish the example app, should we configure the library to be published to GitHub Packages while we're setting this workflow up? We can always switch to a "real"/default registry like NPM's later but this would let us have a place to pull down snapshot versions of the library.

.github/workflows/gh-pages.yml
```diff
@@ -8,6 +8,9 @@ jobs:
   publish:
     name: Publish Storybook to GitHub Pages
     runs-on: ubuntu-20.04
+    permissions:
+      contents: read
+      packages: write
```
kylelaker, line 13: This particular action probably doesn't need to write to `packages`. But I think the other workflow, `gh-packages.yml` will

.github/workflows/gh-packages.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: Publish OSCAL Viewer Build to GitHub Packages
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup NodeJS
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          registry-url: https://npm.pkg.github.com/
```
kylelaker, line 18:
```suggestion
          node-version: 14
          registry-url: https://npm.pkg.github.com/
          cache: npm
```

hreineck:
> I have a few questions here since we're talking about the example application (not library). I think all of this would be absolutely phenomenal for the React Library at the root but we may want to think about the design a bit for the example app.
> 
> 1. Is it really appropriate to deploy the example _app_ to GitHub Packages? It's more of a static asset after an `npm run build` and we can just deploy the `examples/build` folder directly. Do we really care about the full contents of the NPM package and do those provide us any benefit anyway? Perhaps we'd be better served by `actions/upload-artifact` rather than by publishing. Will it provide any benefit inside a `node_modules/` folder? I believe we're actually already pushing that as an artifact: https://github.com/EasyDynamics/oscal-react-library/actions/runs/1174249533.

We don't really care about anything but the static output of `build`, but GitHub Packages may still be a good option since we will eventually be setting up a pipeline to deploy the library to a registry anyway.
If there's nothing explicitly wrong with publishing the entire app, I think it may be simpler to keep using the current approach;
the `upload-artifacts` action seems to be geared more towards sharing artifacts within a single worklow. I could only find this third-party action: https://github.com/dawidd6/action-download-artifact to be able to download from one repo to another.
Would like to know what the team thinks of this - if the approach using artifacts seems better we can certainly use that instead.

> 2. If we're going to go to this effort to publish the example app, should we configure the library to be published to GitHub Packages while we're setting this workflow up? We can always switch to a "real"/default registry like NPM's later but this would let us have a place to pull down snapshot versions of the library.

Good call.

mikeisen1 (approved):
Everything looks good to go. Nice work!

kylelaker (approved):

PR #172, opened 2021-09-30
Switch to `npm ci` in the pipeline to avoid failures

mikeisen1:
Replace `npm install` in the pipeline with `npm ci`. We are doing this to avoid pipeline failures caused by `npm install` modifying `package-lock.json`. `npm ci` will not modify that file and it also provides more detailed messaging when a dependency error arises during a workflow run.


README.md
```diff
@@ -29,7 +29,7 @@ The [example](./example) folder shows how those components can be assembled to c
 ### Building and Running Locally
 
 In order to use the library components or test the library locally, the project components need to be rolled up and
-dependencies need to be installed initially with the command: `npm install`. From here the application can be built
```
rgauss, line 32: I don't think we want to change this for developers building locally, just the pipeline(s)

package.json
```diff
@@ -31,13 +31,13 @@
...
     "test": "react-scripts test",
-    "predeploy": "cd example && npm install && npm run build",
+    "predeploy": "cd example && npm ci && npm run build",
     "eject": "react-scripts eject",
     "lint": "npx eslint src/**/*.js --fix",
     "format": "prettier --write src/**/*.js",
     "deploy": "gh-pages -d oscal-react-library/build",
-    "build-library-and-run-example": "npm install && cd example && npm install && npm start",
-    "build-and-run-example": "cd example && npm install && npm start",
+    "build-library-and-run-example": "npm ci && cd example && npm ci && npm start",
```
rgauss, line 39: Here as well, these are used by developers so I don't think they'd be changing.

rgauss (changes requested):

hreineck (approved):

rgauss (approved):

kylelaker (approved):

PR #179, opened 2021-10-12
Include Production Build of Viewer in Package

hreineck:
Modify package.json to reduce the unnecessary inclusion of source code
in the OSCAL Viewer GitHub Package and instead include the built
production build, which will be used by the pipeline and the end-user.


hreineck:
The output of `npx npm-packlist` shows the files that will be included in the package:
image

PR #182, opened 2021-10-18
 Run Build in Packages Workflow

hreineck:
Add the `npm run build` command to the gh-packages workflow to add the production build to the oscal viewer package.


hreineck:
The previous [run](https://github.com/EasyDynamics/oscal-react-library/actions/runs/1355776120) of the workflow on the commit changing the branch trigger to this branch shows that the workflow now successfully publishes the production build, found in a `build` folder when the user installs the `oscal-viewer` package.

PR #184, opened 2021-10-21
Move OSCAL Viewer Package to GitHub Releases

hreineck:
Move the production build of the Oscal Viewer, previously contained in a
separate `oscal-viewer` package, into the `oscal-react-library` package,
and remove the viewer package.

This fixes issues with the oscal-viewer package, which was previously
set up only for offline development by using local paths in its dependencies,
which are not compatible with publishing to an online registry.
Moving everything into one package allows for the end user to install both the
Viewer and its dependencies all in one place.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
           cd example
           npm ci
           npm run build
-          zip viewer.zip build
+          zip viewer.zip build/*
 
       - name: Create a GitHub Release
         uses: softprops/action-gh-release@v1
@@ -79,9 +79,9 @@ jobs:
       # Should only run when triggered for a feature - for bug fixes
       # this should be done manually since a merge might be needed.
       - name: Push Changes
-        if: ${{ github.event.inputs.type  == 'feature' }}
+        #if: ${{ github.event.inputs.type  == 'feature' }}
         run: |
-          git push origin main
+#git push origin main
           git push -u origin tag-${{ env.PKG_VERSION }}
           npm version preminor
           cd example
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,7 @@
 name: Create a New Release
 on:
+  push:
+    branches: [EGRC-492-package]
   workflow_dispatch:
     inputs:
       release-type:
@@ -46,7 +48,7 @@ jobs:
       # Increment the version to the next bug fix
       # Push the new version to the develop branch
       - name: Set Version (Bug Fix)
-        if: ${{ github.event.inputs.release-type == 'bug' }}
+        #if: ${{ github.event.inputs.release-type == 'bug' }}
         run: |
           npm version patch
           cd example
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,6 +36,8 @@ jobs:
       - name: Set Version (Feature)
         if: ${{ github.event.inputs.type  == 'feature' }}
         run: |
+          git config --global user.email "info@easydynamics.com"
+          git config --global user.name "${{ github.actor }}"
           npm version minor
           cd example
           npm version minor
@@ -45,6 +47,8 @@ jobs:
       - name: Set Version (Bug Fix)
         # if: ${{ github.event.inputs.type == 'bug' }}
         run: |
+          git config --global user.email "info@easydynamics.com"
+          git config --global user.name "${{ github.actor }}"
           npm version patch
           cd example
           npm version patch
@@ -52,7 +56,9 @@ jobs:
 
       # Output the new package version to environment variable
       - name: Get Package Version
-        run: echo "PKG_VERSION=$(npm pkg get version | tr -d '"')" >> $GITHUB_ENV
+        run: |
+          echo "PKG_VERSION=$(npm pkg get version | tr -d '"')" >> $GITHUB_ENV
+          echo ${{ env.PKG_VERSION }}
 
       - name: Build OSCAL Viewer
         working-directory: example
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -24,20 +20,34 @@ jobs:
       - name: Update Npm
         run: npm install -g npm@latest
 
+      # Increment version to next major version.
+      # Store version as an environment variable for later steps to use.
+      - name: Set Version of this Release
+        run: |
+          npm version major
+          cd example
+          npm version major
```
mikeisen1, line 29: According to the aforementioned diagram, we want to increase the second number, i.e. from 1.0.1 to 1.1.0. Doing `npm version major` would increase it from 1.0.1 to 2.0.0. `npm version minor` accomplishes what we want per the diagram.

mikeisen1:
Don't we also want to do a preminor increment of the develop branch when creating a new feature? If we wanted to do so in this workflow we could use [this](https://github.com/actions/checkout#Checkout-a-different-branch) as guidance for checking out and then pushing changes.

mikeisen1:
Are we assuming this workflow is always run from the develop or main branch or could it be run from any branch?

hreineck:
> Are we assuming this workflow is always run from the develop or main branch or could it be run from any branch?

Good point - this should always check out the develop branch and make changes to that first. I'll look into specifying that so when this is triggered manually it always checks out the right branch.

.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
...
+            example/viewer.zip
+
+      # Commit the new version and push changes to main, as well as a "tag-{version}" branch
+      # And push a new preminor version to develop.
+      # Should only run when triggered for a feature - for bug fixes
+      # this should be done manually since a merge might be needed.
+      - name: Push Changes
+        if: ${{ github.event.inputs.type }} == 'feature'
+        run: |
+          git push origin main
```
mikeisen1, line 70:
```suggestion
          git push origin develop:main
```
We are still on the develop branch here. If we want to push to a different branch, main in this case, then we must do the above command. In `git push origin develop:main`, `develop` specifies the source (or current) branch we are on while `main` is the different branch that we want to push to.

mikeisen1 (changes requested):

.github/workflows/release.yml
hreineck: I don't think we are necessarily pushing from "develop", we are checking out develop, making changes, and then pushing to several remote branches.

mikeisen1: Looking at this image, it appears the workflow can be run from develop (or any branch for that matter, I just assumed we would require this branch to be run from develop).

run-publish-workflow

hreineck:
The latest run of the workflow successfully created a [release](https://github.com/EasyDynamics/oscal-react-library/releases/tag/0.2.0) with the OSCAL Viewer as a .zip that's downloadable via the GitHub API.

rgauss:
> The latest run of the workflow successfully created a [release](https://github.com/EasyDynamics/oscal-react-library/releases/tag/0.2.0) with the OSCAL Viewer as a .zip that's downloadable via the GitHub API.

@hreineck, unfortunately that `viewer.zip` contains only an empty `build` folder.

hreineck:
The last run of the workflow created two new branches, per the [diagram](https://easydynamics.sharepoint.com/sites/msd/SitePages/Versioning.aspx): `tag-0.2.0`, which is the new tag branch created after incrementing the version (in this case from `0.2.0-0` to `0.2.0`), and is the same as what would be pushed to `main` when this is run for real; and `test-0.2.0`, which is confusingly named but represents the changes that would be pushed to `develop` as the version is incremented to the next preminor (in this case from `0.2.0` to `0.3.0-0`)

rgauss:
> The last run of the workflow created two new branches, per the [diagram](https://easydynamics.sharepoint.com/sites/msd/SitePages/Versioning.aspx): `tag-0.2.0`, which is the new tag branch created after incrementing the version (in this case from `0.2.0-0` to `0.2.0`), and is the same as what would be pushed to `main` when this is run for real; and `test-0.2.0`, which is confusingly named but represents the changes that would be pushed to `develop` as the version is incremented to the next preminor (in this case from `0.2.0` to `0.3.0-0`)

@hreineck, to be clear, we want a `0.2.0` tag, i.e. visible in [tags](https://github.com/EasyDynamics/oscal-react-library/tags) (which seems to be present from a previous run), not necessarily a branch named `tag-0.2.0`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
+name: Create a New Release
+on:
+  workflow_dispatch:
+    inputs:
+      type:
+        description:
+          'The version type of this release. Must be either "feature" or "bug", representing a new feature release or a bug fix release, respectively,'
+        required: true
```
kylelaker, line 8:
```suggestion
      type:
        description: The type of release being performed
        required: true
        type: choice
        options: [feature, bug]
```

https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

kylelaker: Note that you'll have to change all the `type` references if you go with the new name here

mikeisen1:
Everything looks good! However, I have not personally seen a release. Could you push a new release (and delete it once I've seen it) so I can approve this PR?

hreineck:
> Everything looks good! However, I have not personally seen a release. Could you push a new release (and delete it once I've seen it) so I can approve this PR?

Just pushed a [release.](https://github.com/EasyDynamics/oscal-react-library/releases/tag/0.2.0)

mikeisen1 (approved):
Everything looks great!

rgauss (approved):

PR #186, opened 2021-11-15
Show profile and catalog inheritance

### LINKED ISSUE #161: Show Profile/Catalog Inheritance
As a user involved in the governance, risk, and compliance of systems and components, I want to see the tree of Profile/Catalog inheritance for any OSCAL objects that reference/import Profiles so that I can understand how controls are being resolved.

### Acceptance Criteria
* SSPs show the expanded tree of referenced Profiles/Catalogs titles after the metadata section
* Profiles show the expanded tree of referenced Profiles/Catalogs titles after the metadata section
* Component Definitions show the expanded tree of referenced Profiles/Catalogs titles after the metadata section
* The display of information should distinguish between Profiles and Catalogs in some way

### Implementation Notes
* We can likely use the same React component across SSPs, Profiles, and Component Definitions
* We can likely use the [Material-UI TreeView component](https://material-ui.com/components/tree-view/) as shown below for the example SSP
* We could consider icons for distinguishing between Profiles and Catalogs, or simply use a prefix, i.e. "Profile: "
* We could consider "Profiles/Catalog Inheritance" as the header name for this section
* We are not displaying [profile resolution details](https://pages.nist.gov/OSCAL/concepts/processing/profile-resolution/) at this time
* In the future we will address the resolution path of individual controls, but not in this iteration

Screen Shot 2021-08-27 at 8 31 05 AM

mikeisen1:
Create a react component, `OSCALProfileCatalogInheritance`, which displays the tree of inherited profiles and catalogs for any Component Definition, Profile, or SSP. This allows us to view that inheritance tree, providing users greater insight into which profiles and catalogs were used in the creation of a given OSCAL document.

Additionally created a storybook page for the new component so any developer can view how it looks and functions and potentially reuse it in their code.

closes #161


.eslintrc.json
```diff
@@ -25,7 +25,8 @@
       "react/prop-types": [0],
       "react/destructuring-assignment": [0],
       "react/jsx-props-no-spreading": "off",
-      "prettier/prettier": "error"
+      "prettier/prettier": "error",
+      "no-param-reassign": 0
```
kylelaker, line 29: I don't love the addition of this. Why don't we just use a different name for the variable inside the function rather than reusing a parameter name? Shadowing variables is always a little iffy.

mikeisen1:
Here are some of the other options for the expand/collapse icons. Just wanted to put them here in case anyone thought one of these could be better than the ones currently used.

expand icon: Add, ControlPoint, MoreHoriz, MoreVert, UnfoldMore, ExpandMore

collapse icon: Remove, Unfold Less, ExpandLess, Close, Cancel, HighlightOff, Remove

You can easily view these visually [here](https://mui.com/components/material-icons/) by searching for the above icons in the search box.

src/components/oscal-utils/OSCALProfileResolver.js
```diff
@@ -27,17 +27,19 @@ import getUriFromBackMatterByHref from "./OSCALBackMatterUtils";
  * @see {@link https://pages.nist.gov/OSCAL/documentation/schema/profile-layer/profile/xml-schema/#global_import}
  * @see {@link https://pages.nist.gov/OSCAL/documentation/schema/profile-layer/profile/xml-schema/#global_back-matter_h2}
  */
-export default function OSCALResolveProfileOrCatalogUrlControls(
+export default async function OSCALResolveProfileOrCatalogUrlControls(
```
rgauss, line 30: Is there a reason we're introducing asyncs and awaits for this?

rgauss (changes requested):

src/components/oscal-utils/OSCALProfileResolver.js
mikeisen1: There was some issue when I was first doing the task that `async`/`await` seemed to resolve. It isn't needed I just forgot to remove them from the code.

src/components/OSCALProfileCatalogInheritance.js
```diff
@@ -21,26 +21,31 @@ function renderTree(nodes, id) {
   const generateLabel = (title, type) =>
     type === "profile" ? `Profile: ${title}` : `Catalog: ${title}`;
 
+  const idReassign = id;
   const children = [];
 
   nodes.inherited.forEach((node) => {
-    id[0] += 1;
+    idReassign[0] += 1;
```
kylelaker: We're not really short on characters here. How about `inheritedNodeId` or just `nodeId` to give it a more descriptive name?

src/components/OSCALProfileCatalogInheritance.js
```diff
@@ -0,0 +1,81 @@
...
+  const idReassign = id;
+  const children = [];
+
+  nodes.inherited.forEach((node) => {
+    idReassign[0] += 1;
+    children.push(
+      <TreeItem
+        nodeId={idReassign[0].toString()}
+        label={generateLabel(node.title, node.type)}
+        expandIcon={
```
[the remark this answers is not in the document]
mikeisen1: I chose not to because if we use the `TreeView` default icon properties, then all icons will have the same label. That won't go well for testing purposes so if we want to be able to test the expansion and collapse of the tree, then we can't use the `TreeView` defaults. If we don't care about that, then the defaults are fine.

src/components/OSCALProfileCatalogInheritance.test.js
```diff
@@ -0,0 +1,34 @@
...
+  const result = await waitFor(
+    () => screen.getByText("Catalog: Example Catalog"),
+    { timeout: 5000 }
+  );
+  expect(result).toBeVisible();
+});
+
+test("OSCALProfileCatalogInheritance displays nested inherited objects", async () => {
+  profileCatalogInheritanceRenderer();
+  const inheritanceButton = await screen.findByRole(
```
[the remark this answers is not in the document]
mikeisen1: It seems like it would make sense to at least test the expansion and collapse of the tree to ensure everything works and is displayed as expected in both cases.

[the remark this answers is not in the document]
mikeisen1: But this test can be rewritten once everything is expanded by default.

src/components/OSCALProfileCatalogInheritance.js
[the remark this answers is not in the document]
rgauss: Ideally we don't want to be required to rely on aria labels, instead taking an approach of finding the expected text then the closest button.

As mentioned elsewhere on this PR, we also want the tree to be expanded by default, so I'm not sure will need to test clicking on these at all.

src/components/OSCALProfileCatalogInheritance.test.js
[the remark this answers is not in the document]
rgauss: OK, that's fine.

src/components/OSCALProfileCatalogInheritance.js
rgauss: Another thought here would be to add the UUID of the OSCAL document to the inheritance object, but we also talk about the possibility of using default icons at the `TreeView` level, so this ID may not be needed at all.

src/components/OSCALComponentDefinition.js
```diff
@@ -17,24 +18,27 @@ const useStyles = makeStyles((theme) => ({
...
+        () => {
+          setIsLoaded(true);
+          props.onResolutionComplete();
+        },
+        (errorReturned) => {
+          setError(errorReturned);
+          setIsLoaded(true);
+          props.onResolutionComplete();
+        }
+      )
```
hreineck, line 39: This is getting a bit convoluted. Instead of having a method that takes in the return of the existing `OSCALComponentResolveSources` method, could we have an overall "resolve" wrapper method like the other OSCAL types do?

src/components/OSCALProfileCatalogInheritanceTree.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import { TreeView } from "@material-ui/lab";
+import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
+import ExpandLessIcon from "@material-ui/icons/ExpandLess";
+import { Grid, IconButton, List, Paper } from "@material-ui/core";
+import ListSubheader from "@material-ui/core/ListSubheader";
+
+export default function OSCALProfileCatalogInheritanceTree(props) {
+  const tree =
```
hreineck, line 9: Is the `const` necessary? Why not just `return` this object?

src/components/OSCALProfileCatalogInheritanceTree.js
```diff
@@ -0,0 +1,45 @@
...
+                <IconButton>
+                  <ExpandMoreIcon />
+                </IconButton>
+              }
+              defaultCollapseIcon={
+                <IconButton>
+                  <ExpandLessIcon />
+                </IconButton>
+              }
+              defaultExpanded={props.validIds}
```
hreineck, line 35: Just putting this out here - is the ability to actually collapse the tree necessary? If it makes things simpler to have it permanently expanded that could work as well.

hreineck (changes requested):
Looks good. Both the Viewer and Storybook are working locally for me.
Just a few small changes/questions.

src/components/OSCALProfileCatalogInheritanceTree.js
mikeisen1: I'm not sure that it's absolutely necessary but if we can allow users to collapse the tree we should provide it as it's something, albeit small, that could improve user experience.

hreineck (approved):
Great work!

src/components/OSCALProfileCatalogInheritance.js
```diff
@@ -0,0 +1,61 @@
...
+      </TreeItem>
+    );
+  });
+
+  return children;
+}
+
+export default function OSCALProfileCatalogInheritance(props) {
+  const classes = useStyles();
+  const [validIds, setValidIds] = useState([false, []]);
```
rgauss, line 45: Why do we need all of these instances of IDs separately rather than just using the UUID already in the inherited objects?

rgauss (changes requested):

src/components/OSCALProfileCatalogInheritance.js
mikeisen1: How this is answered will impact what exact changes are made:

Do we want to ensure each nodeId in the tree is unique? We could have the same profile or catalog in different parts of the tree which could cause repeated nodeIds.

rgauss: If we're setting node IDs and keys we do want them to be unique.

Are we sure we need `useState` here?

I'd also recommend changing these to something like `expandedIds` everywhere.

src/components/OSCALProfileCatalogInheritance.js
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from "react";
```
rgauss, line 1: Does this and `OSCALProfileCatalogInheritanceTree` need to be separate classes or can they be merged into a single component?

rgauss (changes requested):

src/components/OSCALProfileCatalogInheritance.js
mikeisen1: We need `useState` so when we set the expanded ids, we re-render the component. If we do not re-render, the tree will not be expanded by default.

src/components/OSCALProfileCatalogInheritance.js
mikeisen1: So for `TreeView`, there are two props where we can set the default expanded nodes. The first is `expanded`. With this prop, we do not need a second React component. However, once the tree is expanded, it cannot be collapsed. The second prop, the one currently being used, is `defaultExpanded`. With this prop, we need a second React component but users can collapse the tree after it is initially expanded (I couldn't figure out a way to not require a second component for this). I figured that while the ability to collapse is not a major feature, some users might want that ability especially if the tree is large.

rgauss: I'm not sure I understand why the use of `defaultExpanded` requires a second component.

mikeisen1: @rgauss , I fixed the issue. 

When I was working on the issue and initially created the `validIds` state, the tree was not initially being expanded. So when I moved to creating the second React component and everything worked as desired, I didn't realize how the current use of `validIds` would let me avoid using the second component, especially being that the props passed in to `OSCALProfileCatalogInheritance` are initially empty. It was simply just an issue (with regards to not requiring a second component) of the condition on which this component is re-rendered.

src/components/OSCALComponentDefinition.js
```diff
@@ -17,12 +18,15 @@ const useStyles = makeStyles((theme) => ({
...
   const [isLoaded, setIsLoaded] = useState(false);
+  const [inheritedProfilesAndCatalogs, setInheritedProfilesAndCatalogs] =
+    useState({});
   const classes = useStyles();
 
   useEffect(() => {
     OSCALComponentResolveSources(
       props.componentDefinition,
       props.parentUrl,
+      setInheritedProfilesAndCatalogs,
```
rgauss, line 29: It seems a bit odd to send in essentially another `onSuccess` function here.  Should we consider simply adding some form of `inheritedProfilesAndCatalogs` argument to the existing `onSuccess` function and call `setInheritedProfilesAndCatalogs` within that?

rgauss (changes requested):

src/components/oscal-utils/OSCALProfileResolver.js
```diff
@@ -145,11 +139,11 @@ export function OSCALResolveProfile(
       parentUrl,
       profile["back-matter"],
       inheritedProfilesAndCatalogs.inherited,
-      onSuccess,
+      () => {
```
rgauss, line 142: Is additional wrapping function needed? Can we not just call `onSuccess(inheritedProfilesAndCatalogs)` directly?

mikeisen1: If we just eliminate the wrapping functions, then we would need to manually re-render the page.

rgauss: Could you elaborate on why that's the case?

rgauss: Disregard, I see what's going on there.

rgauss (changes requested):
Two minor rendering changes:
1) I think the tree needs more of a left margin.  The expand arrow currently bumps up against the card edge.
2) The header is sticky rather than scrolling with the page

src/components/oscal-utils/OSCALProfileResolver.js
mikeisen1: When we don't use a wrapper function, JavaScript attempts to evaluate `onSuccess(inheritedProfilesAndCatalogs)` to a value and no React state is modified inside `OSCALResolveProfileOrCatalogUrlControls`. But when we do use a wrapper function, JavaScript simply evaluates it to a function object, which can easily be called to perform the necessary actions and modify a React state.

Manually re-rendering the page works because we modify `inheritedProfilesAndCatalogs.inherited` inside the function, which will eventually provide `OSCALProfileCatalogInheritance` with the necessary data.

rgauss (approved):
Looks great, thanks for working through all the changes.

PR #200, opened 2021-12-23
Edit SSP Details (UI)

mikeisen1:
Closes EasyDynamics/easygrc#8

Provide the user interface for editing the metadata title and version for an OSCAL Ssp. This also sends an HTTP PATCH request to a backend service and handles the response. This allows someone to, with the backend functionality provided, modify an SSP using the OSCAL Viewer. If the SSP is successfully updated, then the desired changes are reflected on the viewer. If the SSP is not successfully updated, the desired changes are not reflected on the viewer.

For the HTTP request, it is assumed for now that the window url is of the form `http://localhost:8080/system-security-plans?url=http://localhost:8080/oscal/v1/ssps/{uuid}`, as when the Viewer is in REST mode, we will not have the reload form and when we select a file to view, we do so by using the url parameter.

If the PATCH fails, an alert will be sent to the user that the update could not be made.

Changes were mostly made to the metadata component file, so not much needs to be done to display the UI for modifying metadata of component definitions, catalogs, and profiles.


mikeisen1:
Currently there was no update to the storybook or tests. I wanted to create the draft PR so comments could at least start coming in as needed.

src/components/OSCALMetadata.js
```diff
@@ -33,8 +37,131 @@ const useStyles = makeStyles((theme) => ({
...
+    .then((res) => res.json())
+    .then(
+      (result) => {
+        /* If we reach this point, then the OSCAL file has successfully been modified
+         * and we want those changes to be visible to the user.
+         */
+        modifiableMetadata["last-modified"][1](
+          formatDate(result["last-modified"])
+        );
+        modifiableMetadata[modifiedField].value[1](newValue);
```
kylelaker, line 71: I may have missed this in the API spec, but I do want to highlight that this relies on the assumption that the only modified values returned are the (single) field we sent and `last-modified`; and that the field we send isn't a date. That may be fine for now.

What we probably want to do is move the logic for how to display/format/represent as far "down" as we can.

src/components/OSCALMetadata.js
```diff
@@ -33,8 +37,131 @@ const useStyles = makeStyles((theme) => ({
...
+  ) : (
+    <Typography variant={modifiedField === "title" ? "h6" : "body2"}>
+      {modifiableMetadata[modifiedField].value[0]}
+    </Typography>
+  );
+}
+
+function displayEditIcons(modifiedField, modifiableMetadata) {
+  const edit = modifiableMetadata[modifiedField].edit[0];
+  const setEdit = modifiableMetadata[modifiedField].edit[1];
```
kylelaker, line 101: Can we do:

```js
const [edit, setEdit] = modifiableMetadata[modifiedField].edit;
```

I am actually asking if this ES6 feature is something we support; not just blindly suggesting we do it :)

kylelaker (changes requested):
This is off to a really good start! Great work on getting the initial `PATCH` call and mutability in here. I tried to catch as much as I could; I understand this is a draft so no worries if you were already tracking a few of these.

Biggest things are:
 - the safer handling of query string params
 - more secure way to determine the server-side URL

src/components/OSCALMetadata.js
```diff
@@ -33,8 +37,131 @@ const useStyles = makeStyles((theme) => ({
 // Returns a string with a locality-sensitive representation of this date
 const formatDate = (isoDate) => new Date(isoDate).toLocaleString();
 
+function saveModifiedMetadata(modifiedField, modifiableMetadata, newValue) {
+  /* Since the browser url will be of the form
+   * http://localhost:8080/system-security-plans?url=http://localhost:8080/oscal/v1/ssps/{uuid}
+   * and we just want the URL defined in the url parameter of the browser url, we must find the
+   * location of the ? and increment that value by 5 to get the start of the desired url.
+   */
+  const startOfUrl = window.location.href.indexOf("?") + 5;
```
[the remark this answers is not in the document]
rgauss: Note that this will drastically change with EasyDynamics/easygrc#6, so we shouldn't spend too much time on it in this PR.

src/components/OSCALMetadata.js
```diff
@@ -33,8 +37,131 @@ const useStyles = makeStyles((theme) => ({
...
+   * updating a value.
+   *
+   * We also provide a body so we can specify what needs to be modified.
+   */
+  fetch(url, {
+    method: "PATCH",
+    body: JSON.stringify({
+      metadata: {
+        modifiedField: newValue,
+      },
```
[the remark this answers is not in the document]
rgauss: Yes, partial updates are expected.  From the back-end story:
> As a developer building apps against the oscal-rest-service I want to be able to update only specific elements of an existing SSP via REST requests so that I can incrementally edit the SSP without transmitting the entire SSP on every call.

src/components/OSCALMetadata.js
rgauss: The response of the PATCH request is the entire object.  The incoming changes are merged with the existing, the object is persisted, re-fetched from persistence, then returned in the response.

@mikeisen1, the back-end is also responsible for updating the `last-modified` field.

Things will be a bit more complicated here because the top-level component, i.e. `OSCALSsp` should re-render the entire SSP given in the PATCH response, but this metadata component is currently handling the request.

src/components/OSCALMetadata.js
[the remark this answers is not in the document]
rgauss: Also note that the request body needs to maintain the entire JSON model even if only one field is being updated.  If this is for an SSP, `metadata` needs to be within a `system-security-plan` structure.

See the back-end tests for an [example](https://github.com/EasyDynamics/oscal-rest-service/pull/34/files#diff-0fd71ac2e229983b4722ce9d27453b9cbad70e55d56985d1f59ed5f49f115b5eR61).

src/components/OSCALMetadata.js
```diff
@@ -33,8 +37,131 @@ const useStyles = makeStyles((theme) => ({
...
+      </TextField>
+    </Typography>
+  ) : (
+    <Typography variant={modifiedField === "title" ? "h6" : "body2"}>
+      {modifiableMetadata[modifiedField].value[0]}
+    </Typography>
+  );
+}
+
+function displayEditIcons(modifiedField, modifiableMetadata) {
```
rgauss, line 99: This feels like it should be its own lower-level component that is only responsible for telling some other component to enter into editing mode.

rgauss (changes requested):
While we're only targeting a few metadata fields in this particular issue/PR, we want this to form the foundation for edits throughout the app and the current approach seems like it will lead to a lot of duplication of code as we expand the editing capabilities.

As mentioned in a few of the individual comments, we should consider moving some of this functionality into smaller, lower-level components.

We may also want to consider an approach where child components (like `OSCALMetadata`) signal the top-level component to handle the `PATCH` request, since the entire top-level component should be re-rendered after a successful update.  That could take the form of passing something like an `onSave` function down to child components, which may also need some form of JSON path so that the top-level component can properly construct the partial update request.

mikeisen1:
> While we're only targeting a few metadata fields in this particular issue/PR, we want this to form the foundation for edits throughout the app and the current approach seems like it will lead to a lot of duplication of code as we expand the editing capabilities.
> 
> As mentioned in a few of the individual comments, we should consider moving some of this functionality into smaller, lower-level components.

Could we also do this for the editable text field?

mikeisen1:
> We may also want to consider an approach where child components (like `OSCALMetadata`) signal the top-level component to handle the `PATCH` request, since the entire top-level component should be re-rendered after a successful update. That could take the form of passing something like an `onSave` function down to child components, which may also need some form of JSON path so that the top-level component can properly construct the partial update request.

I like this approach because in using it, we can resolve the potential security issues with how we generate the PATCH request url.

src/components/OSCALMetadata.js
```diff
@@ -33,8 +37,131 @@ const useStyles = makeStyles((theme) => ({
...
+      onClick={() => {
+        setEdit(!edit);
+      }}
+    >
+      <EditIcon fontSize="small" />
+    </IconButton>
+  );
+}
+
 export default function OSCALControlGuidance(props) {
```
mikeisen1, line 133: Is there any reason this is named `OSCALControlGuidance` and not `OSCALMetadata`?

rgauss: Looks like a leftover from a copy/paste to me.

rgauss:
> > While we're only targeting a few metadata fields in this particular issue/PR, we want this to form the foundation for edits throughout the app and the current approach seems like it will lead to a lot of duplication of code as we expand the editing capabilities.
> > As mentioned in a few of the individual comments, we should consider moving some of this functionality into smaller, lower-level components.
> 
> Could we also do this for the editable text field?

I would think so.

We may need some form of variants for different sizes, etc.

src/stories/OSCALEditableTextField.stories.js
```diff
@@ -0,0 +1,27 @@
...
+  testModifiableMetadata,
+  textFieldEditModeMetadata,
+} from "../test-data/MetadataData";
+
+export default {
+  title: "Components/Editable Text Field",
+  component: OSCALEditableTextField,
+};
+
+const Template = (args) => <OSCALEditableTextField {...args} />;
```
kylelaker, line 13: In #209, the linter was updated. Functions now have to be declared using the `function foo() { ... }` format, per the lint rules.

kylelaker (commented):
Note and context on linter failures

src/components/OSCALMetadata.js
```diff
@@ -33,12 +35,53 @@ const useStyles = makeStyles((theme) => ({
...
 const formatDate = (isoDate) => new Date(isoDate).toLocaleString();
 
-export default function OSCALControlGuidance(props) {
+export default function OSCALMetadata(props) {
   const classes = useStyles();
+
+  /* A JSON formatted variable so only one variable needs to be passed around between
+   *  functions that deals with the modification of metadata fields.
+   *
+   *  For each top-level field, we have a value sub-field which is an array of two
```
rgauss, line 44: This is a bit confusing.  We may need to revisit this approach after addressing other changes.

src/components/OSCALModificationIcons.js
```diff
@@ -0,0 +1,56 @@
...
+function getIconButtons(props) {
+  return props.modifiableData.edit[0] ? (
+    <>
+      <IconButton
+        aria-label={`cancel-${getElementLabel(props.editedField)}`}
+        onClick={() => {
+          props.modifiableData.edit[1](!props.modifiableData.edit[0]);
+        }}
+      >
+        <CloseIcon fontSize={props.iconFontSize} />
```
rgauss, line 20: Should consider `CancelIcon` here?

src/components/OSCALModificationIcons.js
```diff
@@ -0,0 +1,56 @@
...
+import { IconButton } from "@material-ui/core";
+
+export function getElementLabel(editedField) {
+  return editedField.toString().replace(/,/g, "-");
+}
+
+function getIconButtons(props) {
+  return props.modifiableData.edit[0] ? (
+    <>
+      <IconButton
```
rgauss, line 14: Can we display save first, then cancel?

src/components/OSCALMetadata.js
```diff
@@ -33,12 +35,53 @@ const useStyles = makeStyles((theme) => ({
...
+   *  modifies the value of the top-level field.
+   *
+   *  We have an "edit" sub-field for the top-level "version" and "title" fields. This
+   *  represents whether we are currently editing either of those fields.
+   *
+   *  We have the last modified top-level field because when other metadata fields are
+   *  updated, we want to update the last modified field.
+   */
+  const modifiableMetadata = {
+    "last-modified": useState(formatDate(props.metadata["last-modified"])),
```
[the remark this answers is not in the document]
mikeisen1: Looks like I overlooked that this needed to be updated. It was just leftover from the draft version of this PR when I was not sure how the response would look like.

src/components/OSCALMetadata.js
mikeisen1: I'm going to get rid of this approach, as there are now very few fields in the object.

src/components/OSCALMetadata.js
```diff
@@ -33,12 +35,53 @@ const useStyles = makeStyles((theme) => ({
...
+    version: {
+      ref: useRef("Version TextField Reference"),
```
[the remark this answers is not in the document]
mikeisen1: They are required to provide access to the `TextField` values when we click on the save icon.

kylelaker:
@mikeisen1 FYSA, #226 has been merged; I'm not sure what impact that has on this PR.

src/components/OSCALSsp.js
```diff
@@ -16,15 +16,62 @@ const useStyles = makeStyles((theme) => ({
...
+ * @param update function that will update a state, forcing a re-rendering if the PATCH request is successful
+ * @param editedField path to the field that is being updated
+ * @param newValue updated value for the edited field
+ */
+function onSave(data, update, editedField, newValue) {
+  const portString =
+    window.location.port === "" ? "" : `:${window.location.port}`;
+  const url = `http://localhost${portString}/oscal/v1/ssps/${data["system-security-plan"].uuid}`;
+
+  const dataToSave = JSON.parse(JSON.stringify(data));
```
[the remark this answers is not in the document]
mikeisen1: It was required in the old implementation but is no longer needed.

src/components/OSCALLoader.js
```diff
@@ -49,6 +49,52 @@ const oscalObjectTypes = {
...
+  }
+
+  updateData(data[editedField.shift()], editedField, newValue);
+}
+
+/**
+ *
+ * @param data data that will be passed into the body of the PATCH request, doesn't initially contain the updates
+ * @param update function that will update a state, forcing a re-rendering if the PATCH request is successful
+ * @param editedField path to the field that is being updated
```
rgauss, line 66: Should this be a bit more descriptive, maybe something like `editedFieldJsonPath` or  `fieldJsonPath` throughout the changes?

src/components/OSCALModificationIcons.js
```diff
@@ -0,0 +1,56 @@
...
+      onClick={() => {
+        props.isInEditState[1](!props.isInEditState[0]);
+      }}
+    >
+      <EditIcon fontSize={props.iconFontSize} />
+    </IconButton>
+  );
+}
+
+export default function OSCALModificationIcons(props) {
```
rgauss, line 49: To avoid confusion with control modifications in the OSCAL model, to keep similar to the test field added, and to allow for any future implementation changes (perhaps just buttons instead of icons for example) should we name this something like `OSCALEditableFieldActions`?

rgauss (changes requested):
Mostly minor adjustments, looking good.

src/components/OSCALCatalog.js
```diff
@@ -21,7 +21,10 @@ export default function OSCALCatalog(props) {
 
   return (
     <div className={classes.paper}>
-      <OSCALMetadata metadata={props.catalog.metadata} />
+      <OSCALMetadata
+        metadata={props.catalog.metadata}
+        editedField={["metadata"]}
```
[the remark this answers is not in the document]
mikeisen1: We don't need to specify `metadata`, but we do need to specify `catalog` or one of the other top-level fields.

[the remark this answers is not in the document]
rgauss: Isn't that being sent in what's currently called `patchData`?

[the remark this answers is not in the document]
mikeisen1: Yes, you are correct. The way to access the top-level field just slipped through my mind for a second.

src/components/OSCALLoader.js
```diff
@@ -49,6 +49,52 @@ const oscalObjectTypes = {
...
+    return;
+  }
+
+  updateData(data[editedField.shift()], editedField, newValue);
+}
+
+/**
+ *
+ * @param data data that will be passed into the body of the PATCH request, doesn't initially contain the updates
+ * @param update function that will update a state, forcing a re-rendering if the PATCH request is successful
```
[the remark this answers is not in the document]
mikeisen1: I would think something like `onSuccessfulPatch` or `updateOnSuccessfulPatch` as this is called only when the PATCH request is successful.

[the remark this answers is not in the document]
rgauss: `onSuccessfulPatch` sounds good.

src/components/OSCALSsp.js
```diff
@@ -73,7 +78,14 @@ export default function OSCALSsp(props) {
 
   return (
     <div className={classes.paper}>
-      <OSCALMetadata metadata={ssp.metadata} />
+      <OSCALMetadata
+        metadata={ssp.metadata}
+        isEditable={props.isEditable}
+        editedField={["system-security-plan"]}
```
rgauss, line 84: I thought we were able to drop `editedField` since we have it in `patchData`?

src/components/OSCALEditableFieldActions.js
```diff
@@ -0,0 +1,56 @@
...
+import CancelIcon from "@material-ui/icons/Cancel";
+import EditIcon from "@material-ui/icons/Edit";
+import SaveIcon from "@material-ui/icons/Save";
+import { IconButton } from "@material-ui/core";
+
+export function getElementLabel(editedField) {
+  return editedField.toString().replace(/,/g, "-");
+}
+
+function getIconButtons(props) {
```
rgauss, line 11: Any reason for this being a separate function rather than in `OSCALEditableFieldActions`?

src/stories/OSCALModificationIcons.stories.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import OSCALEditableFieldActions from "../components/OSCALEditableFieldActions";
+
+export default {
+  title: "Components/Modification Icons",
+  component: OSCALEditableFieldActions,
+};
+
+function Template(args) {
+  return <OSCALEditableFieldActions {...args} />;
```
rgauss, line 10: I think we want to rename `OSCALModificationIcons.stories.js` as well.

rgauss (changes requested):

src/components/OSCALEditableFieldActions.js
mikeisen1: I did it because the `getIconsButtons` also returns based on a condition (using a ternary operator), which would mean there's a nested ternary operator return statement if we included everything in one function. When I created the component, I was initially thinking that would make it difficult to read but it doesn't look to be so now.

src/components/OSCALSsp.js
mikeisen1: We can and that should be dropped.

rgauss (changes requested):

rgauss (approved):
Thanks for sticking through this @mikeisen1!

kylelaker (approved):
Overall, this looks great! Thanks for all the work on this; I know it's been a long effort! Great to see the `restPatch` function reused quite a bit (and of course to actually have editing functionality in here!!). I think perhaps we're overusing the ternary operator in a few of these; we might end up being better served by doing really `if`s or breaking things out into functions as we continue to add additional editing functionality across the application. That will likely help in testing parts of of the code base as complexity grows later.

Again, I really appreciate your commitment to getting this feature added!

PR #212, opened 2022-01-03
Add Storybook Addon-Docs Dependency

hreineck:
Add the @storybook/addon-docs npm dependency that is required
to build the Storybook after updating other dependencies.


kylelaker (approved):
Great catch!

kylelaker:
Going forward, we may want to make sure to bump the `@storybook/*` dependency versions together as a group

zclarkEDC (approved):

PR #226, opened 2022-01-14
Initial REST Mode

rgauss:
Closes EasyDynamics/easygrc#6

- Added a .env file containing a REACT_APP_REST_BASE_URL
- Introduced oscalObjectTypes to consolidate definitions of names, defaultUrls, jsonRoot, and restPaths
- Added simulation of a working backend findAll REST call
- Created isRestMode that checks for base url env variable
- Prevented auto-loading if isRestMode is true
- Added Select list that displays objects returned in findAll REST call that then calls given onUuidChange on selection of a MenuItem
- Added a handleUuidChange method that constructs a new URL based on UUID change then reloads
- Updated README
- Added REST mode switch to loader form

### Steps to Manually Test Functionality
- Run the latest version of `oscal-rest-service` locally
  - If building/running from source the `oscal-content` folder should already be present in `target` and the service configured for that location
- As mentioned in the README update, add a `.env.local` file in
the root of the example project with a variable of `REACT_APP_REST_BASE_URL` pointing to the REST service
- Run the example app
- You should see a select drop down and a toggle switch to go back to URL mode

### Known Issues
- If trying to view the example `basic-catalog` (the default pulled if running the scripts from all-in-one) the viewer will fail due to #227
- Profile viewing doesn't work due to upstream issues that we've created (usnistgov/liboscal-java#7, usnistgov/metaschema-java#12) and submitted fixes for (usnistgov/metaschema-java#13, usnistgov/liboscal-java#7), and while the fix has been merged it as not yet been published in a new version to Maven Central


example/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_REST_BASE_URL=
```
hreineck, line 1: Is this file necessary?

rgauss: No.  I thought it might serve as an example but the README is probably sufficient.  I've removed it.

rgauss:
This is now dependent on the `REACT_APP_REST_BASE_URL` pointing to an `oscal-rest-service` implementation that supports find all requests (EasyDynamics/oscal-rest-service#36).

bwwalshjr (approved):
This all looks good to me.

kylelaker (approved):

PR #236, opened 2022-02-03
Handle Nested Groups

### LINKED ISSUE #227: Viewer Breaks on Basic Catalog
The viewer app currently breaks when trying to view the example [basic-catalog.json](https://github.com/usnistgov/oscal-content/blob/master/examples/catalog/json/basic-catalog.json).

### Steps to Reproduce
1. Navigate to the Catalog Viewer page
2. Paste in the [raw link for the basic-catalog](https://raw.githubusercontent.com/usnistgov/oscal-content/master/examples/catalog/json/basic-catalog.json)
3. Click Reload

### Expected Behavior
4. The catalog renders

### Actual Behavior
5. A blank page is returned (and the browser console shows errors)

rgauss:
Closes #227 

Note that the basic catalog [references a missing param of `s1.1.1-prm11`](https://github.com/usnistgov/oscal-content/blob/master/examples/catalog/json/basic-catalog.json#L70) (reported at usnistgov/oscal-content#93) so the code was updated to just display that id as the parameter label if no parameter is found.

### DIFF
--- a/src/components/OSCALCatalogGroup.js
+++ b/src/components/OSCALCatalogGroup.js
@@ -55,9 +55,9 @@ export default function OSCALCatalogGroup(props) {
       </ListItem>
       <Collapse in={open} timeout="auto" unmountOnExit>
         <List className={classes.OSCALControlList}>
-          {group.groups?.map((innerGroup) =>
-            renderGroup(innerGroup, level + 1)
-          )}
+          {group.groups?.map((innerGroup) => (
+            <OSCALCatalogGroup group={innerGroup} key={innerGroup.id} />
+          ))}
           {group.controls?.map((control) => (
             <OSCALControl
               control={control}
--- a/src/components/OSCALCatalogGroup.js
+++ b/src/components/OSCALCatalogGroup.js
@@ -42,26 +42,26 @@ export default function OSCALCatalogGroup(props) {
     return hash;
   };
 
-  const renderGroup = (group, level) => (
+  return (
     <>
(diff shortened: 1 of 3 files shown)


src/components/OSCALCatalogGroup.js
```diff
@@ -26,28 +26,49 @@ export default function OSCALCatalogGroup(props) {
...
+              childLevel={level}
               key={`control-${control.id}`}
             />
           ))}
         </List>
       </Collapse>
     </>
   );
+
+  return renderGroup(props.group, 0);
```
kylelaker, line 73: Because the map for `group.groups` just does an `OSCALCatalogGroup` instead of calling `renderGroup` again, should we

- "inline" render group and remove the function invocation (since the recursion is now more generally around the whole of the `OSCALCatalogGroup`?
- change the `childLevel` to a (new?) `props.level ?? 0` or just hardcode the level to `0` again?

I'm not sure if that'd be the right thing to do. But it sorta feels like we may not really need `renderGroup` anymore though.

kylelaker (commented):
Thanks for cleaning this up! One quick question based on the new changes.

src/components/OSCALCatalogGroup.js
rgauss: Good point, removed the `renderGroup`.  Thanks!

kylelaker (approved):

bwwalshjr (approved):

PR #248, opened 2022-02-09
Fix errors during loading OSCAL documents

### LINKED ISSUE #239: Viewer Gives Error on Switch to SSP
The viewer app in the sandbox environment currently shows an error when switching from catalog to SSP.

### Steps to Reproduce
1. Navigate to the [sandbox environment viewer](https://oscal-viewer.msd.easydynamics.com/)
2. Click the nav menu and click System Security Plan Viewer

### Expected Behavior
3. The default SSP renders

### Actual Behavior
3. An error is shown:
```
Yikes! Something went wrong loading the OSCAL data. Sorry, we'll look into it. (Unexpected token < in JSON at position 0)
```

### Implementation Notes
- We should be sure the behavior is fixed for all 'root' OSCAL objects, not just SSPs.

kylelaker:
There is an error where an HTTP request that should only occur during
REST mode is enabled. This resulted in an HTTP 404 response (that
actually was perceived as a 200 OK) with an HTML body attempting to be
parsed as JSON. This caused failures up the entire stack. This ensures
we only make that call during REST mode.

Another small error is fixed that only appeared locally. This seems to
be caused by using the spread operator (`...`) on `undefined`.

Closes #239

### DIFF
--- a/src/components/OSCALLoaderForm.js
+++ b/src/components/OSCALLoaderForm.js
@@ -47,7 +47,7 @@ export default function OSCALLoaderForm(props) {
   };
 
   useEffect(() => {
-    findAllObjects();
+    if (props.isRestMode) findAllObjects();
     return () => {
       unmounted.current = true;
     };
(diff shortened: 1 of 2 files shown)


rgauss (approved):

zclarkEDC (approved):
Great work Kyle!

PR #249, opened 2022-02-10
Add Google Analytics configuration

kylelaker:
This adds basic, optional configuration of Google Analytics *if* the
`REACT_APP_GOOGLE_ANALYTICS` environment variable is present at the time
that `npm run build` is executed. This makes it easy to integrate and
fairly self-contained. While the react-ga library's code will be bundled
into the deployed app and the code to do the configuration will be as
well, the initialize won't actually happen unless that value was present
at build time.

At the moment, we just initialize and collect default metrics and
perform a "pageview" action within the App's `useEffect` (which we may
want to consider moving to some sort of an action on the Router but we
don't seem to have a full Router; we only have a Switch).


example/src/App.js
```diff
@@ -65,6 +66,19 @@ function App() {
...
+    // it is safe to do so because this is not a condition that can change
+    // between renders. `process.env` becomes a static object.
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const location = useLocation();
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useEffect(() => {
+      ReactGA.pageview(location.pathname + location.search);
+    }, [location]);
+  }
+
```
rgauss, line 81: Would this work?

```suggestion
  if (process.env.REACT_APP_GOOGLE_ANALYTICS) {
    ReactGA.initialize(process.env.REACT_APP_GOOGLE_ANALYTICS);
  }
  const location = useLocation();
  useEffect(() => {
    if (process.env.REACT_APP_GOOGLE_ANALYTICS) {
      ReactGA.pageview(location.pathname + location.search);
    }
  }, [location]);
```
We'd be unnecessarily calling `useLocation` and `useEffect` when not enabled, but that maybe that's acceptable here?

rgauss (commented):
Possible alternative, looks like linting is complaining about other stuff as well.

rgauss (approved):

zclarkEDC (approved):

PR #259, opened 2022-02-16
Add concurrency mutex for deployment actions

kylelaker:
Our deployments have a tendency to fail if they're started too close to
one another. This is because they'll try to do things like overwrite
each others changes or try to push a version of a package that already
exists. Adding the `concurrency` config attribut ensures that we always
abort the older job if a newer one is starting for the same thing.

We add this at the workflow level because doing it at the `job` level
doesn't result in quite as instantaneous cancellation.

### DIFF
--- a/.github/workflows/gh-packages.yml
+++ b/.github/workflows/gh-packages.yml
@@ -4,7 +4,9 @@ on:
   push:
     branches: [develop]
   workflow_dispatch:
-    
+
+concurrency: package-push
+
 jobs:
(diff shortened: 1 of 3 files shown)


kylelaker:
This should prevent issues like https://github.com/EasyDynamics/oscal-react-library/actions/runs/1854637517

rgauss (approved):

zclarkEDC (approved):

PR #260, opened 2022-02-16
Fix tests when GA ID present

kylelaker:
`jest` automatically sets `NODE_ENV` to `"test"` so this should just work. This wasn't caught outside the pipeline because the deployment pipeline is the only place where we set the GOOGLE_ANALYTICS env var.


zclarkEDC (approved):
This is a potential codebuild fix due to jest issues

mikeisen1 (approved):

PR #265, opened 2022-02-21
Add OSCALJsonEditor Component

pkothare:
Uses `react-split` to create a vertically split element into two panes. This only manifests in REST mode and starts off with 50/50 split for the two panes. The end user is then able to adjust the width of the pane up to a certain limit.

- The left pane wraps the `OSCALJsonEditor` component which consists of a configured instance of the `@monaco-editor` component. The editor is does take up 100% of the vertical height of the HTML page, instead, it adapts to screen height and is positioned relative to the `scrollport` based on [sticky positioning](https://www.w3.org/TR/css-position-3/#stickypos-insets).
- The right pane wraps the viewer, which based on the route is one of the specific `OSCALLoader` components.

### TO DOs:
- [x] Migrate static CSS to [`makeStyles`](https://v4.mui.com/styles/basics/#hook-api) used by MUI
- [x] Use fetch to complete PUT request
- [x] Write tests 
- [x] Fix overflow issues with OSCAL components. MUI 4 has significant issues element overflows when generic [spacing](https://v4.mui.com/system/spacing/#api) is used.

### References
- [@monaco-editor/react](https://www.npmjs.com/package/@monaco-editor/react)
- [Monaco Editor](https://microsoft.github.io/monaco-editor/)
- [react-split](https://www.npmjs.com/package/react-split)
- [split.js](https://split.js.org/)
- https://github.com/EasyDynamics/easygrc/issues/23


kylelaker:
Seems like the content uses ~100% of the display width in non-REST mode now. Is that an intentional change?

pkothare:
> Seems like the content uses ~100% of the display width in non-REST mode now. Is that an intentional change?

Intentional, based on the screenshot that @rgauss shared with us on the potential design of including an editor. If we change the display width between rest/non-rest mode, the layout changes pretty dramatically which leads to some cognitive load. Nonetheless, this has been factored by a very small change [Change to XL](https://github.com/EasyDynamics/oscal-react-library/pull/265/files#diff-70aa9f24302e170717b76b7501425ccf9cb910991099eaef6b17a18cdac37dbfL184). So if we decide to go smaller we can. 🤞, all other components are dynamically sized.

> The dropdown for the document and the toggle for REST mode should be aligned (REST mode toggle is currently higher)

Worked this in already, the changes will be up soon.

> The top bar goes away when scrolling -- do we want to keep it? maybe not?

I tried not to modify existing layout/style decisions. Prior to this PR, the header was not static, so no change there. However changing it to be static isn't that hard.

> Save/Discard buttons require scrolling to see (on my monitor).

This was intentional, so that when you do scroll past a certain point (i.e. when the top nav and selector are past the scroll port), the editor (including the buttons) fills up the viewport vertical space. Eventually the `OSCALLoaderForm` will move to a left hand drawer which will pull the editor and the buttons up, and no scroll will be required. We can either reduce the height now, or it can be part of the update when we develop the left hand drawer. It's mostly a matter of testing with different [Editor Vertical Height values](https://github.com/EasyDynamics/oscal-react-library/pull/265/files#diff-6714c64971be470b06a7bcc75bc900cb9939244e1f94eca57ef72892825de691R43)

kylelaker (approved):
This looks good. There are things it might be sorta nice to fix in future tickets but that I don't think should slow down our progress here. A few of them are:

- [ ] base64 encoding should be handled as part of the build process, we should avoid directly writing b64 strings directly into the application (we do have Webpack here which may help)
- [ ] it would be cool if the browser find functionality worked within the JSON editor. it has it's own (better) find functionality but you have to click within it and if you don't the browser doesn't see any of the Editor content
- [ ] TABS?! :laughing:

src/components/OSCALLoader.js
```diff
@@ -192,12 +224,31 @@ export default function OSCALLoader(props) {
   if (!isLoaded) {
     result = <CircularProgress />;
   } else if (oscalUrl) {
-    result = props.renderer(
-      isRestMode,
-      oscalData,
-      oscalUrl,
-      onResolutionComplete,
-      handleRestPatch
+    result = isRestMode ? (
```
rgauss, line 227: Per the acceptance criteria in EasyDynamics/easygrc#23, we need to be able to toggle source editing on and off.  The AC currently says source should be off by default but I think I'm going to change that.

Perhaps we could use the `Code` or `Source` material icon as the toggle.

src/components/OSCALLoaderForm.js
```diff
@@ -135,6 +117,21 @@ export default function OSCALLoaderForm(props) {
             </FormControl>
           </Grid>
         )}
+        {process.env.REACT_APP_REST_BASE_URL && (
+          <Grid item xs={2}>
```
rgauss, line 121: When not in REST mode the toggle overlaps the URL input.

src/components/OSCALLoader.js
```diff
@@ -192,12 +224,31 @@ export default function OSCALLoader(props) {
...
+      <Split className={classes.split} minSize={500} sizes={[50, 50]}>
```
rgauss, line 228: Can we start this out closer to a 1/3 source and 2/3 rendered object rather than a 50/50 split?

And perhaps a `minSize` of `300`?

src/components/OSCALJsonEditor.js
```diff
@@ -0,0 +1,83 @@
...
+              }}
+              startIcon={<SaveIcon />}
+              variant="contained"
+              color="primary"
+            >
+              Save
+            </Button>
+          </Grid>
+          <Grid item>
+            <Button
```
rgauss, line 67: I think this might flow better if the buttons were aligned to the right with Cancel first?

src/components/OSCALJsonEditor.js
```diff
@@ -0,0 +1,83 @@
...
+              onClick={() => {
+                editorRef.current.setValue(
+                  JSON.stringify(props.value, null, "\t")
+                );
+              }}
+              variant="contained"
+              color="secondary"
+            >
+              Discard
```
rgauss, line 76: Should we call this Cancel instead and use the `Cancel` material icon that's used in the UI editing?

src/components/OSCALLoader.js
```diff
@@ -192,12 +224,31 @@ export default function OSCALLoader(props) {
...
+        <Box>
+          <OSCALJsonEditor value={oscalData} onSave={handleRestPut} />
```
rgauss, line 230: Here `oscalData` will often contain fields other than what's returned in the REST response due to things like profile resolution.  For example, SSPs will contain a `resolvedControls` array that should not be shown in source and should not be sent in the `PUT` REST request.

rgauss (changes requested):

src/components/OSCALLoaderForm.js
pkothare: Improved the layout in: https://github.com/EasyDynamics/oscal-react-library/pull/265/commits/926572c0359647b4541a790395fb3d51995f1b62

src/components/OSCALLoader.js
pkothare: Updated the values in https://github.com/EasyDynamics/oscal-react-library/pull/265/commits/298c17f713c60b046dd7c8aae7564227dd4c14ff

src/components/OSCALJsonEditor.js
pkothare: Updated layout and alignment in https://github.com/EasyDynamics/oscal-react-library/pull/265/commits/cf83ad1ea976017b059d2b5f5711bda2e92a7168

src/components/OSCALJsonEditor.js
pkothare: Updated the text and icon in https://github.com/EasyDynamics/oscal-react-library/pull/265/commits/cf83ad1ea976017b059d2b5f5711bda2e92a7168

src/components/OSCALLoader.js
pkothare: Handled the removal of `resolvedControls` in https://github.com/EasyDynamics/oscal-react-library/pull/265/commits/cf83ad1ea976017b059d2b5f5711bda2e92a7168. If there are any others that need to be removed, please let me know. It's pretty opaque to me how those got in added to `oscalData`.

rgauss: The way the additional fields are populated and passed in the resolvers ([example](https://github.com/EasyDynamics/oscal-react-library/blob/feature/json-editor/src/components/oscal-utils/OSCALSspResolver.js#L35)) is currently a bit sloppy.

At some point we should revisit how we pass that additional info around to the components, but for now, perhaps we should consider shoving the original response string in something like `oscalData.oscalSource` rather than the approach of copying and removing what was added?

hreineck (approved):
UI looks great tested locally. Would like to see the toggle functionality as Ray described but other than that ready to merge.

rgauss (approved):

rgauss:
@kylelaker, have your concerns been addressed here?

kylelaker:
@rgauss I think so for the most part. I had previously approved and I think then dismissed because _others_ had raised concerns. I haven't been following the developments on this branch super closely but I have no objections to merging it.

Issue #263, opened 2022-02-18
Only text in navigation menu items are links


kylelaker:
In the example OSCAL Viewer application's navigation menu, only the text of the links actually operates as a link; though the entire menu item has the you-can-click-this style of pointer. So in the image below, you have to click the blue text; if you click any of the white area, the page will not navigate.

![image](https://user-images.githubusercontent.com/850893/154758311-f9e3d179-3c12-4596-9d7b-ca86ee470d3c.png)

## Acceptance Critera

- Clicking anywhere in a menu item (on the link or around it) works correctly to navigate to the intended page

## Implementation Notes

This probably means doing something like turning the entire `<MenuItem>` into a button with a `<Navigate>` or "flipping" the `<Link>` and `<MenuItem>` but I think it's improper for an anchor to contain complex elements.

rgauss:
@kylelaker, can you retroactively estimate the story points on this one?

kylelaker:
@rgauss I added an estimate based on my understanding of the complexity relative to other tickets in this sprint; happy to discuss on/after standup if we want to talk about it as a team though!

PR #266, opened 2022-02-22
Fix navigation links

kylelaker:
This makes it so the entire menu item is clickable rather than just the
link for navigation. This breaks the standard <a> tag formatting (so
it's not blue anymore) but it matches the documentation from material
UI: https://v4.mui.com/components/menus/

Closes: #263


kylelaker:
image

pkothare (approved):

zclarkEDC (approved):

Issue #240, opened 2022-02-08
Viewer Crashes on Load of Wrong OSCAL Doc Type


rgauss:
Currently the viewer app crashes when loading a URL with the wrong OSCAL object type, i.e. reloading an SSP URL in the Catalog Viewer

### Steps to Reproduce
1. Navigate to the [sandbox viewer environment](https://oscal-viewer.msd.easydynamics.com/)
2. Paste an SSP URL into the URL input and click Reload

### Expected Behavior
3. An error is shown (standard red bar with 'Yikes') that indicates the URL response could not be parsed

### Actual Behavior
3. The page crashes (white page with no error is shown)

### Implementation Notes
- We want to be sure this is fixed for all 'root' OSCAL objects, not just catalogs.
- We'll want an automated end-to-end (Cypress) test to confirm behavior

zclarkEDC:
Taking a look at this today @rgauss

kylelaker:
@zclarkEDC I was playing around with this and created https://github.com/EasyDynamics/oscal-react-library/compare/feature/doc-type-crash which may be useful. It doesn't fully take care of the the second implementation note (adding the test) and I think we can go further and cleanup more of the `onError`/`setError` stuff. I also didn't really document anything. We also may not want that `console.error`.

Information on Error Boundaries can be found at: https://reactjs.org/docs/error-boundaries.html. The way error handling changed in React 16+ to just unmount the whole app is at https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html.

![image](https://user-images.githubusercontent.com/850893/153526777-2b17c09c-46bb-476b-a42d-29dec73aade5.png)

A sample is in the screenshot. Note that while "Reload" doesn't look disabled (as expected until #163 is fixed), if you click it, it doesn't do anything and it then becomes gray.

zclarkEDC:
> @zclarkEDC I was playing around with this and created https://github.com/EasyDynamics/oscal-react-library/compare/feature/doc-type-crash which may be useful. It doesn't fully take care of the the second implementation note (adding the test) and I think we can go further and cleanup more of the `onError`/`setError` stuff. I also didn't really document anything. We also may not want that `console.error`.
> 
> Information on Error Boundaries can be found at: https://reactjs.org/docs/error-boundaries.html. The way error handling changed in React 16+ to just unmount the whole app is at https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html.
> 
[...]

PR #267, opened 2022-02-22
Fix Viewer Crashing on Load of Wrong OSCAL Doc Type

zclarkEDC:
Fix for [#240 ](https://github.com/EasyDynamics/oscal-react-library/issues/240)

Should work for Catalogs, SSPs, and Components.

### DIFF
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -1,0 +1,30 @@
+import React from "react";
+import Alert from "@material-ui/lab/Alert";
+
+export function defaultFormatter(error) {
+  return (
+    <Alert severity="error">
+      Yikes! Something went wrong loading the OSCAL data. Sorry, we&apos;ll look
+      into it. ({error.message})
+    </Alert>
+  );
+}
+
+export class ErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.errorFormatter = props.errorFormatter ?? defaultFormatter;
+    this.state = { error: undefined };
+  }
+
+  componentDidCatch(error) {
+    this.setState({ error });
+  }
+
+  render() {
+    if (this.state.error !== undefined) {
+      return this.errorFormatter(this.state.error);
+    }
+    return this.props.children;
+  }
+}
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -1,7 +1,7 @@
 import React from "react";
 import Alert from "@material-ui/lab/Alert";
 
-export function defaultFormatter(error) {
+function defaultFormatter(error) {
   return (
     <Alert severity="error">
       Yikes! Something went wrong loading the OSCAL data. Sorry, we&apos;ll look
@@ -10,7 +10,7 @@ export function defaultFormatter(error) {
   );
 }
 
-export class ErrorBoundary extends React.Component {
+class ErrorBoundary extends React.Component {
   constructor(props) {
     super(props);
     this.errorFormatter = props.errorFormatter ?? defaultFormatter;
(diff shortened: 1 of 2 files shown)


kylelaker:
Should we move `ErrorBoundary` into it's own file and export it so that we can have other error boundaries all over?

kylelaker:
That would probably also mean that `ErrorBoundary` should accept an `errorFormatter` prop or something like that so that you can override how an error is displayed at a specific part of the application:

snippets of changed functions:

```js
function defaultFormatter(error) {
  return (
    <Alert severity="error">
      Yikes! Something went wrong loading the OSCAL data. Sorry, we&apos;ll look
      into it. ({error.message})
    </Alert>
  );
);

export class ErrorBoundary extends React.Component {
  constructor(props) {
    super(props)
    this.errorFormatter = props.errorFormatter ?? defaultFormatter;
    this.state = { error: undefined };
  }

  render() {
    if (this.state.error !== undefined) {
      return this.errorFormatter(this.state.error);
    }
    return this.props.children;
  }
}
```

Idk if the `errorFormatter` should go on `this` or `this.state`. I also don't know if I like that name.

src/components/OSCALLoader.js
```diff
@@ -99,8 +99,30 @@ function restPatch(
...
+    this.state = { error: undefined };
+  }
+
+  static getDerivedStateFromError(error) {
+    return { error };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    this.setState({ error });
+    console.error({ ...this.state, errorInfo });
```
kylelaker, line 114: We probably also want to remove this

zclarkEDC: Yeah I was wondering why that was removed

zclarkEDC: Actually it doesn't work without `componentDidCatch`,

src/components/ErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
...
+export function defaultFormatter(error) {
+  return (
+    <Alert severity="error">
+      Yikes! Something went wrong loading the OSCAL data. Sorry, we&apos;ll look
+      into it. ({error.message})
+    </Alert>
+  );
+}
+
+export class ErrorBoundary extends React.Component {
```
kylelaker, line 13: Should we actually `export default` to match the style of the other components?

src/components/ErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import Alert from "@material-ui/lab/Alert";
+
+export function defaultFormatter(error) {
```
kylelaker, line 4:
```suggestion
function defaultFormatter(error) {
```

Let's _not_ export this

kylelaker (changes requested):

kylelaker (approved):

src/components/ErrorBoundary.js
```diff
@@ -0,0 +1,31 @@
...
+
+  componentDidCatch(error) {
+    this.setState({ error });
+  }
+
+  render() {
+    if (this.state.error !== undefined) {
+      return this.errorFormatter(this.state.error);
+    }
+    return this.props.children;
```
pkothare, line 28:
```suggestion
    return this.props.children || null;
```
The reason why is that `this.props` or `this.props.children` can be undefined, and I was receiving the error:

```
Uncaught Error: ErrorBoundary(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

pkothare (changes requested):

src/components/ErrorBoundary.js
zclarkEDC: Good catch!

pkothare (approved):

PR #268, opened 2022-02-22
Remove duplicate SSP Authorization Boundary Desc

kylelaker:
This removes the duplicate field and rebalances the layout to make it
flow a little more naturally.

### DIFF
--- a/src/components/OSCALSystemCharacteristics.js
+++ b/src/components/OSCALSystemCharacteristics.js
@@ -56,7 +56,7 @@ export default function OSCALSystemCharacteristics(props) {
             </Grid>
             <Grid
               item
-              xs={5}
+              xs={6}
               className={classes.OSCALSystemCharacteristicsInfo}
             >
               <Typography variant="h6">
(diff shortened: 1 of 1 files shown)


kylelaker:
image

src/components/OSCALSystemCharacteristics.js
```diff
@@ -56,7 +56,7 @@ export default function OSCALSystemCharacteristics(props) {
             </Grid>
             <Grid
               item
-              xs={5}
+              xs={6}
```
zclarkEDC, line 59: https://mui.com/api/grid/

We could also do `xs={auto}` right?
Does that look better?

zclarkEDC (approved):

src/components/OSCALSystemCharacteristics.js
kylelaker: Umm I guess we could? But nothing else here really does so I didn't want to get too crazy with it in a bug fix.

kylelaker: Might make a good `enhancement` ticket?

rgauss (approved):

PR #269, opened 2022-02-23
Fixes for successful patch re-rendering

### LINKED ISSUE #272: Successful PATCH Requests Give an Error in UI
Currently UI updates send the correct `PATCH` requests and receive the expected response, however, the UI fails to properly handle that response and throws an error.

### Steps to Reproduce
1. Launch the REST service and UI in REST mode
2. Navigate to an SSP
3. Edit the Title field and click save

### Expected Behavior
4. The rendered SSP should be refreshed with the updated data

### Actual Behavior
4. An error is thrown: `Cannot read properties of undefined (reading 'components'))`

rgauss:
Closes #272 

- Moved top-level `restPatch` function to `OSCALLoader.handleRestPatch` to re-use state variables for rendering
- Removed `update` arguments
- Removed `ssp` state variable in `OSCALSsp`

This probably needs further investigation before creating an issue and reviewing but I wanted to get what seems like a fix up somewhere so it's not lost.

### DIFF
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -35,7 +35,7 @@
         "@material-ui/styles": "^4.11.4",
         "@testing-library/dom": "^8.11.3",
         "@testing-library/jest-dom": "^5.16.2",
-        "@testing-library/react": "^12.1.2",
+        "@testing-library/react": "^12.1.3",
         "history": "^5.2.0",
         "prop-types": "^15.8.0",
         "react": "^17.0.2",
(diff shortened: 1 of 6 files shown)


kylelaker:
This looks good to me! This along with some of the CORS fixes makes it possible for me to PATCH and update the version and title of an SSP without any errors in Dev Tools Console or Network tabs. I'll hold off on approving until after this is marked as ready for review, but I'll take the incremental improvement while we work on fleshing out a full ticket for any associated bug fixes that may be required.

pkothare (approved):

PR #274, opened 2022-02-24
Add automation to update example `package-lock`

kylelaker:
Because the example application's `package-lock.json` references the
parent directory for the vast majority of its dependencies, it having
outdated information isn't necessarily an issue; however, it does
falsely trigger quite a few security alerts within GitHub and Dependabot
itself isn't quite talented enough to understand what's going on.

This adds a new GitHub workflow to follow up every pull request that
modifies `package-lock.json` (because people can forget too!) to
synchronize those changes back to the example application.

One of the things I thought about doing here was just pushing to the
PR branch, but dependabot doesn't tolerate that well (it will stop
updating its own PR) and people may not either.

### DIFF
--- a/.github/workflows/dependency-pr.yml
+++ b/.github/workflows/dependency-pr.yml
@@ -16,6 +16,8 @@ jobs:
   update:
     # If the pull request was closed then nothing needs to be done
     # if: github.event.pull_request.merged == 'true'
+    runs-on: ubuntu-latest
+    name: Synchronize version
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


mikeisen1 (approved):

rgauss (approved):

kylelaker:
I am going to move this to draft status while I look into a potential issue with this email address and the CLA bot.

rgauss:
This can be merged now?

kylelaker:
Yeah I'm going to go for it. I've tested it a bit and I think the CLA issue is solved. I've updated the CLA tool to green check PRs that originate from the source repo. Because this is `pull_request_target` it's hard to test fully while on a PR so we may as well merge and fixup if we need to.

PR #273, opened 2022-02-24
Add support for network and data flow diagrams

kylelaker:
This mimics the existing behavior for Authorization Boundaries

Closes: #222
Closes: #223

PR #276, opened 2022-02-24
Add support for remote URIs in links

kylelaker:
While working on the implementation for #273 it was unclear to me
whether the specification permits diagrams to be referenced directly by
URI or not so I split this out into its own pull request.

The specification specifically states:

> The value of the href can be an internet resource, or a local reference
> using a fragment e.g. #fragment that points to a back-matter resource in
> the same document.
>
> If a local reference using a fragment is used, this will be indicated by
> a fragment "#" followed by an identifier which references an identified
> resource in the document's back-matter or another object that is within
> the scope of the containing OSCAL document.
>
> If an internet resource is used, the href value will be an absolute or
> relative URI pointing to the location of the referenced resource. A
> relative URI will be resolved relative to the location of the document
> containing the link.

This, to me, implied that we should be permitting absolute and relative
links via a URI to internet resources.

To implement this, a "wrapper"-ish function has been created for what
used to be a pure Back Matter lookup function. It handles the logic to
determine whether the link is internal (needs to be looked up in Back
Matter) or an internet resource (is a URI).

There are some incidental changes to code and tests due to the changing
of imports and addition of exported `test-data`.

### DIFF
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -8,7 +8,7 @@ import { Typography } from "@material-ui/core";
 import FormatQuoteIcon from "@material-ui/icons/FormatQuote";
 import DescriptionIcon from "@material-ui/icons/Description";
 import StyledTooltip from "./OSCALStyledTooltip";
-import { getAbsoluteUrl } from "./oscal-utils/OSCALBackMatterUtils";
+import { getAbsoluteUrl } from "./oscal-utils/OSCALLinkUtils";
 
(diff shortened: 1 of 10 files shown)


kylelaker:
This should be testable by changing an existing image's `link.href` to a full URI and seeing that it still works properly.

rgauss (approved):
I thought we already had the notion of this somewhere but I guess not.

Definitely related to our [TODOs on supporting other URL schemes and use cases](https://github.com/EasyDynamics/oscal-react-library/blob/develop/src/components/oscal-utils/OSCALProfileResolver.js#L44).

zclarkEDC (approved):

PR #277, opened 2022-02-24
Edit SSP Control Implementations UI

### LINKED ISSUE #337: Subtask: SSP Update Existing Implemented Requirements UI Changes
Subtask of EasyDynamics/easygrc#9

As a user of the OSCAL Editor Web App I want to be able to edit existing control implementation implemented-requirement objects of an existing SSP in a UI so that I can incrementally define the SSP.

### Acceptance Criteria
- The UI supports editing of existing control implementation implemented-requirements

mikeisen1:
Closes #337

Creates the user interface for editing existing control implementations. These edits can be to statement descriptions, placeholder labels for non-existent set-parameter values, and to existing set-parameter values. The UI then sends in an HTTP PUT request to a backend service and handles the response. This allows a user, given the backend can handle the request, to modify existing control implementations of an SSP using the OSCAL Viewer. Like all edits using the OSCAL Viewer, changes are reflected on the viewer if the request is successful and are not reflected on the viewer if the request is not successful. On an unsuccessful request, an alert is sent to the user notifying them that the changes could not be made.

For the HTTP PUT request, the url path is of the form `/system-security-pan/{ssp-uuid}/control-implementation/implemented-requirements/{implemented-requirement-uuid}`, where we are just passing in the body of one implemented requirement, not the entire SSP.


package.json
```diff
@@ -23,11 +23,13 @@
     "@material-ui/styles": "^4.11.4",
     "@testing-library/dom": "^8.11.3",
     "@testing-library/jest-dom": "^5.16.2",
+    "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
kylelaker: Because our data starts as JSON data anyway, could we do a JSON roundtrip to handle deep clones? Maybe we can break this out into our own `deepClone` function? I don't think that we have any functions or non-JSONable data types.

```js
export function deepClone(obj) {
  return JSON.parse(JSON.stringify(obj));
}
```

src/components/OSCALControlProse.js
```diff
@@ -259,27 +272,159 @@ export function OSCALReplacedProseWithParameterLabel(props) {
...
+      {props.isEditable ? (
+        <Grid container>
+          <OSCALPopover
+            anchorEl={anchorEl}
+            setAnchorEl={setAnchorEl}
+            controlId={props.controlId}
+            onCancel={() => {
+              setAnchorEl(null);
+            }}
+            onFieldSave={(descriptionReference, implementationReference) => {
```
kylelaker, line 328: This is definitely too complicated to declare inline. Can we define it at least at the top of the function if not on its own entirely.

src/components/OSCALControlProse.js
```diff
@@ -259,27 +272,159 @@ export function OSCALReplacedProseWithParameterLabel(props) {
...
+              props.onFieldSave(
+                false,
+                partialRestData,
+                props.update,
+                editedField,
+                null,
+                "PUT",
+                `${rootOscalObjectName}/${props.restData[rootOscalObjectName].uuid}/control-implementation/implemented-requirements/${props.implementedRequirement.uuid}`
+              );
+            }}
```
kylelaker, line 370: We're eventually want to make our own sort of API client library at some point so it's not so hardcoded into the components.

mikeisen1:
I know we mentioned that we wanted the backend service to create a new uuid if necessary, but I left the uuid and the auto-generation of the new uuid in for now until we definitively decide the path taken for that. The new uuid generation is necessary when editing a parameter label, as in that situation, there is no by-component object, which requires a uuid, for that statement.

src/components/OSCALLoader.js
```diff
@@ -42,18 +43,76 @@ const oscalObjectTypes = {
...
+      (result) => {
+        onSuccessfulRequest(result);
+      },
+      () => {
+        alert(
+          `Could not update the ${path.replace(/,0,|,0|,/g, (matched) =>
+            jsonPathRegexReplace(matched)
+          )} field${newValue ? ` with value: ${newValue}.` : "."}`
+        );
+      }
```
kylelaker, line 114: A recent PR removed the `alert()`. Let's not add it back.

src/components/OSCALMetadata.js
```diff
@@ -150,9 +153,13 @@ export default function OSCALMetadata(props) {
                     }
                   : null
               }
+              restMethod="PATCH"
```
kylelaker, line 156: We should probably define an `enum` for this somewhere.

kylelaker (changes requested):
I am going to have to break my review on this into a couple different passes -- here's my first try at it. I'll try to do another here soon.

src/components/oscal-utils/OSCALUtils.js
```diff
@@ -0,0 +1,62 @@
+export function deepClone(obj) {
+  return JSON.parse(JSON.stringify(obj));
+}
+
+/**
+ *
```
tuckerzp, line 6: Function name is fairly descriptive, but it might be nice to have a brief explanation on the function in the doc. Especially given the params are nicely described.

src/components/OSCALLoader.js
```diff
@@ -86,29 +73,45 @@ export default function OSCALLoader(props) {
 
   /**
    *
```
tuckerzp, line 75: Function description would be nice here as well to go with param descriptions.

tuckerzp (commented):
Just a few code documentation things. I will leave the rest to Kyle's next couple passes.

src/components/OSCALControlProse.js
```diff
@@ -259,27 +335,128 @@ export function OSCALReplacedProseWithParameterLabel(props) {
...
             return getTextSegment(segment, index.toString());
           }
-          return getParameterLabelSegment(
-            props.parameters,
-            segment,
-            props.modificationSetParameters,
-            index.toString()
+
+          return (
+            <Typography display="inline" variant="body2">
```
rgauss, line 365: This has decreased the font size for labels.  Are we sure we need to wrap here?

src/components/OSCALControlProse.js
```diff
@@ -239,12 +247,80 @@ function getParameterValueSegment(
   );
 }
 
+function onFieldSaveParameterLabel(
```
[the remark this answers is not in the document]
rgauss: This seems complex enough to warrant some function documentation.

src/components/OSCALPopover.js
```diff
@@ -0,0 +1,91 @@
...
+      }}
+      anchorEl={props.anchorEl}
+      anchorOrigin={{
+        vertical: "bottom",
+        horizontal: "center",
+      }}
+    >
+      <Grid container className={classes.ControlProseEditor} xs={12}>
+        <Typography variant="h6">
+          Editing control {props.controlId}, statement {props.statementId}
```
rgauss, line 34: If we take this UX/UI approach then:

```suggestion
          Implementation of Statement '{props.statementId}' for Control '{props.controlId}'
```

However, as discussed in meetings it might make more sense to bring focus to the prose as already displayed on the page with inline input fields, in which case this might not be needed.

src/components/OSCALPopover.js
```diff
@@ -0,0 +1,91 @@
...
+        </Typography>
+        <Grid container xs={12}>
+          {props.isUserDefinedImplementation ? (
+            <>
+              <Grid item xs={3}>
+                <Typography>Implementation: </Typography>
```
rgauss, line 40: These are parameter values, not exactly 'Implementation', and there can be multiple.

src/components/OSCALPopover.js
```diff
@@ -0,0 +1,91 @@
...
+import OSCALEditableFieldActions from "./OSCALEditableFieldActions";
+
+const useStyles = makeStyles((theme) => ({
+  ControlProseEditor: {
+    padding: theme.spacing(2),
+    width: 500,
+  },
+}));
+
+export default function OSCALPopover(props) {
```
rgauss, line 15: This is pretty specific to editing control implementations and not a generically reusable popover, and we may not stick with a popover as the UX/UI implementation.

As such the files and default function here should be named something more like `OSCALControlPartEditor`.

rgauss (changes requested):

src/components/OSCALControlProse.js
mikeisen1: We do not, so it is getting removed.

src/components/OSCALControlImplementation.js
```diff
@@ -56,11 +56,19 @@ export default function OSCALControlImplementation(props) {
...
                       implementedRequirement={implementedRequirement}
+                      childLevel={0}
                       components={props.components}
                       controls={props.controls}
-                      childLevel={0}
+                      isEditable={props.isEditable}
                       key={implementedRequirement.uuid}
                       modifications={props.modifications}
+                      onFieldSave={props.onFieldSave}
+                      restData={{
```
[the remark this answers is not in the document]
mikeisen1: There isn't a reason for the 'fresh' copy, so it was removed

src/components/OSCALControlProse.js
```diff
@@ -331,6 +406,39 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
...
+              anchorEl={anchorEl}
+              setAnchorEl={setAnchorEl}
+              controlId={props.controlId}
+              isUserDefinedImplementation={props.prose.match(
+                prosePlaceholderRegexpString
+              )}
+              onCancel={() => {
+                setAnchorEl(null);
+              }}
+              onFieldSave={(descriptionReference, implementationReference) => {
```
rgauss, line 429: Again, we shouldn't import functions like `onFieldSaveByComponentParameterValue` from a component then send them back in to rendering declaration of that component.

rgauss (changes requested):

kylelaker (approved):
There are maybe a few things we can still work on here (I mean, there are some `TODO`s) but overall, I think anything I see is really nitpicky and I don't want to hold this up. We've done quite a bit of review across several PRs now and I am, overall, really happy with the state of this. Thanks for all the work on this!

rgauss (approved):
Awesome work!

Created #390 for the tracking future work on the TODO mentioned in standup.

PR #281, opened 2022-02-26
Persist rest mode toggle across page loads

kylelaker:
This makes the configuration of "rest mode" an app-level configuration
that is a parameter to the library components. This matches with how the
Google Analytics is set; additionally, it means that the environment
variable should be set when building the app and that this isn't a
decision made when compiling the library (so it doesn't have to be set
in the CI pipeline!).

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -62,6 +62,10 @@ function App() {
   const classes = useStyles();
 
   const [anchorEl, setAnchorEl] = useState(null);
+  const [isRestMode, setIsRestMode] = useState(
+    !!process.env.REACT_APP_REST_BASE_URL
+  );
+  const [backendUrl] = useState(process.env.REACT_APP_REST_BASE_URL);
 
   const handleAppNavOpen = (event) => {
     setAnchorEl(event.currentTarget);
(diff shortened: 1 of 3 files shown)


src/components/OSCALLoaderForm.js
```diff
@@ -63,7 +61,7 @@ export default function OSCALLoaderForm(props) {
       }}
     >
       <Grid container spacing={3}>
-        {process.env.REACT_APP_REST_BASE_URL && (
+        {!!props.backendUrl && (
```
mikeisen1, line 64: Any reason we're using `!!`?

kylelaker: Good question! Mostly I'm scared of short-circuiting and non-explicit truthiness checks so I wanted to coerce it to a boolean. I suppose that without `!!` the condition will always be the same result? If you say it will be then I'll go ahead and remove the `!!`. I had a concern that there might be a scenario where we actually displayed `""` if `REACT_APP_REST_BASE_URL` was set but it was set to the empty string.

mikeisen1: I'm not concerned about empty strings, as removing the `!!` will still lead to the same result. I'm more concerned if someone provides a string containing only whitespace, which the `!!` handles incorrectly anyways.

kylelaker: True; though we never really validate the REST API URL anyway. So `" "` isn't any different than `"https://google.com/"`. I'll tinker with it and see if I can find a better solution here.

mikeisen1 (approved):

kylelaker:
This is ready to merge but I've moved it to draft because:
1. It now has a simple merge conflict
2. I think it'd conflict even bigger with #265 and it'd be better to get that PR in first and rebase this one

kylelaker:
Rebased and ready to go!

example/src/App.js
```diff
@@ -62,6 +62,13 @@ function App() {
   const classes = useStyles();
 
   const [anchorEl, setAnchorEl] = useState(null);
+  const [isRestMode, setIsRestMode] = useState(
+    // We want to ensure that throughout the app is is always a boolean
```
rgauss, line 66:
```suggestion
    // We want to ensure that throughout the app this is always a boolean
```

rgauss (approved):

kylelaker:
@mikeisen1 I see that you re-requested yourself for review. Is there something in particular here you'd like to see changed or can we go ahead and merge this?

mikeisen1:
> @mikeisen1 I see that you re-requested yourself for review. Is there something in particular here you'd like to see changed or can we go ahead and merge this?

I re-requested a review after seeing your comments on needing to rebase this PR. But everything looks good to go now, and I'll approve it.

Issue #283, opened 2022-02-28
Loading of Non-Existent URL Doesn't Display Error


rgauss:
Currently attempting to load a non-existent URL in the viewer does not display the error text.

### Steps to Reproduce
1. Load a non-existent URL into a the viewer app ([example](https://oscal-viewer.msd.easydynamics.com/system-security-plan?url=https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/bad-branch/system-security-plans/ssp-example.json))

### Expected Behavior
2. The red error notice should be displayed

### Actual Behavior
2. The loading spinner remains present and the console throws an error of:
```
Uncaught (in promise) Error: 404
    at OSCALLoader.js:95:1
```

kylelaker:
This is likely due to the fix in #267 for #240. Error boundaries cannot catch/handle errors that occur [in an asynchronous context](https://reactjs.org/docs/error-boundaries.html#introducing-error-boundaries). The error here occurs within the `fetch` which is async. We can't (or don't want to) `await` the `fetch` (we use the asynchronous callbacks) so instead we need to find a way to catch the error and throw it again in a context where the Error Boundary _can_ handle it. We want to avoid having to add layers of `setError={setError}` into the components, so we'll want to look for another solution that allows us to handle this more seamlessly.

This _might_ be fixable in the short term by reverting the second commit in #267 (204b79609c9946ef1a47ce142975fb60d48acafd) but that feels short-sighted because it doesn't really provide a general solution.

kylelaker:
https://medium.com/trabe/catching-asynchronous-errors-in-react-using-error-boundaries-5e8a5fd7b971 suggests that perhaps we could add a `setState` that `throw`s. And if we were to use a snippet like the one in that article, perhaps it could even be exported from the `ErrorBoundary.ts` file itself.

rgauss:
@kylelaker, thoughts on a size here?

kylelaker:
@rgauss Relatively about the same complexity as the open source tasks; 3?

Issue #163, opened 2021-09-03
Reload Button Stays Disabled on Failed URL Fetch


hreineck:
When a URL is entered that leads to a failed fetch or other error during loading, the reload button stays greyed out, preventing the user from entering a new URL:

![image](https://user-images.githubusercontent.com/43356777/132062877-6d7dc375-8f90-450d-b965-a3d172278d58.png)

## Steps to Reproduce

1. Enter an invalid url into any of the four URL bars.
2. Press reload.

OR navigate to a link with an invalid url such as:
https://oscal-viewer.msd.easydynamics.com/catalog?url=goofy

## Acceptance Criteria

The reload button is still enabled even when the error message is displayed.

## Implementation

This likely has to do with the `isLoaded` or `isResolutionComplete` state variables in `OSCALLoader.js`

kylelaker:
More broadly, this bug is triggered any time the "Yikes!" text appears.

PR #286, opened 2022-03-01
Handle errors during URL fetching

kylelaker:
Because `OSCALLoader` itself is the top-level component that has an
`<ErrorBoundary>` as a child, we cannot rely on the `<ErrorBoundary>` to
catch errors further in the stack. Additionally, Error Boundaries do a
poor job of catching asynchronous errors. Instead, we track error state
(which we did prior to https://github.com/EasyDynamics/oscal-react-library/pull/267, though, this implementation is a little
cleaner as we don't try to pass the `onError` further down the call
stack since we _can_ use the `<ErrorBoundary>` for that.

Since we've refactored the logic to handle the error into its own
(reused) function, it's now also trivial to set the `loaded` and
`isResolutionComplete` variables, which means we can also re-enable the
"Reload" button even after an error.

Resolves https://github.com/EasyDynamics/oscal-react-library/issues/283
Resolves https://github.com/EasyDynamics/oscal-react-library/issues/163

### DIFF
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -10,7 +10,10 @@ export function BasicError(props) {
   const message = props.error.message?.trim();
   return (
     <Alert severity="error">
-      Yikes! Something went wrong loading the OSCAL data. ({message})
+      <details>
+        <summary>Yikes! Something went wrong loading the OSCAL data.</summary>
+        {message}
+      </details>
     </Alert>
   );
 }
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -6,10 +6,11 @@ import Alert from "@material-ui/lab/Alert";
  * message using a Material UI Alert.
  */
 export function BasicError(props) {
+  // Cleanup the message to prevent potential weird spacing around the parenthesis
+  const message = props.error.message?.trim();
   return (
     <Alert severity="error">
-      Yikes! Something went wrong loading the OSCAL data. Sorry, we&apos;ll look
-      into it. ({props.error.message})
+      Yikes! Something went wrong loading the OSCAL data. ({message})
     </Alert>
   );
 }
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -1,31 +1,57 @@
 import React from "react";
 import Alert from "@material-ui/lab/Alert";
 
-function defaultFormatter(error) {
+/**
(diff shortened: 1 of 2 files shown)


kylelaker:
An earlier version of this work _did_ leverage [e7c9642](https://github.com/EasyDynamics/oscal-react-library/pull/286/commits/e7c9642fda7602bbcdebc28a3321c075dc4fbf3c) but this current implementation doesn't. I am going to rebase this work to split that commit into its own PR and cleanup the history on this PR.

kylelaker:
Samples of a few different error types; both ones fixed by #267 and ones that worked before (and some that worked after).

404 URL:
image

No CORS:
image

200 with a non-JSON response (`fetch` interprets all non-absolute URIs as relative):
image

SSP passed to the Catalog Viewer:
image

hreineck (approved):
Nice fix!

src/components/ErrorBoundary.js
```diff
@@ -1,31 +1,57 @@
...
+/**
+ * A basic FallbackComponent for the {@link ErrorBoundary}. Displays a "yikes"
+ * message using a Material UI Alert.
+ */
+export function BasicError(props) {
   return (
     <Alert severity="error">
       Yikes! Something went wrong loading the OSCAL data. Sorry, we&apos;ll look
-      into it. ({error.message})
+      into it. ({props.error.message})
```
rgauss, line 12: While we're here, can we remove the "Sorry, we'll look into it." sentence please?

rgauss (approved):
Minor, unrelated suggestion.

src/components/ErrorBoundary.js
```diff
@@ -1,31 +1,58 @@
...
+ * message using a Material UI Alert.
+ */
+export function BasicError(props) {
+  // Cleanup the message to prevent potential weird spacing around the parenthesis
+  const message = props.error.message?.trim();
   return (
     <Alert severity="error">
-      Yikes! Something went wrong loading the OSCAL data. Sorry, we&apos;ll look
-      into it. ({error.message})
+      Yikes! Something went wrong loading the OSCAL data. ({message})
```
kylelaker, line 13: @rgauss I kept this simple. But do you have any thoughts on:

```suggestion
      <details>
        <summary>Yikes! Something went wrong loading the OSCAL data.</summary>
        {message}
      </details>
```

It'd look a little like:

image

image

image

image

image

rgauss: Sure, that looks good.

PR #290, opened 2022-03-01
Fix release workflow

kylelaker:
This is presently failing for feature builds with a message about how
the working directory is dirty.


hreineck (approved):

kylelaker:
This won't be sufficient. See #294

PR #293, opened 2022-03-01
Bump version to 0.4.0-0

kylelaker:
This sets the prelease field to `-0` to signify that it is a prerelease
of `0.4.0`. The next tagged version will be `0.4.0`.

### DIFF
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@EasyDynamics/oscal-viewer",
-  "version": "0.3.0",
+  "version": "0.4.0-0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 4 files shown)


folksgl:
Are the '-<number' versions going to be made into GitHub Releases? If so just be doubly-sure you're marking the release as a pre-release as well. I've seen large teams mark release candidates as proper releases and it doesn't make the consumers very happy.

kylelaker:
> Are the '-<number' versions going to be made into GitHub Releases? If so just be doubly-sure you're marking the release as a pre-release as well. I've seen large teams mark release candidates as proper releases and it doesn't make the consumers very happy.

Probably not? It depends. I am writing a new ticket now to handle bugs that currently exist in our automated release workflow; I will make sure to mention this suggestion as part of that ticket.

rgauss (approved):

PR #303, opened 2022-03-02
Avoid linting the entire codebase on every PR

mikeisen1:


kylelaker (approved):
MUCH BETTER! 3min is _way_ better than 10min!

rgauss (approved):

PR #315, opened 2022-03-16
Create UI for adding a control implementation

mikeisen1:
As part of editing capabilities, we want to be able to add a new control implementation to an existing SSP. This PR creates the UI for doing so as well as sends the POST request to a backend service which then adds, if the request is valid, a new control implementation to an SSP.

Closes EasyDynamics/easygrc#28


mikeisen1:
There are some things that still need to be done, but I wanted to create a draft PR so we could get the discussion started on the UI implementation.

This PR still needs to:
1. Test the new `OSCALControlImplementationAdd` component
2. Create storybook documentation of the `OSCALControlImplementationAdd` component
3. Send the appropriate REST request to a backend service and handle the response
4. Validate the user entered control implementation

One thing we could consider is using the `multiple` attribute of the `Autocomplete` Material UI component to allow a user to ad multiple new control implementations at once. This was intentionally left out for now, but is something we should consider adding. Doing so would require an update to the REST specification.

src/components/OSCALControlImplementationAdd.js
```diff
@@ -0,0 +1,122 @@
...
+                props.implementedRequirements[
+                  props.implementedRequirements.length - 1
+                ],
+                () => {},
+                () => {
+                  props.onRestSuccess();
+                },
+                (error) => {
+                  props.onRestError(error);
+                }
```
kylelaker, line 109:
```suggestion
                () => {},
                props.onRestSuccess,
                props.onRestError,

```

src/components/OSCALControlImplementationAdd.js
```diff
@@ -0,0 +1,122 @@
...
+// calculates possible new controls
+function getControlIdsAndTitles(controls, implementedControls) {
+  return controls
+    .filter(
+      (control) =>
+        !implementedControls.find(
+          (implementedControl) => implementedControl === control.id
+        )
+    )
+    .map((control) => `${control.id.toUpperCase()} ${control.title}`);
```
kylelaker, line 23:
```suggestion
  return controls
    .filter((control) => !implementedControls.contains(control.id))
    .map((control) => `${control.id.toUpperCase()} ${control.title}`);
```

kylelaker (changes requested):
I haven't tested this yet but there were a few things I wanted to mention early rather than waiting

src/components/OSCALControlImplementation.js
```diff
@@ -52,26 +65,37 @@ export default function OSCALControlImplementation(props) {
...
+        {props.isEditable ? (
+          <Grid item className={classes.OSCALControlImplementationAdd}>
+            <OSCALControlImplementationAdd
+              controls={props.controls}
+              implementedControls={controlIds}
+              implementedRequirements={implementedRequirements}
+              onRestSuccess={props.onRestSuccess}
+              onRestError={props.onRestError}
+              partialRestData={props.partialRestData}
+              setImplementedRequirements={setImplementedRequirements}
```
rgauss, line 95: Is this only used to display the new implemented requirements while the request is processing?  If so, perhaps we should just display circular progress like the editing existing implementation?

src/components/OSCALControlImplementationAdd.js
```diff
@@ -0,0 +1,120 @@
...
+            }}
+            options={Object.keys(controlIdsAndTitles)}
+            renderInput={(params) => (
+              <TextField {...params} label="Select Control" />
+            )}
+          />
+        </Grid>
+      ) : null}
+      <Grid item>
+        <OSCALEditableFieldActions
```
rgauss, line 67: It feels like we're overcomplicating things and overloading `OSCALEditableFieldActions` again.  The use of that component should probably be restricted to editing simple single fields rather than attempting to accommodate larger object changes/additions like this.

I'd recommend considering a simpler approach like the [saving of existing implemented requirements](https://github.com/EasyDynamics/oscal-react-library/blob/develop/src/components/OSCALControlProse.js#L492).  I think that might also simplify some of the logic currently in `onFieldSave` below.

src/components/oscal-utils/OSCALRestUtils.js
```diff
@@ -238,3 +238,34 @@ export function createOrUpdateSspControlImplementationImplementedRequirementStat
...
+ * Creates and performs the REST request for adding a new SSP control implementation
+ * implemented requirement from the data.
+ *
+ * @param {*} partialRootRestData the partial root SSP object
+ * @param {*} newImplementedRequirement the new implemented requirement to add
+ * @param {*} onPreRestRequest function called just before making the REST request
+ * @param {*} onSuccess function called on a successful REST request with the result of the request as an argument
+ * @param {*} onError function called on error with the error as an argument
+ */
+export function createNewSspControlImplementation(
```
rgauss, line 252: We're technically creating a new implemented requirement, so to follow precedence above would recommend a more specific name of `createSspControlImplementationImplementedRequirement`.

rgauss (changes requested):

src/components/OSCALControlImplementation.js
mikeisen1: This is used to allow a user to view the implemented requirement prior to saving.

src/components/OSCALControlImplementation.js
```diff
@@ -52,26 +65,37 @@ export default function OSCALControlImplementation(props) {
...
+                    onRestError={props.onRestError}
+                    partialRestData={props.partialRestData}
+                  />
+                ))}
               </List>
             </Grid>
           </Grid>
         </CardContent>
+        {props.isEditable ? (
+          <Grid item className={classes.OSCALControlImplementationAdd}>
```
rgauss, line 87: I think this entire component should be aligned to the right.

src/components/OSCALControlImplementation.js
rgauss: We should consider simply making the REST request on click of what is currently the check button (changing that icon to save).

I think that change would lead to a better UX and simpler implementation.

src/components/OSCALControlImplementationAdd.js
```diff
@@ -0,0 +1,120 @@
...
+    });
+  return controlIdsAndTitles;
+}
+
+export default function OSCALControlImplementationAdd(props) {
+  const [inEditState, setInEditState] = useState(false);
+  const [isProcessingRequest, setIsProcessingRequest] = useState(false);
+  const [newControl, setNewControl] = useState("");
+  const oldImplementedRequirements = useRef(props.implementedRequirements);
+  const editIcon = <AddBoxIcon />;
```
rgauss, line 31: With the current control/component navigation it's not clear what this button will do.

I'd suggest using a text button with "Add Control" text, perhaps using the secondary color.

rgauss (changes requested):

rgauss (approved):
Awesome work!!

I pushed a minor change around spacing and placement.

kylelaker (approved):
This is _so awesome_ to see!! This functionality is just so cool to actually have.

I merged `develop` into this branch and applied an equivalent fix to the one that was added in #391.

I also wrote a note on how we could refactor something to be a little more functional-style but that's not really important; just wanted to share in case it was interesting.

PR #338, opened 2022-04-01
Restore eslint-plugin-jest to compatible version

kylelaker:
We are using `create-react-app@^6.0.0` which has a peer dependency on
`eslint-plugin-jest@^24.0.0`. We were installing something like 26.1.3
which is outside of that spec. That resulted in dependency resolution
(and therefore, build) failures. This downgrade ensures that all our
plugins fall within a resolveable set of dependencies. The
`package-lock.json` is regenerated as part of this via deleting
`package-lock.json` and running `npm install` which may result in some
incidental changes to other packages' versions.

Oddly, these failures didn't start until more recently _after_ the PR
was merged; however, this does seem to be the root cause.


zclarkEDC:
These checks don't seem to pass, maybe you should try recreating the `package-lock.json`

kylelaker:
> These checks don't seem to pass, maybe you should try recreating the `package-lock.json` @kylelaker

Good idea! I hadn't thought of that!

> The `package-lock.json` is regenerated as part of this via deleting `package-lock.json` and running `npm install` which may result in some incidental changes to other packages' versions.

Oh wait... 😛 

But actually, I've got no idea what's going on here. It works for me locally but it's failing in the pipeline.

zclarkEDC (approved):

Issue #321, opened 2022-03-23
`dependency-pr` workflow job condition never evaluates to true


kylelaker:
In https://github.com/EasyDynamics/oscal-react-library/actions/workflows/dependency-pr.yml, we can see that the workflow itself is being triggered but the _job_ is being skipped. This implies that the condition isn't quite right at at

https://github.com/EasyDynamics/oscal-react-library/blob/46e2909c1f0b233056f7bfb32fae36be93507fe2/.github/workflows/dependency-pr.yml#L16

We've already enabled debug log output for this workflow; however, it doesn't seem that condition evaluation for un-run jobs gets logged.

### Acceptance Criteria

 - The existing `update` job within the `dependency-pr` workflow properly gets executed when the a PR modifying the top-level `package-lock.json` file in the repository gets merged to `develop`.

### Implementation Notes

We may want to consider adding another job that dumps `${{ github.event }}` as a JSON object to stdout and see what the values we're using actually _are_.

kylelaker:
lol well now it evaluates to true but it doesnt actually trigger jobs on the subsequent run... maybe it'd be better to just use on `on.schedule.cron` trigger and just do this daily.

kylelaker:
Okay so we probably want to look at https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens so that we can attach this to the organization. Then, we can rework the workflow to use the App Token to create PRs. The App route is probably best because it allows for attaching to the organization instead of to a person and because it allows `on: pull_request_target`.

naiqbal:
@rgauss Can we move this to done since it was closed?

PR #340, opened 2022-04-01
Use `pull_request` event for dependency-pr flow

kylelaker:
The GitHub docs say that the event is `pull_request_target`; however,
that does not exist as an object on the event. The `pull_request` object
is there and we should use that.

Related to #321

### DIFF
--- a/.github/workflows/dependency-pr.yml
+++ b/.github/workflows/dependency-pr.yml
@@ -17,10 +17,15 @@ jobs:
     steps:
       - name: Dump GitHub event
         run: |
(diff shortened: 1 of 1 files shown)


kylelaker:
Opened a PR to fix the docs as well at https://github.com/github/docs/pull/16752

kylelaker:
If this actually fixes our issues, I will remove the `debug` job in a subsequent PR.

zclarkEDC (approved):

rgauss (approved):

PR #351, opened 2022-04-05
Separation of Model/REST code from Rendering

rgauss:
@mikeisen1, as mentioned in standup, this is the start of pulling OSCAL model and REST handling code out of UI rendering and into utils, which can start to form sort of a pure Javascript SDK for the REST API.

It also starts to better define the responsibilities of functions and components, limiting what is passed around between them.

Note that this does not yet address the actual control editing as there are some other issues to discuss/resolve there, but it should show a path to how we can restructure the model and REST interaction there.


src/components/OSCALControl.js
```diff
@@ -94,7 +94,6 @@ export default function OSCALControl(props) {
               isEditable={props.isEditable}
               onFieldSave={props.onFieldSave}
               partialRestData={props.partialRestData}
-              restPath={props.restPath}
```
rgauss, line 97: OSCAL object renders shouldn't need to be passed a `restPath`.  The component knows what `oscalObjectType` it's rendering and can pass that to REST functions if needed.

kylelaker:
The installation failures here were, I think, fixed in #338. Let me know if that seems to not be the case.

src/components/OSCALEditableFieldActions.js
```diff
@@ -18,14 +18,12 @@ export default function OSCALEditableFieldActions(props) {
             : `save-${props.editedFieldPath}`
         }
         onClick={() => {
-          if (props.onFieldSave.length === 6) {
+          if (props.onFieldSave.length > 0) {
```
rgauss, line 21: I haven't looked in to when we'd call `onFieldSave` without arguments, but we shouldn't pin to the number of arguments if we can avoid it as that may change.

src/components/OSCALLoader.js
```diff
@@ -13,41 +13,6 @@ import OSCALProfile from "./OSCALProfile";
 import OSCALLoaderForm from "./OSCALLoaderForm";
 import OSCALJsonEditor from "./OSCALJsonEditor";
 
-const oscalObjectTypes = {
```
rgauss, line 16: Moved this to `OSCALRestUtils`

src/components/OSCALLoader.js
```diff
@@ -120,37 +85,16 @@ export default function OSCALLoader(props) {
...
-   * @param partialRestData data that will be passed into the body of the REST request, may not initially contain the updates
-   * @param editedFieldJsonPath path to the field that is being updated
-   * @param newValue updated value for the edited field
-   * @param httpMethod the HTTP request type
-   * @param restUrlPath path defining where in the file the modifications are made
-   * @param jsonRootName root OSCAL object, as it appears on the corresponding object file, of the JSON file
-   * @param restPath main url path for access the OSCAL files in REST mode
-   */
-  const handleRestRequest = (
+  const handleFieldSave = (
```
rgauss, line 88: Something named `handleRestRequest` should really only do that, but here we had a single function manipulating the data, building the request, and performing the request, and defining behaviors for success and failure.

`handleFieldSave` is still responsible for all of that, but delegates isolated bits of that functionality to other functions.

src/components/OSCALLoader.js
```diff
@@ -4,7 +4,7 @@ import Split from "react-split";
 import { makeStyles } from "@material-ui/core/styles";
 import { Box, Fab } from "@material-ui/core";
 import CodeIcon from "@material-ui/icons/Code";
-import { populatePartialRestData } from "./oscal-utils/OSCALUtils";
+import { populatePartialRestData, buildRestRequestUrl, performRestRequest, restMethods, oscalObjectTypes } from "./oscal-utils/OSCALRestUtils";
```
kylelaker, line 7: Maybe:

```ts
import * as restUtils from "./oscal-utils/OSCALRestUtils";
```

Or maybe a better local name than `restUtils`. It'd require some refactoring to use the namespace. But it also means that we could rename the actual functions to be `populatePartialData`, `buildRequestUrl`, `performRequest`, and `methods` since the whole "REST" thing is implied by the file/module name.

src/components/OSCALLoader.js
```diff
@@ -358,32 +301,28 @@ export function OSCALCatalogLoader(props) {
...
-    handleRestRequest
+    handleFieldSave
   ) => (
     <OSCALCatalog
       catalog={oscalData[oscalObjectType.jsonRootName]}
       parentUrl={oscalUrl}
       onResolutionComplete={onResolutionComplete}
       onFieldSave={(
         appendToLastFieldInPath,
-        data,
```
rgauss, line 369: We want to keep naming as explicit as possible for others maintaining the code.

src/components/oscal-utils/OSCALRestUtils.js
```diff
@@ -0,0 +1,214 @@
...
+) {
+  let url;
+  if (!restUrlPath || restUrlPath === "") {
+    url = `${process.env.REACT_APP_REST_BASE_URL}/${oscalObjectType.restPath}/${partialRestData[oscalObjectType.jsonRootName].uuid}`;
+  } else if (restUrlPath.startsWith("http", 0)) {
+    url = restUrlPath;
+  } else {
+    url = `${process.env.REACT_APP_REST_BASE_URL}/${restUrlPath}`;
+  }
+  return url;
```
kylelaker, line 108: Maybe we can actually build a [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL)? That way we can actually specify paths vs host etc? This may just be a conversation for a later ticket.

kylelaker (commented):
Quick thoughts on a first look

src/components/oscal-utils/OSCALRestUtils.js
```diff
@@ -0,0 +1,214 @@
...
+ */
+ export const restMethods = {
+  DELETE: "DELETE",
+  GET: "GET",
+  PATCH: "PATCH",
+  POST: "POST",
+  PUT: "PUT",
+};
+
+export const oscalObjectTypes = {
```
rgauss, line 12: It's debatable whether the `name`, `defaultUrl`, and `defaultUuid` really belong here.  We may want to pull them out and into another enum-like object in a different file.

src/components/oscal-utils/OSCALRestUtils.js
```diff
@@ -0,0 +1,214 @@
...
+ * @param implementedRequirement the existing implemented requirement object
+ * @param statementId the id of the statement to be updated
+ * @param componentId the id of the by-component to be updated
+ * @param descriptionReference reference to the text field input containing the new description for the control implementation
+ * @param implementationSetParameters an array of the new set-parameters
+ * @param onPreRestRequest function called just before making the REST request
+ * @param onSuccess function called on a successful REST request with the result of the request as an argument
+ * @param onError function called on error with the error as an argument
+ */
+ export function updateSspControlImplementationImplementedRequirementStatementByComponent(
```
rgauss, line 171: This isn't yet called from anywhere, but shows how we could continue the pattern of isolating the model/data manipulation and REST calls from the rendering, where this would be called from `OSCALControlPartEditor`.

src/components/oscal-utils/OSCALRestUtils.js
rgauss: Yeah, this iteration was just moving the code over into REST utils.  Feel free to repeat this comment on the main PR, or create a separate issue.

src/components/oscal-utils/OSCALRestUtils.js
mikeisen1: I'd say it makes sense to pull them into a separate file, but only when we break `OSCALRestUtils` down into more files to separate out different aspects of the REST requests or OSCAL types and generic OSCAL utils (which could also contain the `deepClone` function). We are not at that point right now, but should note that they are not similar enough to other functions/enums in this file to keep them together in the same file long-term.

src/components/oscal-utils/OSCALRestUtils.js
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * Creates and performs the REST request for updating an SSP control implementation
+ * implemented requirement statement by compontent from the given data.
+ *
+ * @param partialRootRestData the partial root SSP object
+ * @param implementedRequirement the existing implemented requirement object
+ * @param statementId the id of the statement to be updated
+ * @param componentId the id of the by-component to be updated
+ * @param descriptionReference reference to the text field input containing the new description for the control implementation
```
mikeisen1, line 165: Could we consider revising this function, as well as renaming the parameter, so that `descriptionReference` is not a reference to a text field input but rather just a string value that is a description for the control implementation? It would allow for greater flexibility for the function and does not force a developer to pass in a reference to an input field.

rgauss: Addressed this.

rgauss:
> Lines 41 and 116 in `OSCALControlPartEditor` still rely on the `props.restPath` that was removed as part of this PR. We could get the appropriate oscalObjectType by using the `props.partialRestData` or by passing it in as you mentioned in a previous comment.

@mikeisen1, addressed this.

src/components/OSCALLoader.js
```diff
@@ -415,24 +358,20 @@ export function OSCALSSPLoader(props) {
...
-        editedField,
+        partialRestData,
+        editedFieldJsonPath,
         newValue,
-        restMethod,
         restUrlPath
       ) => {
```
mikeisen1, line 365: The number of arguments here do not match up with the `props.onFieldSave` that is called in `OSCALEditableFieldActions`, which has support for 0 or 4 arguments.

mikeisen1 (changes requested):

src/components/OSCALLoader.js
rgauss: @mikeisen1, right, we're passing in a function that can take 5 arguments, but `OSCALEditableFieldActions` is only calling that function with 4 arguments, so `restUrlPath` will be null and `handleFieldSave` calls `restUtils.buildRequestUrl` which will construct the url from the `partialRestData` and `oscalObjectType`.

That works for simple field edits (as implied by the `OSCALEditableFieldActions` component name), and more complex operations like control editing would probably do better calling more specific REST util methods like `updateSspControlImplementationImplementedRequirementStatementByComponent` rather than trying to handle those scenarios with an overloaded `onFieldSave`, but again, I was trying to minimize the changes in this suggestion so did not yet wire that in to `OSCALControlPartEditor`

mikeisen1 (dismissed):

rgauss:
@mikeisen1, I merged `develop` into `edit-control-implementations-ui` and `edit-control-implementations-ui` into here, resolving conflicts (rebase was not behaving for some reason).

mikeisen1 (approved):

rgauss:
As mentioned in standup, merging this with 1 approval since it's just a suggestion PR.

PR #359, opened 2022-04-17
Refactor dependency update workflow

### LINKED ISSUE #321: `dependency-pr` workflow job condition never evaluates to true

kylelaker:
Okay so I've made a PR with the same name probably a few times now but
this time, I think I've really done it! This totally refactors the way
we do dependency updates and may (eventually) limit our need to use
Dependabot for _npm_ packages (we'd still need it for other types of
updates). This new workflow runs daily and handles dependency updates in
the root package as well as the example app (so they should no longer
get out of sync).

Notably, this does (for now) require a PAT to open the PR. Eventually,
I'd like to move to using a GitHub App but this feels like a pretty good
starting point.

We also limit React updates to minor versions. Apparently, React 18 is
out and with Storybook and other things, that just sorta wrecks
dependency resolution at the moment. And I am guessing most major
version updates of React will require non-zero effort and need to be
tracked anyway.

As far as Dependabot, I'd like to see these two run side-by-side for a
little bit, with us preferring to merge PRs from this workflow but
seeing how much we "miss" (and how often this workflow fails).

For an example of what this looks like (albeit, slightly older code):
https://github.com/kylelaker/oscal-react-library/actions/workflows/update-dependencies.yml
and
https://github.com/kylelaker/oscal-react-library/pull/1.

With any luck:

Resolves #321


kylelaker:
Also, I _do_ plan to delete my fork of this repository once this PR is merged and successfully executes on this repo for the first time; but the best way to safely test this action was to do it on my fork (where I can freely play with secrets, change the default branch, and force push all over the place).

kylelaker:
Finally, I want to note that on my fork, it shows _me_ as the PR author because I created the PAT from my own account settings. On this repository, the PR author would continue to be separate from the commit author, and based on whoever "owns" the PAT that is in use (until we move to using an app).

zclarkEDC (approved):

kylelaker:
Since there's now been at least one approval, I am going to pre-emptively request a PAT on the `edc-github` user.

kylelaker:
> Since there's now been at least one approval, I am going to pre-emptively request a PAT on the `edc-github` user.

Done. This should be usable upon merge.

kylelaker:
First workflow run on this repo was https://github.com/EasyDynamics/oscal-react-library/actions/runs/2189920291 and the first generated PR was #363. Going to move forward with deleting the fork.

PR #369, opened 2022-04-22
Use a GitHub App instead of a PAT in workflow

kylelaker:
A GitHub App is easier for multiple users across the enterprise to
manage _and_ grants access to specific repositories and more finely
scoped permissions than what is possible with a PAT associated with a
user in the organization. We also get to choose a cool name for the app.
This app is named "Easy Dynamics OSCAL Automation" which we can rename
at any time. It currently only has access to this repository; however,
we can trivially grant it access to others.

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -56,7 +56,7 @@ jobs:
         # an output. It _does_ register that token as a secret so that it will be
         # filtered from log output automatically
         id: generate-token
-        uses: tibdex/github-app-token@c95b1c441a4dacd6d24f231b9697a5f9a04c607e
+        uses: tibdex/github-app-token@586e1a624db6a5a4ac2c53daeeded60c5e3d50fe
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/update-dependencies.yml
```diff
@@ -51,6 +51,16 @@ jobs:
           npm install
           cd ..
 
+      - name: Login as the automation app
+        # This Action generates a token from the GitHub App and provides it as
+        # an output. It _does_ register that token as a secret so that it will be
+        # filtered from log output automatically
+        id: generate-token
+        uses: tibdex/github-app-token@c95b1c441a4dacd6d24f231b9697a5f9a04c607e
```
kylelaker, line 59: I elected to use a specific commit hash since we are passing the app id and private key to this Action. But we're also passing the generated token to another action that doesn't use a specific hash, so I am happy to change this to just be `@v1` instead and to track the latest version.

rgauss (approved):

kylelaker:
With one approval in place, I went ahead and added the necessary secrets and triggered a `workflow_run` based off this feature branch. The result of that execution is https://github.com/EasyDynamics/oscal-react-library/runs/6132815515?check_suite_focus=true. There was no work to do (since we just updated our dependencies earlier today). But you can see that it successfully got the token.

zclarkEDC (approved):

PR #370, opened 2022-04-22
Add GitHub workflow to automatically assign PRs

kylelaker:
Our repositories require that every pull request have two approvals
before they can be merged. Often, what will happen is a team will be
added as a requested reviewer but then the first person who reviews
performs their review "on behalf of" the team. That's pretty cool but it
also means that there are no longer any assigned reviewers! This further
means that for anyone using the [Review requests][1] tab on GitHub, they
don't see that PR still needs to be reviewed.

This new workflow takes care of that by just re-requesting review from
the team until the review threshold has been met. For simplicity in the
implementation, it pauses asking for reviews once someone has requested
changes (but that matches our branch protection rules anyway).

Because assigning team reviewers requires permissions not available on
the `GITHUB_TOKEN`, creating a workflow like this previously would have
been a bit of a complication but we're now able to leverage the same
GitHub App that we're using for our other workflow automation (without
having to request a PAT).

Now, one might ask “could this be solved by just using a CODEOWNERS
file or passing `team-reviewers` to the `create-pull-request` Action?”
And both of those would solve the first issue: the initial
assignment; however, they don’t handle the continuing review requests
to meet the project’s threshold. The same applies to the various other
Actions that exist to auto-assign reviews to pull requests (and many
can’t support the `team_reviewers` parameter at all).

[1]: https://github.com/pulls/review-requested

### DIFF
--- a/.github/workflows/triage-pr.yml
+++ b/.github/workflows/triage-pr.yml
@@ -64,7 +64,7 @@ jobs:
         # app's elevated permissions as infrequently as possible.
         if: steps.pr-reviews.outputs.review-decision == 'REVIEW_REQUIRED'
         id: generate_token
-        uses: tibdex/github-app-token@v1
+        uses: tibdex/github-app-token@586e1a624db6a5a4ac2c53daeeded60c5e3d50fe
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
--- a/.github/workflows/triage-pr.yml
+++ b/.github/workflows/triage-pr.yml
@@ -49,11 +49,13 @@ jobs:
                 }
               }
             }`;
-            const result = await github.graphql(query, {
+            const variables = {
               owner: context.repo.owner,
               name: context.repo.repo,
               number: context.payload.pull_request.number,
-            });
+            };
+            console.log(`Request variables: ${JSON.stringify(variables)}`);
+            const result = await github.graphql(query, variables);
             const decision = result.repository.pullRequest.reviewDecision;
             console.log(`Review Decision: ${decision}`);
             core.setOutput("review-decision", decision);
--- a/.github/workflows/triage-pr.yml
+++ b/.github/workflows/triage-pr.yml
@@ -1,0 +1,82 @@
+---
+name: Triage pull request
+
+on:
+  pull_request_review:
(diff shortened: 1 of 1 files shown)


.github/workflows/triage-pr.yml
```diff
@@ -0,0 +1,82 @@
...
+        # app's elevated permissions as infrequently as possible.
+        if: steps.pr-reviews.outputs.review-decision == 'REVIEW_REQUIRED'
+        id: generate_token
+        uses: tibdex/github-app-token@586e1a624db6a5a4ac2c53daeeded60c5e3d50fe
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+      - name: Assign reviewers
+        uses: actions/github-script@v6
+        if: steps.pr-reviews.outputs.review-decision == 'REVIEW_REQUIRED'
```
mikeisen1, line 73: My only question regarding this is what is the desired result if one person request changes and then someone else comments on the PR? Won't that lead to a review decision of 'REVIEW_REQUIRED'?

kylelaker: Nope! This checks the _overall_ status of the PR. Which really basically is:

- If _any_ reviews are "Request Changes", then the status is "Changes Requested". End
- If there are not a sufficient number of "Approve"s, then the status is "Approved". End
- The status is "REVIEW_REQUIRED"

kylelaker:
Oh wow!! I didn't realize that it would start operating now! That's interesting!

mikeisen1 (approved):

mikeisen1:
@kylelaker, just thought of this but could we instead make this a reusable workflow/github action? We'll need this ability to re-assign PRs across EasyGRC repositories and it wouldn't make sense to copy this workflow across each repo.

kylelaker:
I talked with @mikeisen1 about this briefly the other day. Obviously, we'll want to eventually find a way to get this to work on _all_ our repositories. Unfortunately, GitHub doesn't really have a builtin a way to do that today; so we'll have to just copy-and-paste it to other repositories for now.

One potentially-interesting option would be to use some sort of webhook-driven model (rather than GitHub Actions) but that would require a server implementation.

hreineck (approved):

PR #384, opened 2022-05-03
Pin react-refresh to 0.11.0

kylelaker:
When there are conflicting versions of `react-refresh`, `npm run start`
seems to lose its mind (well, not that command directly but something
way deep down in `react-scripts` that's responsible for actually running
the dev server).

For now, let's pin it to 0.11.0 so that people can actually develop and
work. And we can evaluate removing this. The conflicting version that
seems to be coming up at the moment is 0.13.0. We probably could have
picked that version too but this is what most the app was resolving and
so that feels safer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10051,15 +10051,6 @@
         "mkdirp": "bin/cmd.js"
       }
     },
-    "node_modules/@storybook/react/node_modules/react-refresh": {
-      "version": "0.11.0",
-      "resolved": "https://registry.npmjs.org/react-refresh/-/react-refresh-0.11.0.tgz",
-      "integrity": "sha512-F27qZr8uUqwhWZboondsPx8tnC3Ct3SxZA3V5WyEvujRyyNv0VYPhoBg1gZ8/MV5tubQp76Trw8lTv9hzRBa+A==",
-      "dev": true,
-      "engines": {
-        "node": ">=0.10.0"
-      }
-    },
     "node_modules/@storybook/react/node_modules/readable-stream": {
       "version": "2.3.7",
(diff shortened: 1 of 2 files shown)


rgauss (approved):

rgauss:
I believe this is the error you'd see:
```
Module not found: Error: You attempted to import /Users/rgauss/Code/oscal-react-library/node_modules/react-refresh/runtime.js which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
You can either move it inside src/, or add a symlink to it from project's node_modules/.
ERROR in ./src/App.js 1:40-127
Module not found: Error: You attempted to import /Users/rgauss/Code/oscal-react-library/node_modules/react-refresh/runtime.js which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
You can either move it inside src/, or add a symlink to it from project's node_modules/.

ERROR in ./src/index.js 1:40-127
Module not found: Error: You attempted to import /Users/rgauss/Code/oscal-react-library/node_modules/react-refresh/runtime.js which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
You can either move it inside src/, or add a symlink to it from project's node_modules/.

ERROR in ./src/reportWebVitals.js 1:40-127
Module not found: Error: You attempted to import /Users/rgauss/Code/oscal-react-library/node_modules/react-refresh/runtime.js which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
You can either move it inside src/, or add a symlink to it from project's node_modules/.

webpack compiled with 3 errors
```

@mikeisen1, @hreineck, @zclarkEDC, this currently blocks development, please review.

zclarkEDC (approved):

PR #389, opened 2022-05-10
Support editing metadata for all document types

kylelaker:
This allows editing metadata for each of the existing document types.
Because the metadata is a totally shared (and root) object and because
all the base objects server-side support a `PATCH` operation, this is a
pretty easy change to patch in. All other changes have to be made via
the JSON editor.

This _is not_ yet ready to be merged. It is going to have conflicts
(well, conflicts really are the best case; more likely, this will
silently break) with other open PRs but this gives a starting point for
making the modifications.

Eventually, we can probably patch through a few other editable types
from an SSP to other documents that reuse those components. But again,
this is easy and its available now so I figured why not go for it.

### DIFF
--- a/src/components/OSCALCatalog.js
+++ b/src/components/OSCALCatalog.js
@@ -10,9 +10,20 @@ export default function OSCALCatalog(props) {
   const classes = useLoaderStyles();
   props.onResolutionComplete();
 
+  const partialRestData = {
+    catalog: {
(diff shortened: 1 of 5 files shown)


kylelaker:
If we decide to move forward with this, it needs to wait until at least after #277

rgauss (approved):

mikeisen1 (approved):

PR #394, opened 2022-05-16
Use `npm update` instead of deleting package-lock.json

kylelaker:
When we update packages, it's usually a bit better to keep
package-lock.json and to run `npm install && npm update` instead.

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -38,7 +38,6 @@ jobs:
 
       - name: Update project dependencies
         run: |-
-          rm package-lock.json example/package-lock.json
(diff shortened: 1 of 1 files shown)


kylelaker:
The inspiration for this was mostly wanted to make sure that the SHA hash changes we see in #393 would have happened with `npm update` as well (they would). But just using `npm update` directly means that we don't have to wonder if it would have different behavior in the future.

Usually SHA-only changes are a red flag but in #393's case, it's because they're changing from SHA1 to SHA512.

rgauss (approved):

mikeisen1 (approved):

PR #398, opened 2022-05-18
Set app title to "OSCAL Editor" when editing

kylelaker:
This makes it so that when REST Mode is enabled the app is the OSCAL Editor and when it is disabled, it is the OSCAL Viewer. Headers and navigation links are updated accordingly.


kylelaker:
50/50 on whether I am abusing `useEffect` too much so please tell me if there's a better way

example/src/App.js
```diff
@@ -90,6 +91,13 @@ function App() {
       ReactGA.pageview(location.pathname + location.search);
     }
   }, [location]);
+  useEffect(() => {
+    setAppType(isRestMode ? "Editor" : "Viewer");
+  }, [isRestMode]);
+  useEffect(() => {
+    document.title = `OSCAL ${appType}`;
+  }, [appType]);
+
```
mikeisen1, line 100:
```suggestion
useEffect(() => {
    setAppType(isRestMode ? "Editor" : "Viewer");
    document.title = `OSCAL ${appType}`
}, [isRestMode]);
```
Using `useEffect` twice seems unnecessary. Why not do this instead?

mikeisen1 (changes requested):

example/src/App.js
```diff
@@ -90,6 +91,10 @@ function App() {
       ReactGA.pageview(location.pathname + location.search);
     }
   }, [location]);
+useEffect(() => {
+    setAppType(isRestMode ? "Editor" : "Viewer");
+    document.title = `OSCAL ${appType}`
```
kylelaker, line 96:
```suggestion
    document.title = `OSCAL ${appType}`;
```

example/src/App.js
kylelaker: @mikeisen1 so apparently this isn't quite cool with React. See the message on commit [a2d084f](https://github.com/EasyDynamics/oscal-react-library/pull/398/commits/a2d084fc71579ea000721774b704ad41c53ab6f3).

mikeisen1 (approved):

zclarkEDC (approved):

rgauss:
@kylelaker, looks like the End-to-End Cypress tests need to be updated as a part of this.

PR #416, opened 2022-06-03
Add missing install command in release workflow

kylelaker:
The example app uses a script from the root package so we need to make
sure to install the root package's dependencies too.


kylelaker:
Should fix the issue on https://github.com/EasyDynamics/oscal-react-library/runs/6731321353?check_suite_focus=true

zclarkEDC (approved):

rgauss (approved):

PR #420, opened 2022-06-06
Use npm ci

tuckerzp:
Rather than using `npm install` when building the application, we should use `npm ci`. `npm ci` does not update `package-lock.json` and it is a bit better about keeping `node_modules` clean.


Bronstrom (approved):
This deletes the `node_modules` folder and doesn't make changes to `package_lock.json`. Appears to be a reliable change to the project for consistency between our remote repos.

kylelaker (approved):
Nice cleanup! This will be good for consistency.

PR #435, opened 2022-06-10
Add script to build and watch library and app

tuckerzp:
Using the concurrently package found at https://github.com/open-cli-tools/concurrently, this adds the ability to make
changes to components without needing to rerun to viewer to actually see the change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "react-scripts": "react-scripts",
-    "live-edit": "concurrently \"npm run start\" \"cd example && npm run start\""
+    "build-all-and-watch": "concurrently \"npm run start\" \"cd example && npm run start\""
   },
   "eslintConfig": {
     "extends": [


package.json
```diff
@@ -46,7 +47,8 @@
     "run-example": "cd example && npm start",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "react-scripts": "react-scripts"
+    "react-scripts": "react-scripts",
+    "live-edit": "concurrently \"npm run start &\" \"cd example && npm run start\""
```
kylelaker, line 51: Okay so I don't want to bikeshed too much but with all the "build-blah-and-blah-blah" should this be something like `build-all-and-watch`

kylelaker, line 51: Do you need the `&` at the end of the first command here? The examples don't show it from the `concurrently` package.

package.json
```diff
@@ -20,6 +20,7 @@
     "@mui/icons-material": "^5.8.0",
     "@mui/lab": "^5.0.0-alpha.82",
     "@mui/material": "^5.8.0",
+    "concurrently": "^7.2.1",
```
kylelaker, line 23: This needs to be a `devDependency`

package.json
kylelaker: This is actually a question -- not necessarily definitive. Maybe casually poll some folks and see what the general feeling is?

tuckerzp: Sounds good! I was hoping for some input on the name.

tuckerzp: Oh I had copy pasted this command from my terminal, we probably do not need that.

Bronstrom: I'm trying to think of a better name without too much wording as well, but if we want this to follow the other name schemes maybe we name it something like `run-example-and-watch`?

tuckerzp:
I apologize for `force-push` here. I needed to fix the merge conflicts as well as address review comments.

kylelaker:
I don't _love_ pulling in a dependency just for a convenience command but in this case, `concurrently` seems popular enough and I think this command provides sufficient value to justify it.

Issue #270, opened 2022-02-23
Improve Usability of the `OSCALEditableTextField` 


kylelaker:
#As a user of the OSCAL Editor, I want to be able to save or cancel changes via the keyboard so that I can edit the field without a mouse for speed/convenience and potentially accessibility reasons.

When operating in "Rest mode", pressing the `Enter` key in an `OSCALEditableTextField` does not "submit" the change. A user is required to click the save or cancel button.

Pressing "Enter" should properly save that particular field and end editing, as if the "save" button had been pressed. Additionally, pressing the "Esc" should exit editing _without_ saving any changes as if the "Cancel" button had been pressed. Both the buttons should have a Tooltip on hover to clarify their purpose and which field they apply to (simiar to the existing `aria-label` logic).

### Acceptance Criteria
 - Within an `OSCALEditableTextField` pressing "Enter" saves the field, as if the Save button had been clicked
 - Within an `OSCALEditableTextField` pressing "Esc" exits editing of the field, as if the Cancel button had been clicked
 - All iconography in an `OSCALEditableTextField` has a tooltip or similar element (such as perhaps `title` or `<label>`) to clarify the purpose of that button
 - `aria-label` attributes contain human-readable text within `OSCALEditableTextField` and its children
 - End-to-End (Cypress) tests have been created

### Implementation Notes
 - The text field itself is a `@material-ui/core/TextField` which may require using the `onKeyPress` event since the elements are not an `Input` within a `<form>`.
 - We want to avoid copying/pasting code, ideally we'll find a way to implement this functionally by more directly invoking the underlying implementation rather than copying `onClick()` code to an `onKeyPress()` function

### Subtasks
- [x] #424
- [x] #431
- [x] #441
- [x] #442
- [x] #443
- [x] (https://github.com/EasyDynamics/oscal-editor-deployment/issues/97)

PR #438, opened 2022-06-13
Handle Save on Enter Key

### LINKED ISSUE #424: Subtask: OSCALEditableTextField Handle Save on Enter

schen2166:
Allows user to save the state of the text field when pressing the Enter key to save changes

This closes #424 and #270

Discussion Question:
I noticed that the save button keeps the changes for any text box. I was wondering whether 
we should implement a way to keep track of previous entries for the text box because
the text box can be prone to accidental saves where a user can save the textbox value 
unintentionally and lose the data from the text box due to the save.

This could be a potential problem if a text box needs to be reverted to a previous 
text value.

### DIFF
--- a/src/components/OSCALEditableTextField.js
+++ b/src/components/OSCALEditableTextField.js
@@ -31,13 +31,13 @@ function textFieldWithEditableActions(
                 if (event.key === "Escape") {
                   setInEditState(false);
                 } else if (event.key === "Enter") {
+                  event.preventDefault()
                   props.onFieldSave(
                     props.appendToLastFieldInPath,
                     props.partialRestData,
                     props.editedField,
                     reference.current.value
                     );
-                    event.preventDefault()
                     setInEditState(false);
                   }
               }}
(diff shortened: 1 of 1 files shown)


src/components/OSCALEditableTextField.js
```diff
@@ -30,6 +30,14 @@ function textFieldWithEditableActions(
...
                 if (event.key === "Escape") {
                   setInEditState(false);
+                } else if (event.key === "Enter") {
+                  props.onFieldSave(
+                    props.appendToLastFieldInPath,
+                    props.partialRestData,
+                    props.editedField,
+                    reference.current.value
+                  );
+                  setInEditState(false);
```
kylelaker, line 40: Do we need `event.preventDefault()` in here somewhere?

schen2166: > 

From what I see online, the event.Default() method prevents the default action from happening. Example, for a link tag, preventDefault() will prevent the link from opening.

I'm unsure what preventing the Default action of the 'Enter' key would do. Would it it prevent edits in textbox from the previous change? What would adding a `event.preventDefault()` do?

tuckerzp (approved):

Bronstrom (approved):

PR #439, opened 2022-06-14
Eliminate unique key prop errors

### LINKED ISSUE #425: Subtask: Unique key prop errors

tuckerzp:
This should close #425

### DIFF
--- a/src/components/OSCALControlProse.js
+++ b/src/components/OSCALControlProse.js
@@ -228,10 +228,9 @@ function getParameterLabelSegment(
   );
 
   if (!constraintsDisplay) {
-    // This throws an error without fragment wrapper
-    // eslint-disable-next-line react/jsx-no-useless-fragment
-    return <>{paramSegment}</>;
+    return paramSegment;
   }
+
   return (
     <SegmentTooltipWrapper
       constraintsDisplay={constraintsDisplay}
@@ -303,13 +302,13 @@ export function OSCALReplacedProseWithParameterLabel(props) {
         .split(RegExp(prosePlaceholderRegexpString, "g"))
         .map((segment, index) => {
           if (index % 2 === 0) {
-            return getTextSegment(segment, index.toString());
+            return getTextSegment(segment, index);
           }
           return getParameterLabelSegment(
             props.parameters,
             segment,
             props.modificationSetParameters,
-            index.toString()
+            index
           );
         })
     : props.prose;


src/components/OSCALControlProse.js
```diff
@@ -420,6 +419,10 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
         setParameter.values ??= [null];
         const setParameterRef = setParametersRefs[segment];
         // TODO - support for more than 1 item in values arrays
+
+        // Due to a lack of a good value for a key, the index is being used
+        // as a last resort.
+        /* eslint-disable react/no-array-index-key */
```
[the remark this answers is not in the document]
kylelaker: I _really_ dislike these magic comment blocks of `eslint-disable` and a matching `eslint-enable` because it makes it too easy to disable the rule for the rest of the file. Instead, I think I'd like to see us just apply the lint to the single line where the linter complains. Generally the block style of these is best when you have a bunch of violations of a rule in a relatively small block. We have one violation in a big block.

src/components/OSCALControlProse.js
```diff
@@ -428,8 +431,10 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
             size="small"
             inputRef={setParameterRef}
             defaultValue={setParameter.values[0]}
+            key={index}
```
kylelaker, line 434:
```suggestion
            // Due to a lack of a good value for a key, the index is being used
            // as a last resort.
            // eslint-disable-next-line react/no-array-index-key
            key={index}
```

kylelaker (changes requested):

src/components/OSCALControlProse.js
tuckerzp: > I really dislike these magic comment blocks of `eslint-disable`

Completely agree on this. Having spent a lot of time deleting places where we had them, I was hesitant to use it. I was unaware that you could just place `// eslint-disable-next-line` the `TextField` attributes, but I like that a lot more.

kylelaker: You might need something like

```js
{
  // Explanation
  // eslint-disable-next-line react/no-array-index-key
}
key={index}
```

when inside one of the JSX blocks. Like we do [here](https://github.com/EasyDynamics/oscal-react-library/blob/develop/example/src/App.js#L110-L115)

src/components/OSCALControlProse.js
```diff
@@ -420,6 +419,10 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
         setParameter.values ??= [null];
         const setParameterRef = setParametersRefs[segment];
         // TODO - support for more than 1 item in values arrays
```
kylelaker, line 421: If you have the familiarity with the surrounding code to do so (and because this ticket rolls up a general tech investments epic), can you convert this TODO to a GitHub issue (labeled as an enhancement) while you're in here?

kylelaker (approved):

Bronstrom (approved):

PR #440, opened 2022-06-14
Auto-assign reviewers only on ready-for-review PRs

tuckerzp:
This updates the triage-pr workflow to ensure that reviewers will only be
assigned to prs that are not a draft.


kylelaker:
So https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request_review doesn't necessarily show that `draft` appears on the `github.event.pull_request` when the event type is `pull_request_review`.

We can test this on another repo or we can add `isDraft` to the GraphQL query and outputs from that step.

kylelaker:
We'd also need to add the `ready_for_review` event type to the `on.pull_request_target.types` list.

kylelaker (approved):

PR #447, opened 2022-06-15
Add aria label to OSCALEditableTextField

nuttercd:
To increase user accessability it is important to have aria lables

### DIFF
--- a/src/components/OSCALEditableTextField.js
+++ b/src/components/OSCALEditableTextField.js
@@ -17,6 +17,7 @@ function textFieldWithEditableActions(
         <Grid item xs={props.size} className={props.className}>
           <Typography>
             <TextField
+              aria-label="Text"
               fullWidth
               inputProps={{
                 "data-testid": `textField-${getElementLabel(
--- a/src/components/OSCALEditableTextField.js
+++ b/src/components/OSCALEditableTextField.js
@@ -17,7 +17,7 @@ function textFieldWithEditableActions(
         <Grid item xs={props.size} className={props.className}>
           <Typography>
             <TextField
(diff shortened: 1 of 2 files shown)


src/components/OSCALEditableTextField.js
```diff
@@ -17,6 +17,7 @@ function textFieldWithEditableActions(
         <Grid item xs={props.size} className={props.className}>
           <Typography>
             <TextField
+              aria-label="Text"
```
kylelaker, line 20: This needs to better explain what's going on here -- "Text" isn't especially meaningful. This should describe the purpose of the field, like what we'd put in a `<label>` element (and if we have a `<label>` element, we can use `aria-labelledby`)

kylelaker (changes requested):
Great work! But I think that there's more that we need to do here. Let's discuss a bit and see if we can make this even better

src/components/OSCALEditableTextField.js
nuttercd: Passed the field name as a label instead of using an aria-label

example/package-lock.json
```diff
@@ -1356,41 +1351,13 @@
           "requires": {
             "@babel/runtime": "^7.12.5",
             "@testing-library/dom": "^8.0.0",
-            "@testing-library/jest-dom": "^5.11.6",
```
kylelaker, line 1359: can we get this lil fella outta here? like the whole file? not sure why `package-lock.json` decided to come along for the ride

kylelaker (changes requested):

kylelaker (approved):

PR #448, opened 2022-06-15
Eliminate all descendant of <p> warnings

### LINKED ISSUE #426: Subtask: Descendant of `<p>` errors

tuckerzp:
This will eliminate all descendant of `<p>` related console warnings. To deal
with this issue, there were a couple places where `Typography` tags were
deleted. After running the tests and looking at the viewer, I am fairly certain
it did not break anything. But it would be good to double check that. 

This will close #426

### DIFF
--- a/src/components/OSCALControlProse.js
+++ b/src/components/OSCALControlProse.js
@@ -447,17 +447,15 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
   return (
     <OSCALStatementEditing container spacing={2}>
       <Grid item xs={11}>
-        <Typography className={props.className}>
-          <StyledTooltip
-            title={statementByComponentDescription ?? props.componentId}
-          >
-            <Link underline="hover" href={`#${props.label}`}>
-              {props.label}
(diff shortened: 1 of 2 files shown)


kylelaker:
So while this may not have broken anything visually, does it change the semantic meaning of any these in a breaking way? Like, are there situations now in the viewer or editor where we have text floating around directly inside a `<div>` without a `<p>`?

src/components/OSCALControlProse.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState, useRef } from "react";
+import React, { useState, useRef, Fragment } from "react";
```
kylelaker, line 1: Why is `Fragment` added as an import?

kylelaker (changes requested):

src/components/OSCALControlProse.js
tuckerzp: Fixed!

Bronstrom (approved):

PR #451, opened 2022-06-16
Eliminate warnings in OSCAL Viewer

### LINKED ISSUE #446: Subtask: Eliminate "component is changing an uncontrolled input to be controlled" error
This happens when you choose something from the dropdown in `OSCALLoaderForm` when you're in editing mode. Initial value is `undefined` then it becomes not `undefined`, which changes it to a controlled input.

tuckerzp:
This removes all warnings in the viewer. The issues in the viwer stemmed from
issues with how we handle state and places where we did not use `useEffect()`.

Hopefully after this pr, future pr's will not allow more warnings into the
viewer.

This will close #446

### DIFF
--- a/src/components/OSCALCatalog.js
+++ b/src/components/OSCALCatalog.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import List from "@mui/material/List";
 import ListSubheader from "@mui/material/ListSubheader";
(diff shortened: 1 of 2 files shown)


kylelaker (approved):
This is awesome!! Great catches!

Bronstrom (approved):

PR #455, opened 2022-06-17
Output console messages to job summary

tuckerzp:


.github/workflows/pr.yml
```diff
@@ -64,13 +63,15 @@ jobs:
       - name: Build library
         run: npm run build
       - name: Run tests
-        run: npm run test -- --ci --watchAll=false
+        run: |
+          TMPFILE=`mktemp -t output.XXXXXXXXXXXXXX` && {
+            npm run test -- --ci --watchAll=false 2>&1 | tee $TMPFILE
+            grep -A 4 "Console" $TMPFILE >> $GITHUB_STEP_SUMMARY
+          }
```
kylelaker, line 70:
```suggestion
          TMPFILE="$(mktemp -t output.XXXXXXXXXXXXXX)"
          npm run test -- --ci --watchAll=false 2>&1 | tee "$TMPFILE"
          grep -A 4 "Console" "$TMPFILE" >> $GITHUB_STEP_SUMMARY 
```

kylelaker (commented):
On the whole I like this but with the addition of the the `example/` directory, if there are warnings in the library _but not_ the app, you'll get output like:

~~~
## `/src` Test Warnings Found

```
some
```

```
warnings
```

No warnings found in `/example`
~~~

So I think it'd be a lot easier to emit output like the following:

~~~
## React Library

```
some
```

```
warnings
```

## Viewer Application

No warnings found in `/example/`
~~~

Basically, always emit the `##` headers, use the names we're more familiar with, and make the warnings or the "no warnings" text in the `if`.

tuckerzp:
> Maybe YAML anchors and aliases could be utilized here

@Bronstrom based off of https://github.com/actions/runner/issues/1182, it does not seem that workflows currently support `YAML anchors`. I may be wrong though, did you see an example of this used elsewhere?

Bronstrom (approved):

zclarkEDC (approved):

PR #458, opened 2022-06-21
Add Legend for Parameter Color-Coding

### LINKED ISSUE #195: Add Legend for Parameter Color-Coding
As a user of the OSCAL React components and/or example viewer, I want to be able to see a legend of what the color-coding of parameters within controls and control implementations means so that I can quickly understand the context of parameter placeholders vs parameter values.

### Acceptance Criteria
- A legend is displayed that indicates the color coding of parameter placeholders vs parameter values.
- `react-testing-library` tests are in place

### Implementation Notes
- This could likely just be two rounded color squares of orange and blue with lightly colored text indicating `Placeholder` vs `Value` in the top-right corner of the control.

### Subtasks
- [ ] #427
- [ ] #428

### LINKED ISSUE #428: Subtask: Add `react-testing-library` tests for Legend

Bronstrom:
In order to enhance user experience, a parameter legend is displayed highlighting the difference between parameter placeholders (orange) and parameter values (blue).

This adds a legend to all control grouping schemes: `OSCALCatalogGroup`, `OSCALControlImplementation`, `OSCALComponentDefinitionControlImplementation`, and `OSCALProfile` groups. The component `OSCALControlParamLegend` is built for purposes of avoiding duplicate code. In addition, `react-testing-library` tests are written to determine the legend displays in each of the control groupings.

Resolves #195, resolves #428


src/components/OSCALComponentDefinitionControlImplementation.test.js
```diff
@@ -47,3 +47,23 @@ test("OSCALComponentDefinitionControlImplementation displays component parameter
...
+      components={componentDefinitionTestData.components}
+      controls={controlsData}
+    />
+  );
+
+  const placeholderBox = screen.getByLabelText("legend-placeholder-label");
+  expect(placeholderBox).toBeVisible();
+  const placeholderBoxLabel = screen.getByText("Placeholder");
+  expect(placeholderBoxLabel).toBeVisible();
+
```
zclarkEDC, line 64: Remove this empty line?

src/components/OSCALCatalogGroup.test.js
```diff
@@ -2,37 +2,53 @@ import React from "react";
...
+
+  const placeholderBox = screen.getByLabelText("legend-placeholder-label");
+  expect(placeholderBox).toBeVisible();
+  const placeholderBoxLabel = screen.getByText("Placeholder");
+  expect(placeholderBoxLabel).toBeVisible();
+
+  const valueBox = screen.getByLabelText("legend-value-label");
+  expect(valueBox).toBeVisible();
+  const valueBoxLabel = screen.getByText("Value");
+  expect(valueBoxLabel).toBeVisible();
```
tuckerzp, line 53: This seems very similar to:

https://github.com/EasyDynamics/oscal-react-library/blob/d002755818b29fbc4f357b441836d08741675430/src/components/OSCALComponentDefinitionControlImplementation.test.js#L60-L68

If it is easy to reduce this repetition, it may be worth looking into. I would not waste much time doing it however.

src/components/OSCALCatalogGroup.test.js
```diff
@@ -2,37 +2,53 @@ import React from "react";
...
-      {
-        id: "ac",
-        class: "family",
-        title: "Access Control",
-        controls: [{ id: "ac-1" }],
-      },
-    ],
-  };
   render(<OSCALCatalogGroup group={testGroup} />);
+
```
zclarkEDC, line 30: Remove this empty line?

zclarkEDC (changes requested):
Minor formatting required, otherwise looks good to merge.

tuckerzp (changes requested):
This looks great so far! I noticed that running `npm run test` does produce some warnings. I'd love to get this dealt with before merging into `develop`.

I only looked into it a bit and do not see any obvious issues. From experience, I'd look into places where `<div>` is placed into a `<Typography>`.

src/components/OSCALCatalogGroup.test.js
Bronstrom: This check is being performed on each of the control groupings for the different models. This could be wrapped into a function pulled from one of these test suites or a new test file could be created specifically for these tests. Any suggestions?

src/components/OSCALControlParamLegend.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { styled, useTheme } from "@mui/material/styles";
+import { Box } from "@mui/material";
+
+export default function OSCALControlParamLegend() {
+  const LegendLabel = styled("div")(
```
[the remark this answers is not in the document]
Bronstrom: I think you're right about us falling back on using the `<div>` tag a lot, and maybe to the point where we're overusing them. From our discussion last week, there seems to be a somewhat loose hierarchy of using container-like HTML tags. `<div>` is a container for block content, while `<p>` is a paragraph container. Due to the legend content being more block-like, I choose to use a `<div>`. Although, I could also see a `<span>` working in this case, because the content sits in-line, but these are generally used for marking up text.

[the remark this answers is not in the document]
tuckerzp: Fair point, I guess I was mostly referring to some of these `<div>` tags recreating something already provided by `React`. I did not know if there was already a more descriptive component we could use in some places.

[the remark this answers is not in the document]
Bronstrom: I think for now a styled `<div>` is a decent option for the `LegendLabel`. I have swapped the surrounding `<div>` for a fragment. We could return to this later if we find a better approach.

src/components/OSCALCatalogGroup.test.js
Bronstrom: `ControlParamLegend.test.js` will solve this. This helper test class is used for each of the control groups to reference, and avoids code duplication. In order for this to be a test class, I needed to include a test and supplied a basic `OSCALCatalogGroup` test.

tuckerzp (approved):
Looks good to me! Thanks for your work on this @Bronstrom

zclarkEDC (approved):

PR #462, opened 2022-06-22
Add Issue Template forms

kylelaker:
This uses GitHub issue templates to create forms for new issues. This
follows what we've configured at https://github.com/EasyDynamics/.github
except that doesn't quite work at the organization level yet.

For now, this allows creating "blank" issues so that we're not forcing
people into boxes before understanding the potential missing features of
this addition.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,0 +1,64 @@
+---
+name: Bug Report
+description: Report a bug
+labels:
+  - bug
+body:
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: >-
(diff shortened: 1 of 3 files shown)


tuckerzp (approved):
I really like the idea of using templates for issues! Since I was not super familiar with how to add templates, I found this [doc page](https://docs.github.com/en/github-ae@latest/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) pretty useful.

I also agree giving the option for a blank issue is useful for now.

Bronstrom (approved):
This seems like a great improvement to submitting issues by helping narrow the scope of contributor messages and providing us the details we need when an issue is reported. I also agree with the choice of using "blank" issues.

Issue #466, opened 2022-06-23
Subtask: Tests for `Object Selector` component


tuckerzp:

tuckerzp:
Would we want to test that `OSCAL objects` are listed correctly here or in https://github.com/EasyDynamics/oscal-editor-deployment/issues/98? I am wondering if testing it here and with `cypress` is an unneeded redundancy. After messing around with trying to test it in #480, it may not be worth the effort if testing it in `cypress` would be trivial.

tuckerzp:
@kylelaker #480 & #472 do deal with this subtask. Just not quite as in depth as originally planned. Should this really be closed as unplanned or just closed?

Issue #336, opened 2022-04-01
Support for Friendly/Semantic URLs in Viewer


rgauss:
As a user of the OSCAL viewer app, I want to be able use bookmarkable, friendly/semantic URLs when the app is configured for REST mode so that I can share and bookmark specific OSCAL URLs to be able to quickly view those artifacts without additional navigation steps.

### Acceptance Criteria
- When the app is [configured for REST mode](https://github.com/EasyDynamics/easygrc/issues/6) a URL with the OSCAL object type and UUID can be used to load the corresponding artifact
  - Example: https://oscal-viewer.msd.easydynamics.com/system-security-plan/cff8385f-108e-40a5-8f7a-82f3dc0eaba8
  - A URL path element for the UUID is preferred but a URL parameter would be acceptable

rgauss:
Also see #176

kylelaker:
@tuckerzp Some additional subtasks to consider (or at least things to verify):

 - Update the route mapping in the [oscal-rest-service](https://github.com/EasyDynamics/oscal-rest-service)'s `ViewerAppController`
 https://github.com/EasyDynamics/oscal-rest-service/blob/c105f405cbf7ae6e81a12f87f2388654e3b6c1ed/oscal-rest-service-app/src/main/java/com/easydynamics/oscalrestservice/api/ViewerAppController.java#L9-L12
 This will ensure that the application will continue working when served from the Docker container (without getting errors about failing to map the path `/profile/120054ed-8e7c-4e22-ac1d-7df598736945` For example, today I see 
![image](https://user-images.githubusercontent.com/850893/176564924-1d767f2e-1d26-43e5-bd79-09adc6576bd1.png)
 - Update the infrastructure definition for the live application (right now, paths are hardcoded to known routes) to avoid 404 responses
 - Cypress end-to-end tests for both navigation via the loader form and direct navigation via URL

PR #472, opened 2022-06-24
Add OSCALDrawerSelector component

### LINKED ISSUE #468: Subtask: Create `Object Selector` component
Here is a proposed design of the `OSCALDrawerSelector` functionality:
- Drawer ~grows to a max width of 20% of total screen~ has a width of `250px`
  - Anything name longer than that max width will be truncated.
  - ~Content currently on screen is pushed to the side, rather than being covered.~
- Drawer is vertically scrollable.
- Items displayed will be placed under the following category hierarchy:
```
v easygrc-demo-content
  > catalogs
  > profiles
  v component-definitions
     Test Component Definition 1      // This is the metadata.title attribute of the component definition 
     Test Component Definition 2
  > system-security-plans
```
- Clicking on a directory displays the content within.
- Clicking on a file loads said file in the correct view.

tuckerzp:
This adds the `OSCALDrawerSelector` component to the `OSCAL Editor`. Rather than needing to manually select an `OSCAL Object` using the loader, you can select the file using the `DrawerSelector` on the side. The linking uses the friendly url's implemented in #336.

This will eventually close #468 & #466


src/components/OSCALDrawerSelector.js
```diff
@@ -0,0 +1,56 @@
...
+  alignItems: "center",
+  padding: theme.spacing(0, 1),
+  // necessary for content to be below app bar
+  ...theme.mixins.toolbar,
+  justifyContent: "flex-end",
+}));
+
+export default function OSCALDrawerSelector(props) {
+  const theme = useTheme();
+  const [open, setOpen] = React.useState(false);
```
kylelaker, line 21: We tend to use `is` for boolean variables: `isOpen`/`setIsOpen`

src/components/OSCALDrawerSelector.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const handleDrawerOpen = () => {
+    setOpen(true);
+  };
+
+  const handleDrawerClose = () => {
+    setOpen(false);
+  };
```
kylelaker, line 29:
```suggestion
  const toggleDrawer = () => {
    setOpen((current) => !current);
  };
```

kylelaker (commented):
Quick thoughts just to provide them early on

src/components/OSCALDrawerSelector.js
tuckerzp: Oh that does make more sense. This is really just me copying and pasting the example code on the [React documentation](https://mui.com/material-ui/react-drawer/#persistent-drawer) as a proof of concept but I will keep that in mind.

tuckerzp:
@Bronstrom or anyone else
So this is again more of a proof of concept. Depending on whether or not we are in `RestMode` the proper menu type will display. I've also imported the `OSCALDrawer` to the `example app`. To make sure I am on the right track, I'd love some feedback and have a few questions:

1. It is acceptable for the issue to have a fixed width, but I kinda hate it. Is there an obvious way to deal with this I am missing?
2. Is the `TreeView` within the `Drawer` a nice way to display these files? I personally think it is, but was curious on other's opinions.

Bronstrom:
> It is acceptable for the issue to have a fixed width, but I kinda hate it. Is there an obvious way to deal with this I am missing?

(I'm thinking you're speaking in regards to the user having the ability to resize the drawer by dragging, rather than the drawer resizing based on the window size.) I don't believe there is anything built into the drawer or any Material UI component that allows for the width to be resized by the user. This is functionality we would need to add. We could implement this with a mouse event listener.

>  Is the `TreeView` within the `Drawer` a nice way to display these files? I personally think it is, but was curious on other's opinions.

A `TreeView` would fit our navigation well because we will have a hierarchy of data, such as the relationship between OSCAL model types (higher level) organizing the examples (lower level).

src/components/OSCALDrawerSelector.js
```diff
@@ -0,0 +1,67 @@
...
+        flexShrink: 0,
+        "& .MuiDrawer-paper": {
+          width: drawerWidth,
+          boxSizing: "border-box",
+        },
+      }}
+      variant="persistent"
+      anchor="left"
+      open={props.open}
+    >
```
kylelaker, line 37: We want to do this like you do for `DrawerHeader` above. We should generally _heavily_ avoid the `sx` prop except for in limited, short styles.

```ts
const Drawer = styled(BaseDrawer)`
  width: ${drawerWidth};
  flex-shrink: 0;
  & .MuiDrawer-paper {
    width: ${drawerWidth};
    box-sizing: "border-box";
  }
`;
```

example/src/App.js
```diff
@@ -52,6 +53,7 @@ const LogoImage = styled("img")`
 
 function App() {
   const [anchorEl, setAnchorEl] = useState(null);
+  const [drawerIsOpen, setDrawerIsOpen] = useState(false);
   const [isRestMode, setIsRestMode] = useState(
```
kylelaker, line 57: What do you think about defaulting to having the nav drawer open in REST mode? The likely scenario in REST mode is to operate on various files and options and there isn't a clear default -- whatever we choose is likely to be imperfect and the user will likely want to choose a different option.

Perhaps something like

```ts
const [isDrawerOpen, setIsDrawerOpen] = useState(isRestMode);
```

Not sure if we'd need to do `useState(Boolean(isRestMode))`; I don't think we do and I can't articulate why I think we might.

example/src/App.js
```diff
@@ -52,6 +53,7 @@ const LogoImage = styled("img")`
 
 function App() {
   const [anchorEl, setAnchorEl] = useState(null);
+  const [drawerIsOpen, setDrawerIsOpen] = useState(false);
```
kylelaker, line 56: nitpick: `isDrawerOpen` to match `isRestMode`

example/src/App.js
```diff
@@ -89,6 +93,47 @@ function App() {
...
+        onClick={handleAppNavClose}
+        component={RouterLink}
+        to="/system-security-plan"
+      >
+        {`System Security Plan ${appType}`}
+      </MenuItem>
+    </Menu>
+  ) : (
+    <OSCALDrawerSelector open={drawerIsOpen} handleClose={handleAppNavClose} />
+  );
```
kylelaker, line 135: Hmm... this is a very off-balance ternary operator. Already the complexity seems to have nudged us to extract it to a variable. Perhaps we should then extract the `<Menu>` to its own component? Or maybe the whole thing?

Or maybe it can be solved a bit by just renaming the variable to `navigation` instead of `menu`?

kylelaker (commented):
Again, a few small thoughts that you may have already thought of

example/src/App.js
tuckerzp: Yeah I agree it is gross. I am not against making a `OSCALMenuSelecter` object of some sort. Not sure if that would be the best name for it.

tuckerzp:
@Bronstrom 
> I'm thinking you're speaking in regards to the user having the ability to resize the drawer by dragging, rather than the drawer resizing based on the window size.

Not quite. I am more so referring to the fact that as the tree gets larger, the drawer will not grow to fit it. So you can imagine a case where some of the listed files are not visible. I assume this is what was being refered to by `In this iteration it is acceptable if the pane is a fixed width`. I would argue that the user would not have a need to manually resize the window.

Another thing I would like to happen is that when the drawer opens up, the page itself moves with the drawer. Rather than the functionality now with the drawer just covering the page. I do not know if that would be a simple thing to add however.

kylelaker (approved):
Great stuff! Can you remind me which PR handles removing the dropdown and which handles moving the rest toggle?

src/components/OSCALDrawerSelector.test.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import OSCALDrawerSelector from "./OSCALDrawerSelector";
+
+test("Test Drawer even loads", () => {
```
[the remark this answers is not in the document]
tuckerzp: The drawer component is being created outside of any other context for the tests. So I do not think `REST` mode matters for this particular test.

example/src/App.js
```diff
@@ -88,6 +92,51 @@ function App() {
...
+      >
+        {`Component ${appType}`}
+      </MenuItem>
+      <MenuItem
+        onClick={handleAppNavClose}
+        component={RouterLink}
+        to="/system-security-plan"
+      >
+        {`System Security Plan ${appType}`}
+      </MenuItem>
```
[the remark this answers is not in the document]
tuckerzp: I like that idea.

package.json
```diff
@@ -99,6 +99,7 @@
     "eslint-plugin-react": "^7.30.1",
     "gh-pages": "^4.0.0",
     "microbundle-crl": "^0.13.11",
+    "msw": "^0.43.0",
```
kylelaker, line 102: Please reset `package.json` and `package-lock.json` to match `develop`

src/components/oscal-utils/OSCALRestUtils.js
```diff
@@ -46,6 +46,31 @@ export const oscalObjectTypes = {
...
+      if (!response.ok) throw new Error(response.status);
+      else return response.json();
+    })
+    .then((result) => {
+      setOscalObjects((current) => ({
+        ...current,
+        [oscalObjectType.jsonRootName]: result,
+      }));
+    });
+}
```
kylelaker, line 72: Hmm this feels a little too specific to one use case to be a shared REST util. It depends on knowing a specific structure for the underling data type under `setOscalObjects` and what it looks like (and we don't document that expectation).

I think I'd rather see an interface like the following (I've written it in TypeScript to make the intent a little more clear). Though I've removed error handling in this example to make it a bit clearer.

```ts
interface OscalObjectType {
  restPath: string;
  jsonRootName: string;
}

export function fetchAllResourcesOfType(
  backendUrl: string,
  oscalObjectType: OscalObjectType,
  handleResult: (any) => void,
): void {
  fetch(`${backendUrl}/${oscalObjectType.restPath}`)
    .then((response) => response.json())
    .then((result) => handleResult(result));
}

oscalObjectTypes.forEach((oscalObjectType) => {
  fetchAllResourcesOfType(backEndUrl, oscalObjectType, (result) => {
    setOscalObjects((current) => ({
      ...current,
      [oscalObjectType.jsonRootName]: result,
    });
  });
}
```

kylelaker: This also feels a bit more testable eventually too (because you separate the responsibility for fetching the data from the responsibility for editing state). One could imagine a `handleResult` that supports testing instead.

src/components/OSCALDrawerSelector.test.js
```diff
@@ -0,0 +1,44 @@
...
+      name: "Component",
+    });
+
+    const profileItem = screen.getByRole("treeitem", {
+      name: "Profile",
+    });
+
+    const sspItem = screen.getByRole("treeitem", {
+      name: "System Security Plan",
+    });
```
kylelaker, line 27:
```suggestion
    const catalogItem = screen.getByRole("treeitem", {
      name: "Catalogs",
    });

    const componentItem = screen.getByRole("treeitem", {
      name: "Components",
    });

    const profileItem = screen.getByRole("treeitem", {
      name: "Profiles",
    });

    const sspItem = screen.getByRole("treeitem", {
      name: "System Security Plans",
    });
```

Okay so how much work would it take to make this work?

kylelaker (changes requested):

kylelaker (approved):

src/components/oscal-utils/OSCALProfileResolver.js
```diff
@@ -79,13 +79,13 @@ export default function OSCALResolveProfileOrCatalogUrlControls(
           inheritedOSCALObject.inherited = [];
 
           modifications["set-parameters"].push(
-            ...(result.profile.modify["set-parameters"] ?? [])
+            ...(result.profile.modify?.["set-parameters"] ?? [])
           );
-          modifications.alters.push(...result.profile.modify.alters);
+          modifications.alters.push(...(result.profile.modify?.alters ?? []));
```
kylelaker, line 84: I feel like these didn't come from your branch? What's going on here? Weird rebase/merge? These changes are already present on `develop`

package-lock.json
```diff
@@ -66544,3 +66544,4 @@
     }
   }
 }
+
```
tuckerzp, line 66547: I have tried so many times to get rid of this file.

src/components/oscal-utils/OSCALProfileResolver.js
tuckerzp: Should be fixed.

Bronstrom (approved):
This is in a good place right now! All the required functionality for this component is here. The loader form dropdown and reload button are now unnecessary in REST Mode and will be addressed in [#467](https://github.com/EasyDynamics/oscal-react-library/issues/467). Future style related updates to `OSCALDrawerSelector` component were suggested but I think should be addressed later - I don't mind handling this. Thanks for your work on this @tuckerzp (and @kperk103 on the testing)!

kylelaker (approved):
Thanks for working iteratively on this for making so many modifications. Good stuff!!

PR #478, opened 2022-06-29
Support Friendly URLs

### LINKED ISSUE #336: Support for Friendly/Semantic URLs in Viewer

tuckerzp:
Rather than loading by OscalURL's in the editor, we now load based on
the OSCALObjects uuid. This allows for a lot more coherent way of
routing throughout the editor. It also allows you to load an element by
placing the uuid in the url.

This may render some portions of the `OSCALLoader` redundant. As I still do not
feel completely confident on what everything in the `OSCALLoader` does, please
pay extra attention in case I missed something. 

Testing still needs to be added to this, but the functionality should 
be working. It makes more sense to deal with this testing in https://github.com/EasyDynamics/oscal-editor-deployment/issues/98

Closes #336

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -111,12 +111,14 @@ function App() {
     },
   ];
 
+  const routePaths = ["", ":id"];
+
   const appBarRoutes = (
     <Route path="/">
       <Route index element={`OSCAL Catalog ${appType}`} />
       {oscalObjectTypes.map((oscalObjectType) => (
         <Route path={oscalObjectType.pathName} key={oscalObjectType.pathName}>
-          {["", ":id"].map((path) => (
+          {routePaths.map((path) => (
             <Route
               path={path}
               key={path}
@@ -147,7 +149,7 @@ function App() {
       />
       {oscalObjectTypes.map((oscalObjectType) => (
         <Route path={oscalObjectType.pathName} key={oscalObjectType.pathName}>
-          {["", ":id"].map((path) => (
+          {routePaths.map((path) => (
             <Route
               path={path}
               key={path}
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -5,7 +5,7 @@ import {
   ThemeProvider,
   StyledEngineProvider,
 } from "@mui/material/styles";
-import React, { useEffect, useState } from "react";
+import React, { createElement, useEffect, useState } from "react";
 import Typography from "@mui/material/Typography";
 import Container from "@mui/material/Container";
 import AppBar from "@mui/material/AppBar";
@@ -19,7 +19,6 @@ import MenuItem from "@mui/material/MenuItem";
 import CssBaseline from "@mui/material/CssBaseline";
 import ReactGA from "react-ga";
 import {
-  Navigate,
   Route,
   Routes,
   Link as RouterLink,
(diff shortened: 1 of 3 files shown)


kylelaker (commented):
On the whole... wow! Love it! I think there are some small adjustments we can make to reduce duplication further.

src/components/OSCALLoader.js
```diff
@@ -132,8 +133,8 @@ export default function OSCALLoader(props) {
     setOscalUrl(event.target.value);
   };
 
-  const handleUuidChange = (event) => {
-    const newOscalUrl = `${props.backendUrl}/${props.oscalObjectType.restPath}/${event.target.value}`;
+  const handleUuidChange = (objectUuid) => {
+    const newOscalUrl = `${props.backendUrl}/${props.oscalObjectType.restPath}/${objectUuid}`;
```
kylelaker, line 137: This breaks the API expected by the `OSCALLoaderForm` props correct? See Line 218?

tuckerzp: I may misunderstand what you mean, but `OSCALLoaderForm` was changed to deal with this.

https://github.com/EasyDynamics/oscal-react-library/blob/eb50a0174471509cd83c1068b8a7d0f12f5b3a42/src/components/OSCALLoaderForm.js#L113

`event.target.value` is the `uuid` here.

kylelaker:
Okay so it looks like there is an odd behavior here.

### Steps to Reproduce

1. Launch the editor, go to any OSCAL Document type (for this example, Catalog)
2. Choose a catalog from the drop down
3. Wait for the catalog to load (or don't)
4. Go to the navigation
5. Click the same OSCAL document type (in this example, Catalog)

### Expected Behavior

The page is blank except for the LoaderForm.

### Actual Behavior

The previous document remains on the page.

tuckerzp:
> Expected Behavior
The page is blank except for the LoaderForm.

> Actual Behavior
The previous document remains on the page.

I agree that it doing that would make sense but I am not 100% sure this is an issue that needs to be solved here. #472 will get rid of the ability to click on a `OSCALType` like that. So going to just `catalog` will no longer be possible. Making this behavior impossible to replicate.

With that said, I also think that what we push to develop should be a working version of the editor. But it seems that `Develop` also has this same behavior. Although this branch does remove the `OSCALLoaderForms` data.

I am fully open to working on fixing this, I just wanted to be sure that it is something worth fixing on this pr.

example/src/App.js
```diff
@@ -89,6 +88,80 @@ function App() {
...
+      loaderElement: OSCALSSPLoader,
+    },
+  ];
+
+  const appBarRoutes = (
+    <Route path="/">
+      <Route index element={`OSCAL Catalog ${appType}`} />
+      {oscalObjectTypes.map((oscalObjectType) => (
+        <Route path={oscalObjectType.pathName} key={oscalObjectType.pathName}>
+          {["", ":id"].map((path) => (
```
kylelaker, line 119: Because we use `["", ":id"]` twice, can we extract it to a variable and then use that in each place?

kylelaker (approved):

PR #485, opened 2022-07-07
Handle duplicate part ids

tuckerzp:
The rev5 catalog contained control parts with duplicated ids. Our
strategy for creating keys for these objects was to just use those id's.
This led to duplicated keys warnings in the Viewer.

To deal with that, it was necessary to use the index of each part within
the list to distinguish between parts with the same id. Because using
the index as a key is not usually prefered, extracting the key to a
variable first was necessary.

### Testing
Here are the files used to replicate the error:

- [Profile](https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/main/profiles/NIST_SP-800-53_rev5_MODERATE-baseline_profile.json)
- [Catalog](https://raw.githubusercontent.com/usnistgov/oscal-content/b97d0c0f682d6686c64077678f5e409427972f23/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_catalog.json)

### DIFF
--- a/src/components/OSCALControlPart.js
+++ b/src/components/OSCALControlPart.js
@@ -81,25 +81,23 @@ export default function OSCALControlPart(props) {
     <OSCALControlPartWrapper ownerState partName={props.part.name}>
       {replacedProse}
       {props.part.parts &&
-        props.part.parts.map((part, index) => {
-          const partKey = `control-part-${part.id}-${index}`;
-          return (
-            <OSCALControlPart
-              part={part}
-              controlId={props.controlId ?? props.control.id}
-              parameters={props.parameters}
-              implementedRequirement={props.implementedRequirement}
-              componentId={props.componentId}
-              modificationAlters={props.modificationAlters}
-              modificationSetParameters={props.modificationSetParameters}
-              key={partKey}
-              isEditable={props.isEditable}
-              onRestSuccess={props.onRestSuccess}
-              onRestError={props.onRestError}
-              partialRestData={props.partialRestData}
-            />
-          );
-        })}
+        props.part.parts.map((part, index) => (
+          <OSCALControlPart
+            part={part}
+            controlId={props.controlId ?? props.control.id}
+            parameters={props.parameters}
+            implementedRequirement={props.implementedRequirement}
+            componentId={props.componentId}
+            modificationAlters={props.modificationAlters}
+            modificationSetParameters={props.modificationSetParameters}
+            // eslint-disable-next-line
+            key={`control-part-${part.id}-${index}`}
+            isEditable={props.isEditable}
+            onRestSuccess={props.onRestSuccess}
+            onRestError={props.onRestError}
+            partialRestData={props.partialRestData}
+          />
+        ))}
     </OSCALControlPartWrapper>
   );
 }
--- a/src/components/OSCALControlPart.js
+++ b/src/components/OSCALControlPart.js
@@ -81,22 +81,25 @@ export default function OSCALControlPart(props) {
     <OSCALControlPartWrapper ownerState partName={props.part.name}>
       {replacedProse}
       {props.part.parts &&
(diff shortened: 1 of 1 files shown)


tuckerzp:
> Is this really the only difference? 

If you look at the [catalog](https://raw.githubusercontent.com/usnistgov/oscal-content/b97d0c0f682d6686c64077678f5e409427972f23/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_catalog.json) and do a `ctrl-f` of `ac-1_obj.a.1"` as an example. I did not find any difference that jumped out at me. If there is one, I can just use that though.

src/components/OSCALControlPart.js
```diff
@@ -81,25 +81,23 @@ export default function OSCALControlPart(props) {
...
+        props.part.parts.map((part, index) => (
+          <OSCALControlPart
+            part={part}
+            controlId={props.controlId ?? props.control.id}
+            parameters={props.parameters}
+            implementedRequirement={props.implementedRequirement}
+            componentId={props.componentId}
+            modificationAlters={props.modificationAlters}
+            modificationSetParameters={props.modificationSetParameters}
+            // eslint-disable-next-line
```
kylelaker, line 93: Explain why though?

kylelaker (changes requested):

kylelaker:
> Looking at https://pages.nist.gov/OSCAL/reference/latest/complete/json-outline/ `parts` are not guaranteed to have `props`, so using `props.value` (the main difference I see) is not a fix.

They're not guaranteed to have an `id` either so using it at all is sorta questionable but I can't imagine a place where one _isn't_ present. But if they are present, per the spec, they should be:
> A human-oriented, locally unique identifier

So I almost feel like this a bug in the catalog. Further, looking at the [source document for the catalog](https://raw.githubusercontent.com/usnistgov/oscal-content/main/src/nist.gov/SP800-53/rev5/xml/NIST_SP-800-53_rev5_catalog.xml), we can see that the ID _is not_ duplicated. The first reference in the source XML is `ac-1_obj.a-1` and the second is `ac-1_obj.a.1`. For some reason, in the resulting JSON (and XML and JSON), they both end up as `ac-1_obj.a.1` and that may be a bug we should report on the [usnistgov/oscal-content](https://github.com/usnistgov/oscal-content) repo? Or perhaps this is an intentional way to represent the data (and we as a viewer should be "smarter" about deduplicating the content in some way)?

I think it's worth having a conversation about.

tuckerzp:
> The first reference in the source XML is `ac-1_obj.a-1` and the second is `ac-1_obj.a.1`.

I had not thought to look at the original `xml` file. That is very odd that when they are converted, they drop the `-` there. I seems like a bug to me given that the `ids` are `"A human-oriented, locally unique identifier"`. It seems they there should be no duplication there.

kylelaker:
Because this doesn't _break_ anything and because we don't (currently) support any kinds of edits on catalogs, I am not especially interested in adding this `index` to the list `key`s. I think this is a creative fix, but I am not sure I want "never display console warnings when provided with non-spec-compliant OSCAL" to become a goal.

If we were crashing in this scenario or if a functionality of the Viewer or Editor was actually broken because of this, I think that I would want to implement a fix; however, the application works just fine other than grumbling a bit.

We'll keep this PR open for a bit and see what comes of the upstream fixes to the content; but I will say that I expect to close this PR without merging at the moment. For that reason, I am going to convert this PR to a Draft to prevent accidentally merging it.

tuckerzp:
@kylelaker https://github.com/usnistgov/oscal-content/issues/117 is now closed, so hopefully this is fixed. We should be good to close this.

kylelaker:
While it has been closed, it seems like the updated content hasn't been published yet; so we'll see what results from that. But you're right. We can close this; it's always possible to re-open it later!

PR #491, opened 2022-07-10
Try to improve OSCALLoaderForm structure

kylelaker:
Currently, the `onUrlChange()` is fired whenever any character is typed.
This is sorta undesirable because it can result in various re-rendering
happening in a few places prematurely.

As-is, this doesn't work. It seems like the actual value received by
`OSCALLoader` "lags" behind the input value by one form submission.

Relies on #490


Bronstrom:
>Relies on https://github.com/EasyDynamics/oscal-react-library/pull/490

With 490 merged, I believe this can be addressed.

I'm glad that were handling this now! I noticed a major slowdown here too when adjusting the form URL, and was thinking it was related to the `onChange` event. Looking over the code limiting the amount of re-rendering and testing the branch, it appears everything works well.

tuckerzp:
It looks like the `oscalUrl` is not actually being changed to the new `value` here? I am unsure why the `setter` of something initialized from `useState` would lag behind one value.

https://github.com/EasyDynamics/oscal-react-library/blob/feature/make-form-behave-predictably/src/components/OSCALLoader.js#L130-#L132

tuckerzp:
I will wait for @kylelaker to review my change and take it out of draft, but I think this should be fixed. There was a weird issue where we were reloading before the state had a change to actually update.

src/components/OSCALLoaderForm.js
```diff
@@ -13,11 +13,18 @@ const OSCALDocumentForm = styled("form")(
 );
 
 export default function OSCALLoaderForm(props) {
+  const url = useRef(null);
```
kylelaker, line 16: I am wondering if we actually need this `useRef`. I added it as a wild guess. But I am curious if we can remove it?

tuckerzp: After playing around with removing it a bit, it seems that `useRef` may be needed. At the very least I am having trouble getting it to work without it.

kylelaker:
Thanks for helping with this @tuckerzp. I approve of the changes you've made here; I think you should feel free to mark this PR as approved as well (if you do approve).

tuckerzp (approved):

mikeisen1 (approved):

Issue #403, opened 2022-05-21
Migrate from `create-react-library` and related tools


kylelaker:
As of April 2022, `create-react-library` and its associated tools are [no longer supported](https://github.com/transitive-bullshit/create-react-library#2022-update). The author of that library created a [blog post comparing related tools](https://transitivebullsh.it/javascript-dev-tools-in-2022#823feddaa1bb4edea19042852b0a5b54). We should look into those tools and pick one that fits our current (and future) needs and investigate performing a migration.

hreineck:
Reading through the linked blog post, https://github.com/nrwl/nx seems like a good option as it has lots of support and documentation on creating React apps, as well as features for multiple projects in one repo ie. our React library + example.
It does a lot of stuff, so maybe it'd be bulky or overkill.
https://github.com/developit/microbundle seems like another viable option.

PR #493, opened 2022-07-11
Make an initial attempt at very basic TypeScript

kylelaker:
This integrates the very beginnings of TypeScript into the codebase.
This is far from good. Nearly everything is currently typed as `any` and
the things that aren't may not have their types specified correctly.

This is a proof of concept and we may not actually use this at any point
in time ever. It just demonstrates that theoretically, incremental
adoption may be an option for us.


kylelaker:
What we're really missing is a library that defines the OSCAL specification as TypeScript types. https://github.com/NREL-CYBER/oscal is super cool and will help with this, but it renames some fields (to use snake_case instead of kebab-case) and requires using some import/export functions defined within the library.

example/package.json
```diff
@@ -15,7 +15,7 @@
     "eject": "node ../node_modules/react-scripts/bin/react-scripts.js eject"
   },
   "dependencies": {
-    "@EasyDynamics/oscal-react-library": "file:..",
+    "@easydynamics/oscal-react-library": "file:..",
```
tuckerzp, line 18: Is there a reason we switched from `EasyDynamics --> easydynamics`? 

I am wondering if that is what is causing the error:
```
Module not found: Error: Can't resolve '@easydynamics/oscal-react-library' in '/home/runner/work/oscal-react-library/oscal-react-library/example/src'
```

kylelaker: That was done as part of publishing to `npm`; we have to be all-lowercase. Make sure to `npm ci` in both directories.

src/components/oscal-utils/OSCALRestUtils.tsx
```diff
@@ -241,12 +259,12 @@ export function createOrUpdateSspControlImplementationImplementedRequirementStat
   if (implementationSetParameters?.length) {
     statementByComponent["set-parameters"] ??= [];
```
tuckerzp, line 260: On lines: `239`, `253`, & `260` there is the error:
```
/OSCAL/oscal-react-library/src/components/oscal-utils/OSCALRestUtils.tsx(239,52): syntax error TS1109: Expression expected.
```
Despite both babel & typescript supporting Logical nullish assignment in their current versions.

kylelaker:
I think a lot of our issues here related back to #403

tuckerzp:
I am attempting to use https://github.com/jaredpalmer/tsdx. I wanted to see how big of an effort it would be to switch to using that.

PR #494, opened 2022-07-11
Move REST Mode Icon

### LINKED ISSUE #467: Subtask: Move `REST Mode` Icon to App Bar

Bronstrom:
With the addition of the Folder-Like REST Object Selector the elements on the page should be repositioned or removed.

The OSCAL REST Mode switch better fits the application being placed in the top menu bar. The switch has also been swapped to the warning orange color to complement the top menu bar's navy background.

Furthermore, this removes the dropdown menu in REST Mode, which is no longer needed for selecting OSCAL model examples with the capabilities of `OSCALDrawerSelector`.

Closes #467


Bronstrom:
`useLayoutEffect()` appears to work better over `useEffect()` in this case as `useLayoutEffect()` runs synchronously after render and before the screen loads. This is much closer to completion, except for the some initial state issues when selecting an OSCAL model in OSCALViewer.

kylelaker:
I like the direction here! What I also really like about this is that it ends up moving all the management of the REST Mode stuff to the app-level (yay!). And that will make it easier to use a [Context](https://reactjs.org/docs/context.html) for this in the future instead of passing the `isRestMode` and the `backendUrl` stuff all the way down.

example/src/App.js
```diff
@@ -137,7 +137,7 @@ function App() {
   );
 
   const oscalObjectLoaderProps = {
-    renderForm: true,
+    renderForm: (isRestMode ? false : true),
```
kylelaker, line 140:
```suggestion
    renderForm: !isRestMode,
```

Bronstrom: Whoops!

Bronstrom:
I removed the loading icon on the initial startup of OSCALEditor. I don't think this is necessary as it seems to falsely give an indication of the screen loading something.

src/components/OSCALLoader.js
```diff
@@ -197,17 +183,28 @@ export default function OSCALLoader(props) {
...
+      setOscalUrl(props.oscalObjectType.defaultUrl);
+      loadOscalData(props.oscalObjectType.defaultUrl);
+      setIsLoaded(false);
+      setIsResolutionComplete(false);
+    }
+  }, [isRestMode]);
 
   let form;
-  if (props.renderForm) {
+  if (props.renderForm && oscalUrl) {
```
kylelaker, line 207: This feels like an odd thing to require in the condition. Even if (for some reason) `oscalUrl` were null/undefined (or otherwise falsey), we probably want the form to still be created.

Imagine a user presses enter on the form while it's empty! Would we just remove the form for that?

src/components/OSCALLoader.js
```diff
@@ -44,7 +44,7 @@ const EditorSplit = styled(Split)`
 export default function OSCALLoader(props) {
   const [isLoaded, setIsLoaded] = useState(false);
   const [isResolutionComplete, setIsResolutionComplete] = useState(false);
-  const { isRestMode, setIsRestMode } = props;
+  const { isRestMode, backendUrl } = props;
```
kylelaker, line 47: Can we just not do this and always refer to them as `props.isRestMode` and `props.backendUrl`? I think when it's like this, it's hard to tell that these get passed in which I think is helpful throughout the function here.

We don't also define `oscalUrl`, `oscalObjectType`, etc... we rely on accessing those via `props`. So let's be consistent.

kylelaker (commented):
I love how much code we're getting rid of here!! A few questions though

kylelaker:
Huh this is certainly an interesting side effect of either this change or one of the earlier changes around the nav drawer 
image

I kinda like this because it means that you keep the exact same document you were working on before. And it works because you're just `GET`ing a document from a URL... I like this but because it is a behavior change (we used to fallback to the default URL for that document type, generally the one from [`EasyDynamics/oscal-demo-content`](https://github.com/EasyDynamics/oscal-demo-content), we should probably discuss this change with the team.

kylelaker:
I also think that in editing mode we probably need a little more spacing between the app bar and the content 
image

kylelaker:
Finally, it looks like toggling REST mode back on still results in an empty page? We might need to dig into that.

tuckerzp:
> toggling REST mode back on still results in an empty page

I was wondering about this too. Would we want it to load back the object we were editing? It does not make a ton of sense to just arbitrarily open the first `Profile` for example when on `/profile`. Plus when viewing something, we do not necessarily have that same file to edit.

src/components/OSCALLoader.js
Bronstrom: Okay, yeah it would probably make sense to make this `props.isRestMode` and `props.backendUrl`. I was just trying to accommodate for the removal of `setIsRestMode`.

Bronstrom:
> I also think that in editing mode we probably need a little more spacing between the app bar and the content

I was working on this yesterday and is completed, but wanted to make some other style changes before I commit.

Bronstrom:
> It looks like toggling REST mode back on still results in an empty page

Yes, this is something I noticed yesterday too. So when switching back to REST mode, and the URL includes the uuid it should load that document?

kylelaker:
> So when switching back to REST mode and the URL includes the uuid, the application should load that document?

For sure! But I also think the UUID should be removed from the URL when *leaving* REST mode.

Bronstrom:
This is finally ready for review! I wanted to make a couple notes.

I found whenever removing the `uuid` in the link, the `uuid` is completely lost. When leaving REST mode after selecting an example, we want to keep the `uuid` but also don't want this to display. Rather than creating a state to store this information and have it passed through underlying components to the Loader, it seemed like a better approach to use `window.history.pushState()`. This allows for a URL to be defined as the new history entry’s URL, and the browser won't load the URL — which allows for `uuid` to be kept when in Viewer mode despite the URL being displayed differently. Please let me know if anyone has a better solution for this.

For now, I moved "Powered by Easy Dynamics" and the GitHub icon center of the menu bar. I think this is decent placement for now. These could also be placed in a Footer, but it seems like this should be addressed separately as this would be a separate new feature.

Lastly, I believe any additional tests for this issue should be added as cypress tests.

Bronstrom:
Here's how the Editor and Viewer looks after these updates:

On initial build & run (in the Editor):
image

After selecting an example Catalog from the Object Selector (in the Editor):
image

Switching to REST mode after selecting a document (in the Viewer):
image

kylelaker:
I _really_ don't like the situation we're ending up in where we have blank pages in some situations. Not sure exactly what I want to see instead. I am going to think on it a bit and maybe we can discuss during parking lot tomorrow.

kylelaker:
Otherwise I absolutely love this!!

Bronstrom:
> I really don't like the situation we're ending up in where we have blank pages in some situations.

Yeah, I'm not really a fan either. Besides the dropdown, we previously had a loading icon constantly on the page until the user selected a component. I personally, didn't think it made much sense because nothing would ever load. I think it would be good to have this as a discussion tomorrow too.

example/src/App.js
```diff
@@ -64,6 +64,19 @@ function App() {
...
+    // Open the drawer when in REST mode and no uuid is present.
+    // Note: The lowest subdirectory of the url is extracted to see if
+    // it contains a uuid.
+    if (
+      isRestMode &&
+      currentUrl.substring(currentUrl.lastIndexOf("/") + 1) === ""
+    ) {
+      setIsDrawerOpen(true);
+    }
+  }, [isRestMode]);
```
Bronstrom, line 78: I want to note this in this PR: this is probably a hacky solution to extract the uuid in `App.js` to open the drawer, but one of the better approaches for now. Another method would be to use `useParams()` in `App.js` to grab the uuid. Although this doesn't seem to work, due to the timing complexity of `OSCALLoader` working with `App.js`. If we end up making changes to how the url displays the uuid, we will probably need to update this.

src/components/OSCALLoader.js
```diff
@@ -197,26 +184,50 @@ export default function OSCALLoader(props) {
...
+    if (oscalObjectUuid) {
+      setIsLoaded(true);
+      setIsResolutionComplete(true);
+      setHasDefaultUrl(true);
+      // Handle uuid displaying in url depending on REST mode
+      window.history.pushState(
+        "",
+        "",
+        `/${props.oscalObjectType.jsonRootName}/${
+          props.isRestMode ? `${oscalObjectUuid}` : ``
```
kylelaker, line 204:
```suggestion
          props.isRestMode ? `${oscalObjectUuid}` : ""
```

kylelaker (commented):
This usage of `window.location` and `window.history` feels really unfortunate. Are we really not able to use hooks for either of these? Or some method of `react-router` to handle the `history` stuff?

Bronstrom:
@kylelaker I'll admit, I'm not a huge fan of `window.location` and `window.history` either and they seem kinda hacky. I was hoping to use hooks for this, such as using `useParams()` to grab the `uuid`, but due to the timing complexities of `OSCALLoader` and `App.js` it seems like the `uuid` is unable to be acquired this way. I didn't thoroughly look into `react-router` — I can look more into this today.

example/src/App.js
```diff
@@ -60,6 +62,20 @@ function App() {
...
```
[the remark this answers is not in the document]
Bronstrom: Received an unnecessary dependency warning when adding it to the list:

image

kylelaker (approved):
For the most part, this seems to work and while some things are inelegant it seems like the best solution available to us for now. Please ensure that tickets to find ways to improve are added to the backlog.

Bronstrom:
> Please ensure that tickets to find ways to improve are added to the backlog.

Sound good to me, will create a ticket for this!

tuckerzp (approved):
This looks fantastic! Thanks for your work on this! I am very glad to have the `Drawer Component` done.

PR #496, opened 2022-07-12
Update or remove TODOs

### LINKED ISSUE #492: Update or remove all `TODO`s in the code base
Throughout the code base, there are many `TODO` comments that reference old Jira tickets (we are no longer using Jira) or that are no longer applicable to the surrounding code.

For every `TODO` in the code base, we need to make sure that a GitHub issue exists for it; then we need to decide whether referencing the `TODO` in the code is appropriate (it very well may not be) and either update or remove it as appropriate.

kperk103:
This updates or removes all TODOs in the code base. If being updated, the corresponding GitHub issue will be linked
with the TODO.

Closes #492.

### DIFF
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -19,6 +19,8 @@ export const OSCALBackMatterCard = styled(Card)(
 `
 );
 
+// TODO: Remove workaround for missing media type
+// https://github.com/EasyDynamics/oscal-react-library/issues/512
 const getURLMediaType = (url) => {
   const lastUrlPath = url.split("//").pop().split("/").pop();
   return lastUrlPath.match(/\.[A-Za-z]{3,4}($|\?)/) || "Unknown";
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -19,8 +19,6 @@ export const OSCALBackMatterCard = styled(Card)(
 `
 );
 
-// TODO: Temporary fix for missing media type (https://github.com/GSA/fedramp-automation/issues/103)
-// Uses file extension instead
 const getURLMediaType = (url) => {
   const lastUrlPath = url.split("//").pop().split("/").pop();
   return lastUrlPath.match(/\.[A-Za-z]{3,4}($|\?)/) || "Unknown";
--- a/src/components/OSCALControlModification.js
+++ b/src/components/OSCALControlModification.js
@@ -83,12 +83,6 @@ function isRelevantId(controlPartId, controlId, element, field) {
 function getModifications(controlPartId, controlId, modList, modText) {
   // Add everything with ids that match controlPartId
 
-  /* TODO: Differences in implementations of the OSCAL standard makes
-  /* this check needed; Certain implementations specify the control ID 
-  /* in the "by-id" field for top-level modifications, 
-  /* others leave "by-id" undefined. 
-  /* See if this can be cleaned up at some point if these differences are resolved.
-  */
   const controlParts = modList.filter((element) =>
     isRelevantId(controlPartId, controlId, element, "by-id")
   );
@@ -138,7 +132,6 @@ export default function OSCALControlModification(props) {
     modLength += len;
   }
 
-  // TODO(EGRC-407): Implement remove modifications
   // if (alter.removes) { }
 
   // Display modifications if there are any
--- a/src/components/OSCALControlModification.js
+++ b/src/components/OSCALControlModification.js
@@ -87,7 +87,6 @@ function getModifications(controlPartId, controlId, modList, modText) {
   // the "by-id" field, others do not, which is what makes this check needed.
   // If these implementation differences are resolved, this should be cleaned up.
   // https://github.com/EasyDynamics/oscal-react-library/issues/498
-
   const controlParts = modList.filter((element) =>
     isRelevantId(controlPartId, controlId, element, "by-id")
   );
@@ -136,10 +135,6 @@ export default function OSCALControlModification(props) {
     );
     modLength += len;
   }
-
-  // TODO: There should also be a way to get all remove modifications so that
-  // users can better understand control modifications with an OSCAL Profile.
-  // https://github.com/EasyDynamics/oscal-react-library/issues/89
 
   // Display modifications if there are any
   if (!modLength) return null;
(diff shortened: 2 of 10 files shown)


src/components/OSCALControlProse.js
```diff
@@ -412,14 +414,14 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
...
-        // TODO - Populate implementationSetParameters with values from statementByComponent or empties
+        // TODO - Populate implementationSetParameters with values from statementByComponent or empties (https://github.com/EasyDynamics/oscal-react-library/issues/503)
         const setParameter = statementByComponent?.["set-parameters"]?.find(
           (element) => element["param-id"] === segment
         ) ?? { "param-id": segment };
         setParameter.values ??= [null];
         const setParameterRef = setParametersRefs[segment];
 
-        // TODO - support for more than 1 item in values arrays
+        // TODO - support for more than 1 item in values arrays (https://github.com/EasyDynamics/oscal-react-library/issues/501)
```
tuckerzp, line 424: For example here the code itself explains that we allow only 1 item and the issue referencing the code can then be used to tell us we need to support multiple values.

tuckerzp (commented):
This is looking great so far! I hope you have been able to learn a bit about the code base by doing this.

As mentioned in the issue, I am personally of the opinion that most `TODO`s created into an issue do not need to be referenced in the code unless keeping it there helps to clarify otherwise confusing code. 

I would just err on the side of avoiding duplication unless needed.

src/components/OSCALControlProse.js
kperk103: That makes sense! After this next commit, all of the `TODO`s will have GitHub issues that are linked to the specific part in the code base so I'll remove the references to the `TODO`s.

Bronstrom:
This is looking great so far! And I really like your choice of imbedding code into the issues, so the developer knows where to look. I have a couple suggestions when your creating your issues.

In general, I would try to be as specific as possible where the issue is when your writing these issues, such as such as component and part. Imbedding the code definitely helps with this, but I think this will make it crystal clear.

- For example in #498: instead of saying "this check" say something like "the `isRelevantId()` check for determining `controlParts` in `OSCALControlModifiication.js`

This might be a little more difficult to address and you might have to ask who left the `TODO`, but generally we also want to note the "why" of the fix when creating an issue. This could be such as the NIST standard has been updated (and link a reference to documentation) or this is a new user feature. Some of these might be okay to skip for now and we come back to address the "why" later.

src/components/OSCALControlModification.js
```diff
@@ -138,7 +132,6 @@ export default function OSCALControlModification(props) {
     modLength += len;
   }
 
-  // TODO(EGRC-407): Implement remove modifications
   // if (alter.removes) { }
```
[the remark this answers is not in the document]
kperk103: Just removed it, thanks!

src/components/OSCALControlModification.js
```diff
@@ -138,8 +137,9 @@ export default function OSCALControlModification(props) {
     modLength += len;
   }
 
-  // TODO(EGRC-407): Implement remove modifications
-  // if (alter.removes) { }
+  // TODO: There should also be a way to get all remove modifications so that
+  // users can better understand control modifications with an OSCAL Profile.
+  // https://github.com/EasyDynamics/oscal-react-library/issues/89
```
kylelaker, line 142: This one can be removed entirely. There's no supporting code. It's just an entirely new feature that is tracked in the referenced issue.

src/components/OSCALBackMatter.js
```diff
@@ -19,8 +19,6 @@ export const OSCALBackMatterCard = styled(Card)(
 `
 );
 
-// TODO: Temporary fix for missing media type (https://github.com/GSA/fedramp-automation/issues/103)
-// Uses file extension instead
```
kylelaker, line 23: While the issue upstream has potentially been fixed, we have not removed our workaround. We still need to do so. We probably want to track this in a separate ticket, unless you want to talk about how to remove the specific workaround here.

kylelaker (changes requested):
@kperk103 Thanks for your work on this! There are a lot of great improvements here. A few things that I think it'd be good to pay attention to:

1. Consistency in spacing. We want the TODO to be "attached" to the code it's documenting, so generally, remove the blank lines between the TODO and the following code.
2. Consistency in formatting. Let's put the URL on its own line at the end of the comment. We don't want comment lines to get too long because they get hard to read.

So great start and I think we're pretty close to the finish line on this one!

kperk103:
> 1. Consistency in spacing. We want the TODO to be "attached" to the code it's documenting, so generally, remove the blank lines between the TODO and the following code.
> 2. Consistency in formatting. Let's put the URL on its own line at the end of the comment. We don't want comment lines to get too long because they get hard to read.

Thank you for the feedback! My most recent commit has fixed these styling inconsistencies and also should have addressed your requested changes.

src/components/OSCALControlProse.js
```diff
@@ -352,11 +352,10 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
...
 
-  // TODO - This approach may not scale well.  We're creating a ref for every
-  // editable field for every control statement/prose for every component before
-  // the user has even asked to edit.  However, other attempts resulted in a
-  // React error of: "Rendered more hooks than during the previous render".
-  // This should be investigated further.
+  // TODO: A React error is thrown unless a ref is created
+  // here. There should be an approach that does not require a ref created
+  // or each of these before editing. https://github.com/EasyDynamics/oscal-react-library/issues/499
+
```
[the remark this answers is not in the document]
kperk103: Yes I can!

src/components/oscal-utils/OSCALControlResolver.js
```diff
@@ -46,7 +46,6 @@ export function getStatementByComponent(
   statementId,
   componentId
 ) {
-  // TODO Remove underscore replacement when OSCAL example content is fixed (https://github.com/usnistgov/oscal-content/issues/58, https://easydynamics.atlassian.net/browse/EGRC-266)
   // Locate matching statement to statementId
   const foundStatement = implReqStatements?.find(
     (statement) =>
```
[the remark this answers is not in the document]
kperk103: Here is the ticket I made for it: https://github.com/EasyDynamics/oscal-react-library/issues/511

src/components/OSCALBackMatter.js
kperk103: I added the `TODO` back into the code base and here is the ticket I made for it: https://github.com/EasyDynamics/oscal-react-library/issues/512

Bronstrom (approved):
This looks good to me @kperk103! Thanks for your work on the `TODO`s and creating several new issues.

kylelaker (approved):
Thanks for your work on this!! I am going to go ahead and merge this. Lots of improvement here.

PR #507, opened 2022-07-15
Open drawer for better navigation

tuckerzp:
This will attempt to open the navigation more proactively. A few of the cases I have found include:

- When a `uuid` is not currently set in `url`
- When we switch out of `REST` mode.

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -23,7 +23,6 @@ import {
   Routes,
   Link as RouterLink,
   useLocation,
-  useParams,
 } from "react-router-dom";
 import Grid from "@mui/material/Grid";
 import Switch from "@mui/material/Switch";
@@ -64,13 +63,19 @@ function App() {
   );
   const [backendUrl] = useState(process.env.REACT_APP_REST_BASE_URL);
   const [hasDefaultUrl, setHasDefaultUrl] = useState(false);
-  const oscalObjectUuid = useParams()?.id ?? "";
 
   useEffect(() => {
-    if (!oscalObjectUuid) {
+    const currentUrl = window.location.href;
+    // Open the drawer when in REST mode and no uuid is present.
+    // Note: The lowest subdirectory of the url is extracted to see if
+    // it contains a uuid.
+    if (
+      isRestMode &&
+      currentUrl.substring(currentUrl.lastIndexOf("/") + 1) === ""
+    ) {
       setIsDrawerOpen(true);
     }
-  }, [oscalObjectUuid]);
+  }, [isRestMode]);
 
   const appType = React.useMemo(
     () => (isRestMode ? "Editor" : "Viewer"),
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -23,6 +23,7 @@ import {
   Routes,
   Link as RouterLink,
   useLocation,
+  useParams,
 } from "react-router-dom";
 import Grid from "@mui/material/Grid";
 import Switch from "@mui/material/Switch";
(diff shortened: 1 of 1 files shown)


src/components/OSCALDrawerSelector.js
```diff
@@ -95,6 +96,13 @@ function createTree(backendUrl, handleClose) {
 
 export default function OSCALDrawerSelector(props) {
   const theme = useTheme();
+  const oscalObjectUuid = useParams()?.id ?? "";
+
+  useEffect(() => {
+    if (!oscalObjectUuid) {
+      props.handleOpen();
+    }
+  }, [oscalObjectUuid]);
```
tuckerzp, line 105: This has the weird consequence of opening the `drawer` upon any load. I think it is because `uuid` is not guaranteed to be set in time of the drawer being created?

example/src/App.js
```diff
@@ -72,7 +72,10 @@ function App() {
   }, [appType]);
 
   const handleAppNavOpen = (event) => {
-    setAnchorEl(event.currentTarget);
+    if (event) {
```
tuckerzp, line 75: We do not have a `onClick` `event` here to open the `nav`. I would not really consider having a `uuid` in the `url` an `event`, so I added this check.

tuckerzp (commented):
I still need to address the fact that switching out of `REST` mode does not open nav. One fix would be to just remove the `uuid` from the `url` upon clicking the button off. But, as discussed a bit below, that may not be the best approach.

src/components/OSCALDrawerSelector.js
tuckerzp: Upon looking at this more, it seems that `useParams().id` is not defined. Which is odd because I can get the `Loader` to correctly print the `uuid`.

tuckerzp:
A quick update in case someone wants to work off of me here.

---

The `useEffect(()=>{}, [uuid])` only ever seems to be updated once. `uuid` from `useParams()` starts out as `undefined` and remains that way. I have a strong feeling if we can figure out why its not defined, the `useEffect` part should work. 

---

The functionality from going to the `Editor -> Viwer` & vice versa is wonky and should be discussed. I think that currently in `develop` going to the `Viewer` after editing a file, will just load the default OSCAL content page. The current version of this branch will load a `Viewer` perspective of the file you were just editing. I actually think that the latter makes a lot more sense. Regardless going back to the `Editor` after the `Editor -> Viewer` switch is broken and needs to be fixed. There are two options I can think of:
1. `Loading Default OSCAL content page is correct`: When switching back to the editor, we should be greeted by a black page with the drawer being set open. This also means we need to drop the `id` parameter of the `url` when we go from `Editor -> Viewer`. I think fixing the `useEffect` issue and dropping the `id` parameter will fix everything here. 
2. `Loading the edited OSCAL content page is correct`: When going back to the editor, it really should just load that page again. A really jank way of doing this would to force a page reload somehow since we would never need to remove `uuid` from the `url`.

src/components/OSCALDrawerSelector.js
Bronstrom: There is some issue with the complexity of `OSCALLoader` and it's communication to `App.js`, where `App.js` is unable to grab `useParams()`. You're right about the `uuid` not being set in time. Callback functions like `useEffect()` can be used to capture states after render, although `useParams()` is a hook itself and cannot be used within another hook.

Bronstrom:
`useParams()` does not seem to work, due to the timing complexity of `OSCALLoader` working with `App.js`. This is probably a more hacky solution, but it seemed like taking the end of the url, which at this point always ends blank or contains the uuid after the last `/`, is the best way to extract the uuid and determine whether the drawer should be open or closed. If we end up making changes to how the url functions, we will probably need to update this.

PR #514, opened 2022-07-20
Add custom favicon

### LINKED ISSUE #304: Add a custom favicon to the OSCAL Viewer sandbox app
As a user of the OSCAL Web App, I want to a unique favicon so that I can distinguish tabs and bookmarks in my browser.

Today, the [Sandbox viewer](https://oscal-viewer.msd.easydynamics.com) uses the default React [`favicon.ico` file](https://github.com/EasyDynamics/oscal-react-library/blob/develop/example/public/favicon.ico). We should consider adding a unique graphic/icon to use as a shortcut icon.

### Acceptance Criteria

 - `favicon.ico` has been replaced with a custom graphic _and/or_ we have added a `<link rel="shortcut icon">` pointing to a custom graphic
 - Navigating to the [Sandbox viewer](https://oscal-viewer.msd.easydynamics.com) (in a browser that supports favicons) shows a favicon other than the standard React app icon

### Implementation Notes

 - We may want a different favicon for _our_ deployment vs the one in the repo, this would need to be handled as part of the CI pipeline (this is especially true if we elect to use the Easy Dynamics logo or favicon from the public site)
 - It may be preferable to consider graphics more holistically, both for EasyGRC itself and for the public materials

kperk103:
This will update the favicon from the default React icon to Easy Dynamic's logo.

Closes #304.


kylelaker:
Hmm... we might want to think about this for a moment. We claim in `LICENSE` that this repo and its contents are licensed under the MIT license. And that's not quite true of our corporate logo.

I wonder if we should inject this file during the CI/CD deployment process instead of just adding it directly to the repo?

kperk103:
> Hmm... we might want to think about this for a moment. We claim in `LICENSE` that this repo and its contents are licensed under the MIT license. And that's not quite true of our corporate logo.
> 
> I wonder if we should inject this file during the CI/CD deployment process instead of just adding it directly to the repo?

That makes sense. Is there anywhere I can look to get a better understand of the CI/CD pipeline and how I would go about this?

kylelaker (approved):
I'm going to retract my previous concern. We already have a logo in here and I think that this is the easiest solution available. We have other planned work to further reconsider how we handle branding assets and I think we can resolve my other concerns then.

Bronstrom (approved):

tuckerzp (approved):

PR #519, opened 2022-07-25
Create Footer

Bronstrom:
Here's my idea for a footer within our application.

This moves all info promoting Easy Dynamics ("Powered by Easy Dynamics" and the GitHub icon) to the footer. This first rendition of the footer is a "sticky footer", meaning it sits at the bottom of the viewport when there is no content and when there is content (an example is selected) the footer will display beneath it. While Easy Dynamics is not necessarily  "front and center" like it was at the top `AppBar`, it still will display right away when the application is launched at the bottom of the screen.

Furthermore, I thought it made sense to move `appBarRoutes` to display in the center of the AppBar. This helps avoid having `Object Selector` cover up the title. Although, not sure how I feel about it being centered here. Let me know if anyone has opinions on this. We could honestly just leave it where it was before and have the title also display in `Object Selector` when opened.

Also I want to note, this branch is created off of feature/move-rest-mode-icon as this branch seems most relevant with the placement of the REST mode icon in `AppBar`.

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -256,60 +256,25 @@ function App() {
             <AppBar position="static">
               <Toolbar>
                 <Grid container alignItems="center">
-                  <Grid item md={4} align="left">
-                    <Grid container alignItems="center">
-                      <Grid item align="left">
-                        <OpenNavButton
-                          edge="start"
-                          onClick={handleAppNavOpen}
-                          color="inherit"
-                          aria-label="menu"
-                          size="large"
-                        >
-                          <MenuIcon />
-                        </OpenNavButton>
-                        {navigation}
-                      </Grid>
-                      <Grid item align="left">
-                        <Typography variant="h6">
-                          <Routes>{appBarRoutes}</Routes>
-                        </Typography>
-                      </Grid>
-                    </Grid>
+                  <Grid item md={3} align="left">
+                    <OpenNavButton
+                      edge="start"
+                      onClick={handleAppNavOpen}
+                      color="inherit"
+                      aria-label="menu"
+                      size="large"
+                    >
+                      <MenuIcon />
+                    </OpenNavButton>
+                    {navigation}
                   </Grid>
-                  <Grid item md={4}>
-                    <Grid container alignItems="center" justifyContent="center">
-                      <Grid item align="center">
-                        <Typography
-                          variant="body2"
-                          sx={{ color: "white", fontStyle: "italic" }}
-                        >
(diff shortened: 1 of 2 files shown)


kylelaker:
I think this is a good start but I think it's a little bit of an issue that the footer (and therefore the GitHub link and EDC branding) aren't on screen 99% of the time. What if the footer were "sticky" and always at the bottom?

Bronstrom:
> What if the footer were "sticky" and always at the bottom?

I was wondering if the branding not appearing a lot of the time would be an issue. I'll work on putting this together.

mikeisen1:
What would be the reasoning behind moving the corporate branding from the app bar and to the footer? I understand if we want to keep on the app bar and add it to the footer but don't see why we move it from the header. If anything, I like the idea of keeping the app bar "sticky" so we can also have the navigation/object selector accessible at anytime.

Bronstrom:
> What would be the reasoning behind moving the corporate branding from the app bar and to the footer?

At the time of moving the REST Mode icon, I was thinking the `AppBar` was getting a bit cluttered. Also, I generally think of company branding and relevant links to appear in the footer of webpages, while the application title and its navigation features appear in the `AppBar`/top menu bar.

>  If anything, I like the idea of keeping the app bar "sticky" so we can also have the navigation/object selector accessible at anytime.

I like this idea a lot. Also, if we decide to go this route of having a footer, I think it would make sense to complement a sticky footer with a sticky `AppBar`.

Bronstrom:
I want to lay out my opinions about the choice of having a footer.

First of all, when I think of a footer, it contains relevant links to legal information, details about the company, contact info, and social media apps. To me this seems like the right place for the GitHub icon to go as it's social media-like (kinda). I also think "Powered by Easy Dynamics" being fitting here to as it describes who is behind the application and it links us to our website for additional information. I believe the menu bar should lay out information pertaining to the application, such as containing the title and functional elements. 

In general, content in both the menu bar and the footer are considered important information to the user. Having these both placed in a fixed position allows for the user to switch to a different OSCAL example or check out Easy Dynamics anywhere they are on the page. 

I think this separation between application specific parts and company information will be valuable as a whole.

Bronstrom:
image

kylelaker:
@Bronstrom Thank you for this but it has gone awhile without updates; for now, I think we have other things that cover updating some UI/UX designs. We will be able to refer to this pull request later if needed but I think keeping it open results in unnecessary noise on the PR list.

PR #525, opened 2022-07-27
Handle Blank "Set-Parameters" Object

### LINKED ISSUE #422: Handle Blank "Set-Parameters" Object
As someone using the OSCAL Viewer/Editor on an SSP that has many components, control descriptions, and control parameters, I want control parameters to be accurately displayed in a readable UI.

In an SSP, a control or control part can have a description as well as any number of fillable parameters. Both of these are currently editable in the OSCAL Editor. However, in a case where parameters are available to be filled, but ONLY a description is entered by the user, OSCAL Viewer will render a "blank" parameter:
image
Ideally, if the parameters are not edited, the default text should be displayed as normal:
image

### Implementation Notes

In the [OSCAL SSP Schema](https://pages.nist.gov/OSCAL/reference/latest/system-security-plan/json-reference/#/system-security-plan/control-implementation/implemented-requirements/statements), a given control or control part can have a `by-components` array that specifies a `description` field and a `set-parameters` object pertaining to a given component. Saving a description in the Editor and not filling the parameters results in the creation of an object in `by-components` that has a `description` field but not a `set-parameters` object. The OSCAL Viewer/Editor interprets this with the behavior described above.

kperk103:
Currently, if a control or control part is given a `description` but
no `set-parameters` object, the placeholder is replaced with a blank
parameter. This will now allow a `description` to be given without a
`set-parameters` object and the placeholder will not be replaced.

Closes #422.


src/components/OSCALControlProse.js
```diff
@@ -267,22 +268,45 @@ function getParameterValueSegment(
...
   return (
     <SegmentTooltipWrapper
-      constraintsDisplay={constraintsDisplay}
+      constraintDisplay={constraintsDisplay}
       key={`segment-wrapper-key-${key}`}
     >
-      <ParamValue component="span" key={`param-value-key-${key}`}>
-        {parameterValue}
-      </ParamValue>
+      <ParamLabel component="span" key={`param-label-key-${key}`}>
```
tuckerzp, line 307: These `ParamLabels` seem very similar. Could they be extracted to a function?

Something like
```
const paramLabel = (text, key) => { return <ParamLabel> }
```

Same thing for the `ParamValue`, its very little duplication but it would not take much to fix.

kperk103: That makes sense. I'll try doing that

kperk103: @tuckerzp My most recent commit has addressed this, let me know if it is what you were thinking!

tuckerzp: That is what I meant! Looking at the file as a whole though, we could make it even more general to the file itself. For example, all `<ParamLabel>` tags are using a `parameterLabelText` & a `key`. Would it be much more effort to generalize both the functions you created to reduce duplication in the entire file? You may want to rename the function so it is not almost exactly similar to the type declaration at the top of the file though (`paramLabel` vs `ParamLabel`) .

kperk103: I can do that!

src/components/OSCALControlProse.test.js
```diff
@@ -59,24 +59,64 @@ function proseParamValuesEditingRenderer() {
...
+  const changeDescription = () =>
+    fireEvent.change(descriptionTextField, {
+      target: { value: "Test Description" },
+    });
+  const saveChanges = () =>
+    screen
+      .getByRole("button", {
+        name: "save-control-1_smt.a",
+      })
+      .click();
```
kylelaker, line 88: Please see https://github.com/EasyDynamics/oscal-react-library/pull/529#discussion_r934536462 which I apparently submitted just a moment too late.

kylelaker (changes requested):

src/components/OSCALControlProse.test.js
kperk103: That makes sense! My most recent commit has reverted the tests back to their previous commit.

kylelaker: Thanks! Really appreciate your flexibility and @tuckerzp and I will sync a bit on testing practices to get a little more consensus here.

src/components/OSCALControlProse.js
```diff
@@ -267,22 +278,34 @@ function getParameterValueSegment(
...
     );
   }
-
+  if (!constraintsDisplay.length) {
+    return styledParamLabel(`param-label-key-${key}`, parameterLabelText);
+  }
   return (
     <SegmentTooltipWrapper
-      constraintsDisplay={constraintsDisplay}
+      constraintDisplay={constraintsDisplay}
```
mikeisen1, line 305: Maybe I'm missing something but why is the prop here named `constraintDisplay` and not `constraintsDisplay` as in other places?

kperk103: I must have mistyped! It should be `constraintsDisplay` and not `constraintDisplay`. My most recent commit now reflects this change.

src/components/OSCALControlProse.js
```diff
@@ -255,7 +265,8 @@ function getParameterValueSegment(
   implReqSetParameters,
   parameterId,
   modificationSetParameters,
-  key
+  key,
+  parameters
```
mikeisen1, line 269: When creating a new function or adding parameters to an existing one, we need to update the function documentation as well (the `@param` stuff).

tuckerzp (approved):
Thanks for your work on this!

kylelaker:
@kperk103 Can you include screenshots of you app similar to the ones in #422 to compare the behavior on this branch?

kperk103:
> @kperk103 Can you include screenshots of you app similar to the ones in #422 to compare the behavior on this branch?

Screenshot (5)

Here you can see that in line `a`, only a description has been given, and the placeholder value is still there, which is the desired functionality.

kylelaker (approved):

PR #529, opened 2022-07-29
Test that blank `set-parameters` is handled correctly

schen2166:
This is testing the changes made by https://github.com/EasyDynamics/oscal-react-library/pull/525. Specifically, it adds tests to check that if the description is changed, the placeholder values for the blank parameters are displayed, and that if the description and parameters are changed, the new values will be displayed instead of the placeholders.

### DIFF
--- a/src/components/OSCALControlProse.test.js
+++ b/src/components/OSCALControlProse.test.js
@@ -59,7 +59,7 @@ function proseParamValuesEditingRenderer() {
 }
 
 function testOSCALControlProseEditing(parentElementName, renderer) {
-  test(`${parentElementName} displays description and parameter inputs`, async () => {
+  test(`${parentElementName} displays default description and parameter inputs`, async () => {
     renderer();
     screen
       .getByRole("button", {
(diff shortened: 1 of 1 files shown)


tuckerzp (changes requested):
A lot of these tests have repeated functionality that is very similar for each test. Is there any way to extract that duplication to a function? I think it would also make the tests easier to follow.

tuckerzp (changes requested):

kylelaker:
What about Control Prose is being tested here? Can you be more descriptive in the commit message or the main PR body about what exactly the goal is here? I appreciate any work to improve test coverage, but being more descriptive will make it easier to understand scope and intent when reviewing.

tuckerzp:
> What about Control Prose is being tested here?

I think these are tests for #525 which is addressing #422. But I agree an update to the title or description would make review easier.

kylelaker:
> But I agree an update to the title or description would make review easier

Yes. Perhaps "Test that blank `set-parameters` is handled correctly"

kperk103:
> What about Control Prose is being tested here? Can you be more descriptive in the commit message or the main PR body about what exactly the goal is here? 

Just changed the pr title and the description so that it is more clear what we are testing. Thanks for the feedback!

tuckerzp (approved):
Thanks for your work on this @schen2166 @kperk103! We can have a more through review on #525.

src/components/OSCALControlProse.test.js
```diff
@@ -68,9 +68,26 @@ function testOSCALControlProseEditing(parentElementName, renderer) {
...
+  const changeDescription = () =>
+    fireEvent.change(descriptionTextField, {
+      target: { value: "Test Description" },
+    });
+  const saveChanges = () =>
+    screen
+      .getByRole("button", {
+        name: "save-control-1_smt.a",
+      })
+      .click();
```
kylelaker, line 88: This is a lot of logic that isn't within a `test` explicitly. I know that @tuckerzp asked to reduce duplication but I actually don't think that this is especially correct.

Sometimes in tests, it's actually a bit better to have some duplication where and when it's expected, especially in setup logic (I actually don't even like `beforeEach` but it's already here so we've got it).

I'm sorry to ask you to go back and forth but could we go back to the previous iteration where this was included within the `test` functions? Perhaps some of the values (`"control-1_prm_1"`, etc) are _candidates_ for extracting to their own variables. But I am not even convinced that's necessary.

The reason for this is that when our tests break, we want to have as little confusion as possible about where things went wrong. Having very explicit (and even repetitive) setup logic is helpful for understanding that. Reducing the amount of different functions or block levels that I have to look at to see what at what line exactly the failure is occurring is extremely helpful. As-is, I think that reducing the duplication unnecessarily increases the cognitive complexity to the point that I don't really even understand what's happening anymore.

kylelaker (changes requested):

PR #538, opened 2022-08-04
Display control info for all components

tuckerzp:
The `component-uuid` was not being passed down to all controls.

closes #421 as well as https://github.com/EasyDynamics/oscal-rest-service/issues/100

PR #540, opened 2022-08-05
Add unit tests for Component UUID edit on save

Bronstrom:
This adds unit tests for [PR #538](https://github.com/EasyDynamics/oscal-react-library/pull/538). Test data was created to test for statements including decimal statements.

### DIFF
--- a/src/components/OSCALControlProse.test.js
+++ b/src/components/OSCALControlProse.test.js
@@ -4,15 +4,22 @@ import {
   OSCALReplacedProseWithParameterLabel,
   OSCALReplacedProseWithByComponentParameterValue,
 } from "./OSCALControlProse";
-import { exampleModificationSetParameters } from "../test-data/ModificationsData";
+import {
+  exampleModificationSetParameters,
+  exampleModificationSetParametersDecimal,
+} from "../test-data/ModificationsData";
 import {
   controlImplTestData,
+  controlImplWithDecSmtTestData,
   controlProseTestData,
+  controlProseDecimalTestData,
   exampleParams,
+  exampleDecimalParams,
 } from "../test-data/ControlsData";
 import { sspRestData } from "../test-data/SystemData";
 
 const labelTestData = "label-1";
+const labelTestDataDecimal = "label-1.1";
 
 function proseParamLabelsRenderer() {
   render(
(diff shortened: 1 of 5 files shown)


Bronstrom:
I'm not a huge fan of how much test data I had to add to test the statements containing decimals, but it seems like this is the best option with how our testing is setup. I tried to lessen duplicate code tests in `OSCALControlProse` by refactoring it a bit.

src/components/OSCALControlProse.test.js
```diff
@@ -59,66 +132,75 @@ function proseParamValuesEditingRenderer() {
...
-        name: "save-control-1_smt.a",
-      })
-      .click();
-    expect(descriptionTextField.value).toBe("Test Description");
-
-    screen.getByLabelText("control-1_prm_1");
-    expect(paramValueTextField.value).toBe(
-      "control 1 / component 1 / parameter 1 value"
-    );
+    renderOSCALControlProseEditing(renderer, rollName);
```
tuckerzp, line 147: So I do appreciate removing duplication in code, but think that it can be helpful with tests. Please see https://github.com/EasyDynamics/oscal-react-library/pull/525#discussion_r934537769. This duplication was very intentional.

I think in general taking all the code out into various functions makes the tests harder to read and understand.

src/components/OSCALControlProse.test.js
```diff
@@ -40,6 +47,72 @@ function testOSCALControlProseParamLabels(parentElementName, renderer) {
...
+      componentName.slice(1).toLowerCase()
+    } description of implementing control 1`
+  );
+
+  expect(paramValueTextField.value).toBe(
+    `${controlName} / ${componentName} / parameter 1 value`
+  );
+}
+
+function EditDescription(
```
tuckerzp, line 74: Going with my other comment, is this really just editing the description? It seems to not only edit the description, but verify that the change is made. That is not exactly obvious when seeing the function used in the tests.

I think either allowing the duplication or being more descriptive of what this function does would be helpful.

tuckerzp (changes requested):

tuckerzp (approved):

kylelaker (approved):

Issue #156, opened 2021-08-18
Handle Markdown in Prose and Other Fields


hreineck:
As a user involved in the governance, risk, and compliance of systems and components exploring controls, I want the prose of control statements, guidance, and other components that may use the [markup type](https://pages.nist.gov/OSCAL/reference/datatypes/#markup-multiline) to be in a neat and human-readable format.

## Acceptance Criteria
- Instances of markdown appearing in areas specified below are handled in a reasonably clean and readable way:
  - Emphasized text is bolded or italicized, or the * characters are simply removed.
  - Hyperlinks are implemented, or the parentheses/brackets characters are removed
  - If any other instances of markdown are identified eg. quoted, subscripted, or superscripted text they are handled in a clean and readable way.
- The following fields should support markdown:
  - Prose statements
  - Metadata Title
  - System Implementation remarks
  - Control Implementation descriptions (SSP and Component Definition)
- A (lightweight) architectural decision record (see [examples](https://github.com/EasyDynamics/easygrc-infrastructure/tree/main/doc/adr)) has been created that describes our choice of Markdown renderer and specifically talks to the security aspects of that library 

## Implementation Notes

Some examples of markdown appearing as plain text in the [800-53 rev5 catalog:](https://raw.githubusercontent.com/usnistgov/oscal-content/master/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_catalog.json)

![image](https://user-images.githubusercontent.com/43356777/129962705-85fd8f8d-5c67-4f5f-971b-5e788d176b15.png)

![image](https://user-images.githubusercontent.com/43356777/129962790-7cd27126-695a-47c7-b91a-5bde348002b9.png)

Markdown also appears in the prose of several of the guidance buttons:

![image](https://user-images.githubusercontent.com/43356777/129962950-40c71ee2-7341-4cf8-b963-161e3a57a5c5.png)

kylelaker:
Markdown + the `url` parameter needs to be careful to not accidentally open up an XSS vulnerability.

kylelaker:
https://marked.js.org/ or something like it could be an option. But we'd definitely want to sanitize and probably find a way to strip any `<script>` or maybe even `<style>` tags in a sanitization process (since many Markdown parsers support raw HTML). A nice thing about Markdown is that it's fairly readable as-is without being rendered. We may want to consider using workers to asynchronously perform the rendering as well and we can cancel them if they work for too long.

[...]

PR #542, opened 2022-08-05
Handle markdown in control and systems implementations

kperk103:
This will allow control implementation descriptions and
system implementation remarks to handle markdown so that
the text appears in a clean and readable format.

Part of #156.


example/package.json
```diff
@@ -44,5 +44,11 @@
   ],
   "files": [
     "build"
-  ]
+  ],
+  "jest": {
```
tuckerzp, line 48: I think this change may already be present in `develop`

Bronstrom: Is this dependency needed in `example`? From what I understood from the issue, we're altering `src/component`s with markdown. Are we altering anything with markdown in `example`?

kperk103: We aren't altering anything with markdown in `example`, however in one of my previous commits (d17d0a4), the following error is displayed when trying to build and test:
Screenshot (10)
This is the same error I was getting when trying to run the component tests locally before adding the dependency in `package.json`. Adding the same dependency in `example/package.json` fixed the error as well. You are correct that nothing is being altered with markdown in `example`, so I am confused as to why this is necessary.

Bronstrom (approved):
Good work - looks good to me!

src/components/OSCALControlProse.js
```diff
@@ -384,6 +385,9 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
   // https://github.com/EasyDynamics/oscal-react-library/issues/499
   const statementByComponentDescription =
     statementByComponent?.description || null;
+  const statementByComponentDescriptionMarkdown = (
```
tuckerzp, line 388: To stay consistent with discussions from #548 on markdown vs markup, can we change this to:
```suggestion
  const statementByComponentDescriptionMarkup = (
```

kperk103: That makes sense! My most recent commit has this change

kperk103:
@tuckerzp Here are some examples of markdown being handled in control implementation descriptions and system implementation remarks:
Screenshot (13)
Screenshot (15)

kylelaker:
> @tuckerzp Here are some examples of markdown being handled in control implementation descriptions and system implementation remarks: Screenshot (13) Screenshot (15)

Wait no way! This is so cool! It works in the tooltips too?! I literally love this.

kylelaker (changes requested):
Still need to request changes though; sorry. Looks like maybe putting the Prose inside a `<Typography>` causes new console warnings. @tuckerzp can you support that?

tuckerzp:
> new console warnings

After all my fuss about warnings, I am embarrassed to have missed that! I will assist in fixing them. 

https://github.com/EasyDynamics/oscal-react-library/actions/runs/2825466728

kylelaker:
> Looks like maybe putting the Prose inside a `<Typography>` causes new console warnings

I think if we need to (if we can't just always tell it to use `Typography`) we can pass something (a `Typography`) for `paragraph` or whatever as a prop to the underlying component, instead of wrapping things at a higher level.

src/components/OSCALSystemImplementation.js
```diff
@@ -44,7 +45,9 @@ export default function OSCALSystemImplementation(props) {
               <OSCALSectionHeader>System Implementation</OSCALSectionHeader>
             </Grid>
             <Grid item xs={12}>
-              <Typography>{props.systemImplementation.remarks}</Typography>
+              <Typography>
+                <OSCALMarkdownProse text={props.systemImplementation.remarks} />
+              </Typography>
```
tuckerzp, line 50: Is this `Typography` even needed?

kperk103:
> Still need to request changes though; sorry. Looks like maybe putting the Prose inside a `<Typography>` causes new console warnings. @tuckerzp can you support that?

I believe that if I use the `OSCALMarkupLine` component from https://github.com/EasyDynamics/oscal-react-library/pull/548/files#diff-09fcaba6157b2ce435ef376c32a2e3b1e705969d23bbc7ebeebf6288c2738786 it gets rid of these console warnings. I can wait until that PR is merged to make these changes if you believe that is the best way to get rid of these warnings.

src/components/OSCALSystemImplementation.js
kperk103: Without the `Typography`, the font of the system implementation remarks changes so it is not consistent with the rest of the page. The first screenshot shows it with the `Typography` and the second screenshot shows it without the `Typography`:
Screenshot (18)
Screenshot (16)

tuckerzp:
> I believe that if I use the `OSCALMarkupLine`...  it gets rid of these console warnings

I agree that would get rid of the console warnings, but am not completely sure if we want these things to be a single line. I would argue that the `description` would make sense to be a `OSCALMarkUpMultiLine`. 

> we can pass something (a `Typography`) for paragraph or whatever as a prop to the underlying component, instead of wrapping things at a higher level.

@kylelaker I may misunderstand what you mean by this, but would doing something similar to [here](https://github.com/EasyDynamics/oscal-react-library/pull/543#discussion_r940488448) be a potential way forward?

I am just not entirely convinced that everywhere we get this warning switching to `OSCALMarkupLine` is the best solution and think our answer may need to be more robust.

kylelaker:
> but would doing something similar to [here](https://github.com/EasyDynamics/oscal-react-library/pull/543#discussion_r940488448) be a potential way forward

I _think_ that's what I'm saying, yeah.

kperk103:
> I agree that would get rid of the console warnings, but am not completely sure if we want these things to be a single line. I would argue that the `description` would make sense to be a `OSCALMarkUpMultiLine`.

If I have a multiline `description` in the control implementation and use `OSCALMarkupLine`, it still works:

Screenshot (19)

The same also works for a multiline `remark` in system implementations.

If you still think it would be best for me to do something similar to the link you provided, I can look into that as well.

tuckerzp:
> If I have a multiline description in the control implementation and use `OSCALMarkupLine`, it still works:

@schen2166 & @kylelaker I am curious on your thoughts about this. It seems that maybe the `OSCALMarkUpMultiLine` is not needed for this current issue? I am unsure if we need to necessarily have both of them now when viewing at least only requires what we have defined as `OSCALMarkupLine`. I have not seen a place where `OSCALMarkUpMultiLine` is strictly needed over `OSCALMarkupLine`.

schen2166:
>  I am curious on your thoughts about this. It seems that maybe the OSCALMarkUpMultiLine is not needed for this current issue?
For the scope of this issue on resolving markdown in prose, I don't see an immediate need to support multiline markdown however it may be something that should be supported in the future. in my opinion, getting `MarkupMultiline` working doesn't need to be a priority. `MarkupLine` should have be able to resolve the issues related to this task

kperk103:
> `MarkupLine` should have be able to resolve the issues related to this task

I agree that this can be used to resolve the issue while still allowing functionality for multiple lines. If this is the solution we choose, I think I will need to wait until https://github.com/EasyDynamics/oscal-react-library/pull/548/files#diff-09fcaba6157b2ce435ef376c32a2e3b1e705969d23bbc7ebeebf6288c2738786 is merged in order to use this component.

kylelaker:
> It seems that maybe the `OSCALMarkUpMultiLine` is not needed for this current issue? I am unsure if we need to necessarily have both of them now when viewing at least only requires what we have defined as `OSCALMarkupLine`. I have not seen a place where `OSCALMarkUpMultiLine` is strictly needed over `OSCALMarkupLine`.

Which one does the spec say that this is?

kperk103:
> Which one does the spec say that this is?

If I'm understanding it correctly, I think the spec says it is a markup-multiline

src/components/OSCALMarkupProse.js
```diff
@@ -8,6 +8,17 @@ import ReactMarkdown from "react-markdown";
  */
 
 export function OSCALMarkupMultiLine(props) {
+  if (props.paragraphComponent) {
+    return (
+      <ReactMarkdown
+        components={{
+          p: props.paragraphComponent,
```
tuckerzp, line 15: I am worried what this does if `props.paragraphComponent` is `undefined`. I am unsure of the functionality of rewriting components if given a `undefined` value. We may want to put some kind of check. 
```suggestion
          p: props.paragraphComponent ? props.paragraphComponent :  p ,
```


I also think this might cause the `StyledToolTip` to have issues.

kylelaker: `p: props.paragraphComponent ?? 'p'`

kylelaker: Iff we need a check in the first place

kperk103: Thanks! My most recent commits address this

src/components/OSCALControlProse.js
```diff
@@ -384,6 +385,12 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
   // https://github.com/EasyDynamics/oscal-react-library/issues/499
   const statementByComponentDescription =
     statementByComponent?.description || null;
+  const statementByComponentDescriptionMarkup = (
+    <OSCALMarkupMultiLine
+      text={statementByComponentDescription}
+      paragraphComponent="p"
+    />
+  );
```
kylelaker, line 393: Should we do this in the case where `statementByComponentDescription` is `null`?

```tsx
const statementByComponentDescriptionMarkup = statementByComponentDescription ? (
  <OSCALMarkupMultiLine
    text={statementByComponentDescription}
    paragraphComponent="p"
  />
);
```

kylelaker (changes requested):

src/components/OSCALControlProse.js
kperk103: That's a great idea, thank you!

src/components/OSCALControlProse.js
```diff
@@ -384,6 +385,13 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
   // https://github.com/EasyDynamics/oscal-react-library/issues/499
   const statementByComponentDescription =
     statementByComponent?.description || null;
+  const statementByComponentDescriptionMarkup =
+    statementByComponentDescription ? (
+      <OSCALMarkupMultiLine
+        text={statementByComponentDescription}
+        paragraphComponent="p"
```
tuckerzp, line 392: I thinking mapping `p -> p` is a bit weird here. This goes back to my original comment suggesting the checking for `paragraphComponent` to be `null`. I don't think we should force a rewrite of the `p` tag.

tuckerzp (approved):

kylelaker (approved):
Thank you for all your hard work on this and for going back and forth on so many items. Great stuff!

PR #546, opened 2022-08-08
Make fields with prose statements support markdown

schen2166:
There are statements in the controls descriptions and guidance areas where markdown is not supported which can make it hard to read. Convert prose statements in the guidance and descriptions of controls into markdown.

PR #548, opened 2022-08-08
Make fields with prose statements support markdown 

schen2166:
There are statements in the controls descriptions and guidance areas where markdown is not supported which can make it hard to read. Convert prose statements in the guidance and descriptions of controls into markdown.
This is came from issue https://github.com/EasyDynamics/oscal-react-library/pull/546


schen2166:
Following shows that the formatting is changed to have a new line before and after all of the orange placeholder texts. Content remains the same however the markdown in prose has been addressed

src/components/OSCALControlProse.js
```diff
@@ -168,7 +169,8 @@ function getTextSegment(text, key) {
   }
   return (
     <Typography component="span" key={key}>
-      {text}
+      {/* {text} */}
+      <OSCALMarkdownProse text={text} />
```
kylelaker, line 173: Why not just replace the `Typography` with an `OSCALMarkdownProse`?

schen2166: The code above resulted in the output on the right
I also tried this one

(The output for the second one is the same as the one above)

Would the next approach be to look at `react-markdown` and adjust the settings with the new line or is there another way to change the `getTextSegment()` function

schen2166: Wrapping an `OSCALMarkdownProse` component also gives the same output. I believe it is because `react-markdown` adds a new line at the end of their component so there will be a new line at the end of each span which knocks down text to the next line.

kylelaker: See https://github.com/remarkjs/react-markdown/issues/42

src/components/OSCALMarkdownProse.js
```diff
@@ -8,5 +8,13 @@ import ReactMarkdown from "react-markdown";
...
-  return <ReactMarkdown>{props.text}</ReactMarkdown>;
+  return (
+    <ReactMarkdown
+      components={{
+        p: React.Fragment,
+      }}
+    >
+      {props.text}
+    </ReactMarkdown>
+  );
```
kylelaker, line 19: We can't do this all the time. Sometimes we need the `<p>` tags. I think we need an `OSCALMarkupLine` and `OSCALMarkupMultiline` component that handle sensible defaults and `component` limitations for each. Coordinate with @tuckerzp on how to name the compnents.

schen2166: Making separate components will work. Another approach could be to write functions in `OSCALMarkdownProse` which handle either single or multi-line markdown conversion. The default could be converting it into `<p>` tags and another function can handle conversions to React Fragments.

What is your opinion on this?

tuckerzp: I'd think the current `OSCALMarkupProse` would need minor changes to act as `OSCALMarkupMultiline`. Then you would also implement `OSCALMarkupLine`. I think that they could reside in the same file though.

src/components/OSCALControlProse.js
```diff
@@ -162,15 +163,12 @@ function getImplReqSetParameters(implReqStatements, componentId) {
...
   if (!text) {
     return null;
   }
-  return (
-    <Typography component="span" key={key}>
-      {text}
-    </Typography>
-  );
+
+  return <OSCALMarkupSingleLine text={text} />;
```
kperk103, line 171: I'm not sure how much of a concern this is at the moment, but by removing `Typography` and replacing it with `OSCALMarkupSingleLine`, it changes the text font of lines that have been given a description.
Screenshot (9)
Screenshot (10)
The first screenshot is how it currently appears in `develop`, and the fonts are all the same. The second screenshot is how it appears when `Typography` is replaced with `OSCALMarkupSingleLine`, and the font for the lines that have been given descriptions is different from the rest.

kperk103:
```suggestion
  return (
    <Typography component="span" key={key}>
      <OSCALMarkupSingleLine text={text} />
    </Typography>
  );
```

I believe this will fix the issue

schen2166: Thank you, I reinstated the Typography tags around the Markup component

src/components/OSCALControlProse.js
```diff
@@ -432,7 +430,7 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
     .map((segment, index) => {
       if (index % 2 === 0) {
         // This is not a parameter placeholder
-        return getTextSegment(segment, index.toString());
+        return getTextSegment(segment);
       }
       if (isEditingStatement) {
         // We're currently editing this statement, so build param input
```
schen2166, line 436: I am wondering what the map in this RegExp is trying to accomplish. My interpretation of it is that it renders a textfield component where the parser finds the regular expression (`{{ insert: param, ([0-9a-zA-B-_.]*) }}`) 

Currently there is an issue that I found when the regular expression is embedded within placeholder text. See image below

src/components/OSCALComponentDefinitionControlImplementation.test.js
```diff
@@ -43,10 +43,14 @@ test("OSCALComponentDefinitionControlImplementation displays component parameter
...
-    "Does something with < control 1 / parameter 1 label > and < control 1 / parameter 2 label >"
+  const resultByProseBeginning = getByTextIncludingChildern(
+    "Does something with"
   );
-  expect(resultByProse).toBeVisible();
+  const param1 = within(resultByProseBeginning).getByText("< control 1 / parameter 1 label >");
+  const param2 = within(resultByProseBeginning).getByText("< control 1 / parameter 2 label >");
+  expect(resultByProseBeginning).toBeVisible();
+  expect(param1).toBeVisible();
+  expect(param2).toBeVisible();
```
Bronstrom, line 53: This is a solution for resolving the tests currently. With the way we're implementing our markup, we will need to grab the parameter values from within our main results body.

src/components/OSCALControlProse.js
```diff
@@ -335,7 +337,7 @@ export function OSCALReplacedProseWithParameterLabel(props) {
             index
           );
         })
-    : props.prose;
+    : getTextSegment(props.prose, 0);
```
schen2166, line 340: Note that this is the case where `props.prose` does not contain any placeholder text. `getTextSegment` requires two parameters, the `text` and the `key/index` (index is utilized when inthe array from the map function) 
I put zero as the second parameter because there is only one index in the array if there were no placeholder text in the array.

Is there a better value for the second parameter?

schen2166:
Here is the summary of the big fixes from this issue in the Editor



**Above**, Guidance button linking, **Middle**: Guidance descriptions without placeholder text **Bottom**: Bold and Italics

src/components/OSCALComponentDefinitionControlImplementation.test.js
schen2166: I implemented the changes using regular expressions to check if they are in the document for this test

src/components/OSCALMarkdownProse.js
schen2166: `OSCALMarkupLine` has been implemented and merged in

src/components/OSCALControlProse.js
Bronstrom: Ideally if we're handling keys, we want a value here that isn't static like '0', or we will receive warning's in the console saying  "Each child in a list should have a unique 'key' prop." We commonly use `uuid`s as keys, although `prose` doesn't doesn't have a `uuid` associated with it. Maybe we could use some other prop that is passed in with `prose` and unique.

Bronstrom (changes requested):
Looking good so far @schen2166. I have a couple suggestions.

Also looks like there is a merge conflict with an import. We'll need to handle that as well.

src/components/OSCALControlImplementationImplReq.test.js
```diff
@@ -40,10 +40,19 @@ export default function testOSCALControlImplementationImplReq(
...
-      "Does something with control 1 / component 1 / parameter 1 value and control 1 / component 1 / parameter 2 value"
+    const nonplaceholder1 = getByTextIncludingChildern(/does something with/i);
+    const placeholderText1 = getByTextIncludingChildern(
+      /control 1 \/ component 1 \/ parameter 1 value/i
     );
-    expect(result).toBeVisible();
+    const nonplaceholder2 = getByTextIncludingChildern(/and/i);
+    const placeholderText2 = getByTextIncludingChildern(
+      /control 1 \/ component 1 \/ parameter 2 value/i
+    );
```
Bronstrom, line 50: I see you're handling these parameters with regex. For consistency purposes among our other tests, could these be handled as strings?

schen2166: We have to options to make the tests consistent which do you think is better?
1. Convert the Regexp to strings
2. Convert the Strings to Regexp

Bronstrom: Convert the regex to strings in the two methods handled in this PR. It would be a lot less work.

schen2166: `DefinitionControlImplementation` test was able to convert to a string 
`ControlImplementationImplReq` doesn't work with strings. Also the linter will complain with the `\` which escape the `/`
I wouddn't mind converting the strings to RegExps for consistency purposes.

Bronstrom: Interesting, I would just say to leave that single one as a regex for now. Don't worry about converting the other string parameters to Regex.

tuckerzp:
To fix some issues with history, this is being redone in #559

PR #550, opened 2022-08-09
Add markdown support in metadata titles

kperk103:
This will add markdown support for metadata titles. 

Part of #156.


kperk103:
This adds markdown support for metadata titles on each page, along with the titles in the drawer component as shown below: 
Screenshot (21)

kylelaker (approved):
Good stuff! This emphasizes for me that we need to do a better job of styling the drawer but I do like this a lot! Thanks!

PR #555, opened 2022-08-10
Remove unneeded prop passing in loader

tuckerzp:
Testing the loader turned out to be a bit trickier than anticipated. For now removing some prop passing is a good enough improvement.

### DIFF
--- a/src/components/OSCALLoader.js
+++ b/src/components/OSCALLoader.js
@@ -42,6 +42,16 @@ const EditorSplit = styled(Split)`
   }
 `;
 
+/**
+ * Returns url parameter provided by the browser url, if it exists. If the url
+ * parameter exists, we want to override the default viewer url.
+ *
+ * @returns The url parameter of the browser url, or null if it doesn't exist
+ */
+export function getRequestedUrl() {
+  return new URLSearchParams(window.location.search).get("url");
+}
+
 export default function OSCALLoader(props) {
   const [isLoaded, setIsLoaded] = useState(false);
   const [isResolutionComplete, setIsResolutionComplete] = useState(false);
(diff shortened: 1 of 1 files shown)


src/components/OSCALLoader.js
```diff
@@ -43,6 +43,9 @@ const EditorSplit = styled(Split)`
 `;
 
 export default function OSCALLoader(props) {
+  // THIS SHOULD BE PRINTING
+  console.log(props.renderForm, hasDefaultUrl);
+
```
tuckerzp, line 48: As the comment suggests, this really should be logging something. I am not sure if it has something to do with render being asynchronous, not being able to create just a `OSCALLoader`, or something else entirely.

tuckerzp: This was solved by using `render`

src/components/OSCALLoader.test.js
```diff
@@ -0,0 +1,39 @@
...
+        oscalUrl={sspUrl}
+      />
+    );
+
+    // WHEN
+    render(loader);
+    expect(loader.props.oscalUrl).toEqual(sspUrl);
+
+    handleUrlChange.mockImplementation((u) => [u, changeURL]);
+    loader.props.oscalUrl = newSspUrl;
```
tuckerzp, line 34: I would like to be able to somehow mock the oscalUrl state change. From what I have seen, this is done a lot by simulating a click. The loader is not a button so that is made a bit difficult.

tuckerzp:
I think that using just a `Loader` might not be the right approach. At the very least using something like an `SSPLoader` gives you more ability to actually load something.

tuckerzp:
I think to have the `Reload` button actually reload, it needs to not be disabled. To do that `onResolutionComplete` needs to be true. To make it true, we somehow need to do `Loader.onResolutionComplete()`. Which kinda goes back to the original issue of not being able to use a component's methods.

https://github.com/EasyDynamics/oscal-react-library/blob/887b6637c3a9e8bfa9648000f111dfe88447449a/src/components/OSCALLoader.js#L182-L184

kylelaker (approved):

Bronstrom (approved):

PR #562, opened 2022-08-12
Fix OSCALMarkupLine so that metadata title shows

kperk103:
Originally, OSCALMarkupLine was handling the text as props, and
in a recent PR it was changed to handle it as children. As a result,
the metadata title no longer showed up. This PR makes changes in how
the metadata title text is being passed to OSCALMarkupLine so that
it works according to the new implementation of the component.

### DIFF
--- a/src/components/OSCALEditableTextField.js
+++ b/src/components/OSCALEditableTextField.js
@@ -61,7 +61,7 @@ function textFieldWithEditableActions(
   return (
     <>
       <Typography display="inline" variant={props.typographyVariant}>
-        <OSCALMarkupLine text={props.value} />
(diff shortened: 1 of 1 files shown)


kylelaker (approved):
Thanks for catching this!

tuckerzp (approved):

tuckerzp:
> May be should prioritize https://github.com/EasyDynamics/oscal-react-library/pull/493 or something like it

I agree this change in particular would have been fixed a lot quicker if we had been using `TypeScript`. A few prs were slowed down purely because the error plain `JavaScript` says was not very clear.

PR #571, opened 2022-08-16
Remove profile resolution relative URL workarounds

### LINKED ISSUE #505: Remove profile resolution relative URL workarounds
We include a few profile link resolution solutions in `OSCALProfileResolver` using `itemUrl`:

https://github.com/EasyDynamics/oscal-react-library/blob/87fe8416a683d4a2f4edabc68fcf4a8454b21ddd/src/components/oscal-utils/OSCALProfileResolver.js#L45-L47

as well as in `OSCALSspResolver` using `profileUrl`:

https://github.com/EasyDynamics/oscal-react-library/blob/87fe8416a683d4a2f4edabc68fcf4a8454b21ddd/src/components/oscal-utils/OSCALSspResolver.js#L29-L31

Both of these can be removed if the links are resolved in the residing `oscal-content`.

tuckerzp:
This cleans up how we handle the creation of URLs. In general, we now make
more effort to cleanup URLs to be absolute and resolved (removing `..` etc)
from the URLs. This results in a more consistent behavior across the
application.

Primarily, this is initiated by usnistgov/oscal-content#60 being resolved;
we had done additional workarounds there to handle incorrect structure in
relative `href`s. Removing the workaround highlighted the possibility for
additional optimizations and improvements.

Closes #505
Closes EasyDynamics/oscal-editor-deployment#128


src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -21,14 +13,8 @@ export default function resolveLinkHref(
...
-  return fixJsonUrls(
-    getAbsoluteUrl(
-      getUriFromBackMatterByHref(backMatter, href, mediaTypeRegex).href,
-      parentUrl
-    )
-  );
+  const fooBarBaz = !href.startsWith("#")
+    ? href
+    : getUriFromBackMatterByHref(backMatter, href, mediaTypeRegex).href;
+  return getAbsoluteUrl(fooBarBaz, parentUrl);
```
kylelaker, line 19: lol we should probz rename this

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -1,18 +1,10 @@
...
-  // Replacing all instances of xml with json in the path *should* get us the correct json URL
-  return absoluteUrl.replace(/xml/g, "json");
-}
-
 export function getAbsoluteUrl(href, parentUrl) {
-  return href.startsWith("http") ? href : `${parentUrl}/../${href}`;
+  if (href.startsWith("https://") || href.startsWith("http://") || !parentUrl) {
+    return href;
+  }
+  return new URL(href, parentUrl).toString();
```
hreineck, line 7: Seems URL() fails when `parentUrl` is a relative link (eg. when the object in backMatter points to a relative link as its `rlink` parameter)
For example, when I run a Profile that contains a relative link, the console shows the error: 
`TypeError: Failed to construct 'URL': Invalid base URL
(OSCALLinkUtils.js:7:10)`
[Reference docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) 

We probably need another conditional here to check whether `parentUrl` is a relative link, and if so resolve the link correctly.

This overlaps with [github.com/EasyDynamics/oscal-editor-deployment/issues/141](https://github.com/EasyDynamics/oscal-editor-deployment/issues/141)

src/components/oscal-utils/OSCALProfileResolver.js
```diff
@@ -39,15 +39,8 @@ export default function OSCALResolveProfileOrCatalogUrlControls(
   onError,
   pendingProcesses // tracks state of recursive calls
 ) {
-  let itemUrl = origItemUrl;
+  const itemUrl = new URL(origItemUrl, parentUrl).toString();
```
hreineck, line 42: See above comment on `OSCALLinkUtils.js` -- this probably needs a check for a `TypeError` when `parentUrl` is not an absolute link.

src/components/oscal-utils/OSCALLinkUtils.js
hreineck: Actually, thinking about this a little more, we could require that `parentUrl` is an absolute link as it is created. This variable originates from the environment variable `REST_APP_BASE_URL` that is set by the user.
An important case when this environment variable must be a relative link is in the OSCAL Editor Docker Deployment; the relative link is used so that the Docker container operates agnostic of the port the user chooses.

Seems we either need a check here in `OSCALLinkUtils.js` to fix the URL in the case of a relative link,
or to add a check elsewhere in the code as `REST_APP_BASE_URL` is initially resolved to grab the user's specified port and build an absolute URL.
We could also rethink the use of the environment variable at build time for this purpose if it's too clunky, but if I recall correctly we didn't find a nice way of specifying the base REST URL in the container any other way.

Would like to see everyone's thoughts on a good approach discussed here: https://github.com/EasyDynamics/oscal-editor-deployment/issues/141

example/src/App.js
```diff
@@ -52,6 +52,10 @@ const LogoImage = styled("img")`
   margin-right: 1em;
 `;
 
+function getBackEndUrl(urlString) {
+  return (!urlString) ? "" : new URL(urlString, window.location.href).toString();
+}
```
kylelaker, line 57: I get that ternaries are good down in the like actual JSX bits but here, I think it's too clever. Expanding this lets us add comments for each bit, which I think makes this non-obvious logic more clear.

```suggestion
function getBackEndUrl(urlString) {
  // If given something falsey, we need to also return something falsey.
  // Then ensures that we don't accidentally for the application to always be
  // in editor/REST-mode.
  if (!urlString) {
    return null;
  }
  // Because the URL may be a relative path and various application logic
  // assumes that it will always be an absolute URL, it is easiest to just
  // handle converting the potentially relative URL to an absolute URL and
  // explicitly use the current origin as the base.
  try {
    return new URL(urlString, window.location.href).toString();
  } catch (err) {
    // If the given URL is invalid, for some reason, fallback to just using
    // the provided value. This may result in more predictable errors
    // elsewhere.
    console.log(`Unable to create an absolute URL for ${urlString}`);
    return urlString;
  }
}
```

example/src/App.js
```diff
@@ -61,7 +82,8 @@ function App() {
     // be different).
     !!process.env.REACT_APP_REST_BASE_URL
   );
-  const [backendUrl] = useState(process.env.REACT_APP_REST_BASE_URL);
+  const [backendUrl] = useState(getBackEndUrl(
+    process.env.REACT_APP_REST_BASE_URL));
```
kylelaker, line 86:
```suggestion
  const [backendUrl] = useState(
    getBackEndUrl(process.env.REACT_APP_REST_BASE_URL)
  );
```

kylelaker:
@tuckerzp Can you take a look at these new merge conflicts?

src/components/OSCALBackMatter.js
```diff
@@ -113,18 +118,15 @@ export default function OSCALBackMatter(props) {
...
-                    }
+                    icon={<OpenInNewIcon />}
                     key={rlink.href}
                     label={getMediaType(rlink)}
                     component="a"
                     role="button"
-                    href={getAbsoluteUrl(rlink.href, props.parentUrl)}
+                    href={getAbsoluteUrl(
+                      rlink.href,
+                      props.parentUrl.startsWith(".") ? null : props.parentUrl
```
kylelaker, line 128: Do we still need this?

src/components/OSCALBackMatter.js
```diff
@@ -91,7 +91,12 @@ export default function OSCALBackMatter(props) {
 
   const getMediaType = (rlink) =>
     rlink["media-type"] ||
-    guessExtensionFromHref(getAbsoluteUrl(rlink.href, props.parentUrl));
+    guessExtensionFromHref(
+      getAbsoluteUrl(
+        rlink.href,
+        props.parentUrl.startsWith(".") ? null : props.parentUrl
```
kylelaker, line 97: This too

src/components/OSCALDiagram.js
```diff
@@ -14,7 +14,7 @@ export default function OSCALDiagram(props) {
     diagramUri = resolveLinkHref(
       props.backMatter,
       link.href,
-      props.parentUrl,
+      props.parentUrl.startsWith(".") ? null : props.parentUrl,
```
kylelaker, line 17: Oof I really don't know if I like this... I want to find something better than `startsWith(".")` if we do

src/components/OSCALBackMatter.js
tuckerzp: Without using the absolute url `guessExtensionFromHref` fails to parse the type of `diagram.png`. `diagram.png` is not an absolute url, so the type is always `Unknown`.

https://github.com/EasyDynamics/oscal-react-library/blob/5b4a79269232343deb0c53ce7498d450063b0fae/src/components/oscal-utils/OSCALLinkUtils.js#L57-L70

src/components/OSCALBackMatter.js
tuckerzp: Are you suggesting to remove the `getAbsoluteUrl` call here? I do not think that `rlink.href` is an absolute url so I feel like we may still need to keep the function call.

src/components/OSCALDiagram.js
tuckerzp: Since we rely on the error checking with `URL()`, this is needed at least for our tests. Otherwise we pass the parent url as `./`, which makes an invalid url. I also think removing it may lead to some wonkiness with the actual `Viewer` but am not sure if we have any sample data to be sure.

example/src/App.js
```diff
@@ -52,6 +52,27 @@ const LogoImage = styled("img")`
   margin-right: 1em;
 `;
 
+function getBackEndUrl(urlString) {
+  // If given something falsey, we need to also return something falsey.
+  // Then ensures that we don't accidentally for the application to always be
+  // in editor/REST-mode.
```
hreineck, line 58: May be a grammar issue/typo here? @kylelaker

hreineck (changes requested):
I did some local testing and it looks like these changes fix our relative links troubles. Awesome work!
I only found one nitpick; and I would like to see Kyle's comments addressed and then this has my approval.

example/src/App.js
kylelaker:
```suggestion
  // This ensures that we don't accidentally force the application
  // to always be in editor/REST mode by creating a URL.
```

Oh wow what made me write that horror in the first place

kylelaker (approved):
Nice fixes! That seems pretty creative and I like it. Thank you!

Bronstrom (approved):
Good work @tuckerzp! Looks good to me.

PR #574, opened 2022-08-17
Improve media type detection in back matter resource links

### LINKED ISSUE #512: Remove missing media type workaround
Once https://github.com/GSA/fedramp-automation/issues/103 has been closed, the workaround for missing media types can be removed in `OSCALBackMatter.js`. Currently, file extension is being used.
https://github.com/EasyDynamics/oscal-react-library/blob/b34991180dc344729d8de8952222de6b072cef79/src/components/OSCALBackMatter.js#L24-L27

Bronstrom:
Since the `media-type` property is optional, we still need a clean and testable way to give users some indication of the type of file they are attempting to access. While the preferred method remains to display the `media-type` attribute of the `rlink`, we fall back to using any file type-looking string at the end of the URL. Failing that, we fall back to displaying “Unknown”. In the future, we may consider using alternative means to determine the media type.

This resolves the TODO comment and converts a regular expression to more imperative code that hopefully will be easier to maintain and to extend in the future.

### Testing

Ensure https://raw.githubusercontent.com/GSA/fedramp-automation/master/dist/content/baselines/rev4/json/FedRAMP_rev4_MODERATE-baseline_profile.json successfully loads and displays `OSCALBackMatter` links as simply file extensions or a valid rlink `media-type`.

Closes #512

### DIFF
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -9,7 +9,7 @@ import FormatQuoteIcon from "@mui/icons-material/FormatQuote";
 import DescriptionIcon from "@mui/icons-material/Description";
 import OpenInNewIcon from "@mui/icons-material/OpenInNew";
 import StyledTooltip from "./OSCALStyledTooltip";
-import { getAbsoluteUrl } from "./oscal-utils/OSCALLinkUtils";
+import { getAbsoluteUrl, getURLMediaType } from "./oscal-utils/OSCALLinkUtils";
 import { OSCALSection, OSCALSectionHeader } from "../styles/CommonPageStyles";
 import { OSCALMarkupLine, OSCALMarkupMultiLine } from "./OSCALMarkupProse";
 
@@ -20,11 +20,6 @@ export const OSCALBackMatterCard = styled(Card)(
     flex-direction: column;
 `
 );
-
-const getURLMediaType = (url) => {
-  const fileExtension = new URL(url).pathname.split(".").reverse()[0];
-  return fileExtension.length <= 4 ? fileExtension.toUpperCase() : "Unknown";
-};
 
 function TitleDisplay(props) {
   const title = props.resource.title || "No Title";
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -9,7 +9,10 @@ import FormatQuoteIcon from "@mui/icons-material/FormatQuote";
 import DescriptionIcon from "@mui/icons-material/Description";
 import OpenInNewIcon from "@mui/icons-material/OpenInNew";
 import StyledTooltip from "./OSCALStyledTooltip";
-import { getAbsoluteUrl, getURLMediaType } from "./oscal-utils/OSCALLinkUtils";
+import {
+  getAbsoluteUrl,
+  guessExtensionFromHref,
+} from "./oscal-utils/OSCALLinkUtils";
 import { OSCALSection, OSCALSectionHeader } from "../styles/CommonPageStyles";
 import { OSCALMarkupLine, OSCALMarkupMultiLine } from "./OSCALMarkupProse";
 
@@ -88,7 +91,7 @@ export default function OSCALBackMatter(props) {
 
   const getMediaType = (rlink) =>
     rlink["media-type"] ||
-    getURLMediaType(getAbsoluteUrl(rlink.href, props.parentUrl));
+    guessExtensionFromHref(getAbsoluteUrl(rlink.href, props.parentUrl));
 
   const backMatterDisplay = (resource) => (
     <Grid item xs={3} key={resource.uuid}>
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -21,6 +21,11 @@ export const OSCALBackMatterCard = styled(Card)(
 `
 );
 
+const getURLMediaType = (url) => {
+  const fileExtension = new URL(url).pathname.split(".").reverse()[0];
+  return fileExtension.length <= 4 ? fileExtension.toUpperCase() : "Unknown";
+};
+
 function TitleDisplay(props) {
   const title = props.resource.title || "No Title";
   const color = props.resource.title ? "initial" : "error";
@@ -87,7 +92,8 @@ export default function OSCALBackMatter(props) {
   }
 
   const getMediaType = (rlink) =>
-    rlink["media-type"] || getAbsoluteUrl(rlink.href, props.parentUrl);
+    rlink["media-type"] ||
+    getURLMediaType(getAbsoluteUrl(rlink.href, props.parentUrl));
 
   const backMatterDisplay = (resource) => (
     <Grid item xs={3} key={resource.uuid}>
--- a/src/components/OSCALBackMatter.js
+++ b/src/components/OSCALBackMatter.js
@@ -21,13 +21,6 @@ export const OSCALBackMatterCard = styled(Card)(
 `
 );
 
-// TODO: Remove workaround for missing media type
-// https://github.com/EasyDynamics/oscal-react-library/issues/512
-const getURLMediaType = (url) => {
-  const lastUrlPath = url.split("//").pop().split("/").pop();
-  return lastUrlPath.match(/\.[A-Za-z]{3,4}($|\?)/) || "Unknown";
-};
-
 function TitleDisplay(props) {
   const title = props.resource.title || "No Title";
   const color = props.resource.title ? "initial" : "error";
@@ -94,8 +87,7 @@ export default function OSCALBackMatter(props) {
   }
 
   const getMediaType = (rlink) =>
-    rlink["media-type"] ||
-    getURLMediaType(getAbsoluteUrl(rlink.href, props.parentUrl));
+    rlink["media-type"] || getAbsoluteUrl(rlink.href, props.parentUrl);
 
   const backMatterDisplay = (resource) => (
     <Grid item xs={3} key={resource.uuid}>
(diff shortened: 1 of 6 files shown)


src/components/OSCALBackMatter.js
```diff
@@ -21,11 +21,9 @@ export const OSCALBackMatterCard = styled(Card)(
...
 );
 
-// TODO: Remove workaround for missing media type
-// https://github.com/EasyDynamics/oscal-react-library/issues/512
 const getURLMediaType = (url) => {
-  const lastUrlPath = url.split("//").pop().split("/").pop();
-  return lastUrlPath.match(/\.[A-Za-z]{3,4}($|\?)/) || "Unknown";
+  const fileExtension = new URL(url).pathname.split(".").reverse()[0];
+  return fileExtension.length <= 4 ? fileExtension.toUpperCase() : "Unknown";
 };
```
kylelaker, line 27: Can we move this to (maybe) Link Utils, export it, and write tests?

src/components/OSCALBackMatter.js
```diff
@@ -21,11 +21,9 @@ export const OSCALBackMatterCard = styled(Card)(
 `
 );
 
-// TODO: Remove workaround for missing media type
-// https://github.com/EasyDynamics/oscal-react-library/issues/512
 const getURLMediaType = (url) => {
-  const lastUrlPath = url.split("//").pop().split("/").pop();
-  return lastUrlPath.match(/\.[A-Za-z]{3,4}($|\?)/) || "Unknown";
+  const fileExtension = new URL(url).pathname.split(".").reverse()[0];
```
kylelaker, line 25:
```suggestion
  const fileExtension = new URL(url).pathname.split(".").pop();
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop

or

```suggestion
  const fileExtension = new URL(url).pathname.split(".").at(-1);
```

kylelaker:
The `package-lock.json` files should not be modified. Please check merges/rebases to see what's up

Bronstrom:
> The package-lock.json files should not be modified. Please check merges/rebases to see what's up

Issues should be resolved now.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -32,3 +32,8 @@ export default function resolveLinkHref(
     )
   );
 }
+
+export function getURLMediaType(url) {
+  const fileExtension = new URL(url).pathname.split(".").pop();
+  return fileExtension.length <= 4 ? fileExtension.toUpperCase() : "Unknown";
+}
```
kylelaker, line 39: I want to see a lot of tests for this. Things like:

```
getURLMediaType("https://easydynamics.com/favicon.ico");
getURLMediaType("https://github.com/EasyDynamics/oscal-react-library/pull/574/files");
getURLMediaType("https://raw.githubusercontent.com/EasyDynamics/oscal-react-library/develop/README.md");
getURLMediaType("https://csrc.nist.gov/projects/united-states-government-configuration-baseline");
getURLMediaType("https://doi.org/10.6028/NIST.SP.800-97");
```

And more. I want to see lots of tests to make sure we understand the expected behavior of this function.

kylelaker: For clarity:

```ts
expect(getURLMediaType("https://easydynamics.com/favicon.ico")).toBe("ICO");
```

or whatever

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -32,3 +32,16 @@ export default function resolveLinkHref(
...
   );
 }
+
+export function getURLMediaType(url) {
+  let fileExtension = "";
+  try {
+    fileExtension = new URL(url).pathname.split(".").pop();
+    if (fileExtension === "/") {
+      fileExtension = "href";
+    }
```
Bronstrom, line 42: An href (like "https://github.com/") returns the media type "/". I think it would make more sense for this to be "HREF", so I added a conditional check here.

kylelaker: I would strongly prefer that this ends up as “Unknown”

kylelaker: Also, `"Unknown"` is becoming a [magic string](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants). We should create a variable for this.

```
const fallbackFileExtension = "Unknown";
```

which we should use just about everywhere within this function.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -32,3 +32,16 @@ export default function resolveLinkHref(
...
+  } catch {
+    fileExtension = "Unknown";
+  }
+  return fileExtension.length <= 4 ? fileExtension.toUpperCase() : "Unknown";
+}
```
kylelaker, line 47:
```suggestion
// One could imagine that this is instead an inner function of
// `guessExtensionFromHref`. For style reasons that may or may not make sense
// but I've separated it here for ease of writing and clarity.
const getFileExtension(url) {
  try {
    // TODO: Consider new URL(url).pathname.split("/").pop().split(".").pop()
    return new URL(url).pathname.split(".").pop().trim();
  } catch {
    // console.warn("Unable to detect extension for: " + url);
    return "";
  }
}

export function guessExtensionFromHref(url) {
  const fallbackFileExtension = "Unknown";
  // TODO: Use an HTTP HEAD request to attempt to get a `Content-Type` header
  // for the underlying document. The request(s) should likely properly handle
  // redirects in order to handle DOI URLs etc. This would require refactoring
  // the length check in the return as well. In fact, doing so may replace the
  // need for an extension-based check at all. But in implementing that we
  // should remember that a HEAD is more expensive than splitting strings.
  const detectedExension = getFileExtension(url);
  const extLength = detectedExtension.length;
  // This is a guess based on typical file types. If the length is less than 2
  // or more than 4, it's probably not actually a file extension but instead we've
  // grabbed some unexpected part of the filename.
  if (extLength > 4 || extLength < 2) {
    return fallbackFileExtension;
  }
  return detectedExtension;
}
```

src/components/oscal-utils/OSCALLinkUtils.test.js
```diff
@@ -0,0 +1,67 @@
...
+  it("gets Easy Dynamics SSP JSON Example", async () => {
+    expect(
+      getURLMediaType(
+        "https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/main/system-security-plans/ssp-example.json"
+      )
+    ).toEqual("JSON");
+  });
+
+  it("gets href", async () => {
+    expect(getURLMediaType("https://github.com/")).toEqual("HREF");
```
kylelaker, line 61:
```suggestion
    expect(getURLMediaType("https://github.com/")).toEqual("HREF");
```

`HREF` says nothing about a file type.

Bronstrom: Will check for "Unknown" here instead.

src/components/oscal-utils/OSCALLinkUtils.js
Bronstrom: I'm liking the suggestions here.

After adding the `console.warn` statement, we will see this output whenever running  `OSCALLinkUtils.test.js`in the console - I think this will be okay. It seems like `pathname.split("/").pop().split(".").pop()` proves to be better than `pathname.split(".").pop().trim()` with the case of something like. The former will return "Unknown" (which is valid), while the latter returns the output "PNG/".

I will look a little more into using an HTTP HEAD request to attempt to get a `Content-Type` header.

kylelaker: I think `HEAD` requests are separate ticket (that you or I should write)

Bronstrom: It seems like the local linter doesn't like the `console` statement, but the linter for this repo accepts it?

Bronstrom: > I think `HEAD` requests are separate ticket (that you or I should write)

Sounds good, I don't mind writing this at some point today.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -32,3 +32,44 @@ export default function resolveLinkHref(
...
+/**
+ * Guesses an extension type based on a provided href, and provides
+ * "Unkown" when an invalid.
+ *
+ * @param url A given href.
+ * @returns A string providing a valid or "Unknown" extension type.
+ */
+export function guessExtensionFromHref(url) {
+  const fallbackFileExtension = "Unknown";
+  // TODO: Use an HTTP HEAD request to attempt to get a `Content-Type` header
```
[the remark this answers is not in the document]
Bronstrom: Yes, this is going to be created today and the ticket was be linked.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -43,7 +43,6 @@ function getFileExtension(url) {
   try {
     return new URL(url).pathname.split("/").pop().split(".").pop();
   } catch {
-    console.warn(`Unable to detect extension for: ${url}`);
```
Bronstrom: I think I would rather not include this in the code base, especially with `console` statements causing linter warning. Is this necessary for the user to see?

tuckerzp: I agree. I do not see much benefit to the user from leaving console warnings in the codebase.

src/components/oscal-utils/OSCALLinkUtils.js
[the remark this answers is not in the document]
tuckerzp: https://github.com/EasyDynamics/oscal-react-library/issues/580

tuckerzp (approved):
I think this is a lot cleaner than using the previous `regex`. Thanks for your work on this @Bronstrom!

kylelaker (approved):

PR #576, opened 2022-08-18
Display Parties Contact Information

### LINKED ISSUE #405: Display Parties Contact Information
### Acceptance Criteria 
- [x] Discussion of design is documented on this task
- [x] Parties Information is displayed

tuckerzp:
This adds a `Contact` dialog box to metadata parties. When clicked, it opens a dialog displaying all the parties contact information. This also rewrites the tests a bit to no longer have parent components of `OSCALMetadata` test functionality. 

### Testing file

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/f03073fddf33b0d65d3d0e84a7ad76c74e7be6c9/system-security-plans/ssp-example.json gives a good example of party contact info to view

### Reviewing / Things to look for
This pr seems a bit more complex than it actually is upon further investigation. Here is a list of what changes have been made for reviewing. 
- `OSCALMetadata` has new sub components being used to display part contact information.
- The tests have been expanded upon, they are more imperative and verbose now. 
- The storybook has additional components. This is fairly repetitive. Running `npm run build-storybook && npm run storybook` will let you view it locally. 
- New sample data has been added to be used in the tests / storybook.

closes #405

### DIFF
--- a/src/components/OSCALMetadata.js
+++ b/src/components/OSCALMetadata.js
@@ -1,30 +1,29 @@
-import React from "react";
-import { styled } from "@mui/material/styles";
-import Grid from "@mui/material/Grid";
-import Paper from "@mui/material/Paper";
-import Typography from "@mui/material/Typography";
+import BusinessIcon from "@mui/icons-material/Business";
+import ContactPageIcon from "@mui/icons-material/ContactPage";
+import EmailIcon from "@mui/icons-material/Email";
+import HelpOutlineIcon from "@mui/icons-material/HelpOutline";
+import HomeIcon from "@mui/icons-material/Home";
+import MapIcon from "@mui/icons-material/Map";
+import PhoneIcon from "@mui/icons-material/Phone";
+import SmartphoneIcon from "@mui/icons-material/Smartphone";
+import Avatar from "@mui/material/Avatar";
+import Button from "@mui/material/Button";
+import Card from "@mui/material/Card";
+import CardActions from "@mui/material/CardActions";
+import CardHeader from "@mui/material/CardHeader";
+import Dialog from "@mui/material/Dialog";
 import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
-import ContactPageIcon from "@mui/icons-material/ContactPage";
 import DialogTitle from "@mui/material/DialogTitle";
+import Grid from "@mui/material/Grid";
 import Link from "@mui/material/Link";
-import ListSubheader from "@mui/material/ListSubheader";
-import Card from "@mui/material/Card";
-import MapIcon from "@mui/icons-material/Map";
-import EmailIcon from "@mui/icons-material/Email";
 import List from "@mui/material/List";
 import ListItem from "@mui/material/ListItem";
-import CardContent from "@mui/material/CardContent";
-import Avatar from "@mui/material/Avatar";
-import GroupIcon from "@mui/icons-material/Group";
-import PhoneIcon from "@mui/icons-material/Phone";
-import CardActions from "@mui/material/CardActions";
-import Button from "@mui/material/Button";
-import HomeIcon from "@mui/icons-material/Home";
-import BusinessIcon from "@mui/icons-material/Business";
-import SmartphoneIcon from "@mui/icons-material/Smartphone";
-import Dialog from "@mui/material/Dialog";
-import HelpOutlineIcon from "@mui/icons-material/HelpOutline";
+import ListSubheader from "@mui/material/ListSubheader";
+import Paper from "@mui/material/Paper";
+import { styled } from "@mui/material/styles";
+import Typography from "@mui/material/Typography";
+import React from "react";
 import OSCALEditableTextField from "./OSCALEditableTextField";
 
 export const OSCALMetadataPartiesHeader = styled(ListSubheader)(
@@ -43,7 +42,7 @@ const OSCALMetadataLabel = styled(Typography)(({ theme }) => ({
   color: theme.palette.text.secondary,
 }));
 
-const OSCALMetadataTitle = styled(Grid)`
+const OSCALMetadataTitle = styled("div")`
   height: 56px;
 `;
 
@@ -55,11 +54,16 @@ const OSCALMetadataAdditional = styled(Paper)(
   ({ theme }) => `padding: ${theme.spacing(1)};`
 );
 
-const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
+const OSCALMetadataPartiesCardHolder = styled(Grid)(({ theme }) => ({
   backgroundColor: theme.palette.background.paper,
   position: "relative",
   overflow: "auto",
-  maxHeight: "18em",
+  maxHeight: "26em",
+  paddingBottom: "1em",
+  paddingRight: ".5em",
+  paddingTop: ".5em",
+  display: "flex",
+  direction: "row",
 }));
 
 // Returns a string with a locality-sensitive representation of this date
@@ -147,16 +151,6 @@ function OSCALMetadataPartyContactTypeHeader(props) {
 }
 
 export function OSCALMetadataPartyDialog(props) {
-  const [open, setOpen] = React.useState(false);
-
-  const handleOpen = () => {
-    setOpen(true);
-  };
-
-  const handleClose = () => {
-    setOpen(false);
-  };
-
   const PartyInfoTypes = {
     address: "address",
     telephone: "telephone",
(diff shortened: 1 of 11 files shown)


Bronstrom:
Looks like you got a good start to this issue!

I think `Card` works well here to organize the parties. When given multiple parties, the outlines around the `Card`s look a little wonky. I would suggest to remove the outlines.

I like the idea of using `CardActions` for more information — I think it would make sense to wrap all `Card` content for a party in one. When looking at one of our `OSCALCatalog` examples, it looks like most of the metadata is displaying underneath the party title and `Avatar` with some "undefined" values. Make sure to validate this info, and hide what's undefined.

Also, I think we should still display the secondary information, which I believe was the role title, we had previously underneath the party name at all times, while the additional metadata appears after selecting the `Card`s "action" area.

kylelaker:
> Make sure to validate this info, and hide what's undefined

For consistency in the UI, I'd rather not hide sections. I think having "This contact has no phone numbers" or whatever (with better wording) would be good. And I think if there aren't _any_ contact details, then we just don't provide the button

src/components/OSCALMetadata.js
```diff
@@ -38,17 +52,169 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+      break;
+    case "mobile":
+      icon = <SmartphoneIcon />;
+      break;
+    case "office":
+      icon = <BusinessIcon />;
+      break;
+    default:
+      break;
+  }
```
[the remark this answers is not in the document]
kylelaker: We need to have some kind of fallback because values here can also be locally defined.

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,174 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+
+const UNKNOWN_TYPE_ICON = <HelpOutlineIcon />;
+
+function getPhoneIcon(contactType) {
+  return TELEPHONE_ICON_MAPPING[contactType] ?? UNKNOWN_TYPE_ICON;
+}
+
+function getAddressIcon(contactType) {
+  return ADDRESS_ICON_MAPPING[contactType] ?? UNKNOWN_TYPE_ICON;
+}
```
kylelaker, line 81:
```suggestion
function getIconOfType(mapping, contactType) {
  if (!mapping || !contactType) {
    return UNKNOWN_TYPE_ICON;
  }
  return mapping[contactType] ?? UNKNOWN_TYPE_ICON;
}

function getPhoneIcon(contactType) {
  return getIconOfType(TELEPHONE_ICON_MAPPING, contactType);
}

function getAddressIcon(contactType) {
  return getIconOfType(ADDRESS_ICON_MAPPING, contactType);
}
```

Realizing that the `type` can be `null`/`undefined`, we probably want to harden this just a bit. Undecided on whether it's cleanest to extract the logic to a generic helper or not. But we at least want to handle `contactType` being falsey.

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,174 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+
+export function OSCALMetadataPartyTelephone(props) {
+  const { telephone } = props;
+  return (
+    <Typography>
+      {getPhoneIcon(telephone.type)}
+      <Link href={`tel:${telephone.number}`}>{telephone.number}</Link>
+    </Typography>
+  );
+}
```
kylelaker, line 115: I think these would make really good Storybook items. Please add these there. Actually, almost everything here probably would be a good addition.

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,181 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+      <address>
+        <Typography>
+          {`${address["addr-lines"]?.join(", ")} ${address.city} ${
+            address["postal-code"]
+          } ${address.country}`}
+        </Typography>
+      </address>
+    </>
+  );
+}
```
kylelaker, line 104:
```suggestion

export function OSCALMetadataPartyAddress(props) {
  const { address } = props;
  const addrString = [
    ...address["addr-lines"],
    `${address.city}, ${address.state} ${address["[postal-code"]}`,
    address.country
  ].filter(line => line).join("\n");
  return (
    <>
      {getAddressIcon(address.type)}
      <address>
        <Typography>
        {addrString}
        </Typography>
      </address>
    </>
  );
}
```

With better formatting, variable names, and error handling.

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,209 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+          <Grid container spacing={1}>
+            <Grid item xs={4}>
+              <Typography variant="h6" component="h3">
+                <MapIcon />
+                Addresses
+              </Typography>
+              <List>
+                {getPartyInfoList(
+                  props.party.addresses,
+                  "address",
```
kylelaker, line 188: Can you find a way to use an enum for this (and email and telephone when used in the same context)?

Bronstrom:
This is looking pretty good! I have a couple of style related suggestions:

I think the contact type (Addresses, Phone, and Email) would look nice if center aligned. Also here, I would suggest lining the icon up with the contact type (sorta like a vertical center alignment). Maybe use a Grid or resize the icons to where it fits nice

image

Bronstrom:
Also I would suggest placing "CONTACT" underneath the Party name like this, which can be achieved with a Grid:

image

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,215 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+      </ListItem>
+    ));
+  };
+
+  return (
+    <CardActions>
+      <Button size="small" onClick={handleOpen}>
+        Contact
+      </Button>
+      <Dialog
```
hreineck, line 174: Could some kind of shaded-in background be added so that this looks more like a button? I might be dumb but I didn't notice that the "contact" button was a clickable link at first.

hreineck (changes requested):
Great work! The new contact card looks awesome.
I have 2 small issues.

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,215 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+          <Grid item>
+            <Avatar>
+              {props.party.type === "organization" ? <GroupIcon /> : null}
+            </Avatar>
+          </Grid>
+          <Grid item>
+            <Typography>
+              {props.party.name}
+              {props.partyRolesText}
+            </Typography>
```
hreineck, line 252: This currently has the name and "roles" text to render directly next to each other. See FedRAMP's SSP Template for an example with roles text:
https://github.com/EasyDynamics/oscal-react-library/pull/576/

The older version in develop uses `ListItem` which seems to work nice for rendering the roles text on a secondary line:
https://github.com/EasyDynamics/oscal-react-library/blob/develop/src/components/OSCALMetadata.js#L121L125

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,215 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+          </Grid>
+        </Grid>
+      </CardContent>
+      <OSCALMetadataPartyDialog party={props.party} />
```
Bronstrom, line 256:
```suggestion
            </Typography>
            <OSCALMetadataPartyDialog party={props.party} />
          </Grid>
        </Grid>
      </CardContent>
```

This addresses my previous suggestion.

src/components/OSCALMetadata.js
```diff
@@ -38,17 +53,215 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+        key={partyInfoType === "email" ? item : `${item?.type}--${item?.name}`}
+      >
+        {TYPE_MAPPING(item)[partyInfoType]}
+      </ListItem>
+    ));
+  };
+
+  return (
+    <CardActions>
+      <Button size="small" onClick={handleOpen}>
```
Bronstrom, line 171:
```suggestion
      <Button size="small" variant="outlined" onClick={handleOpen}>
```

To give the user better indication of this being a button, we could also add an outline to it as well.

Bronstrom (changes requested):

src/components/OSCALMetadata.js
kylelaker: Also maybe more icons? https://mui.com/material-ui/material-icons/?query=contactpage&selected=ContactPage

tuckerzp:
@Bronstrom 
> I think the contact type (Addresses, Phone, and Email) would look nice if center aligned.

I am not sure how great it looks. 

image

image

src/components/OSCALMetadata.js
```diff
@@ -38,17 +54,218 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+                {getPartyInfoList(
+                  props.party["telephone-numbers"],
+                  PartyInfoTypes.telephone,
+                  "No telephone information provided"
+                )}
+              </List>
+            </Grid>
+            <Grid item xs={4}>
+              <Typography variant="h6" component="h3">
+                <AlternateEmailIcon />
```
Bronstrom, line 218: I think this is more typical for an email icon - [Material Icons - Material UI (mui.com)](https://mui.com/material-ui/material-icons/?query=email&selected=Email) instead of an @ symbol.

Bronstrom (dismissed):

src/components/OSCALMetadata.js
```diff
@@ -38,17 +59,226 @@ const OSCALMetadataPartiesCard = styled(Paper)(({ theme }) => ({
...
+}
+
+export function OSCALMetadataPartyTelephone(props) {
+  const { telephone } = props;
+  return (
+    <Typography>
+      {getPhoneIcon(telephone.type)}
+      <Link href={`tel:${telephone.number}`}>{telephone.number}</Link>
+    </Typography>
+  );
```
kylelaker, line 131: Can this use a similar structure to `OSCALMetadataPartyAddress`? Maybe even a common component:

```tsx
function TextWithIcon(props) {
  const { icon, text } = props;
  return (
    <Grid container spacing={1}>
      <Grid item>{icon}</Grid>
      <Grid item>
        <Typography>{text}</Typography>
      </Grid>
    </Grid>
  );
}
```

kylelaker (approved):
This looks really good and is so cool! At some point, we should support `location-uuid`s (or whatever that field is called) but I think that can be in another ticket (can you work on writing one up).

Thank you!!!

src/components/OSCALMetadata.js
kylelaker: This has been fixed in the newest revision of the code. The `partyRolesText` appears both in the card as well as in the modal.

src/components/OSCALMetadata.js
kylelaker: This has been implemented in the latest revision. Buttons to open the contact modal are outlined

src/components/OSCALMetadata.js
kylelaker: The suggested icon was adopted

src/components/OSCALMetadata.js
kylelaker: We used a stack!

src/components/OSCALMetadata.js
kylelaker: Per @Bronstrom's suggested, the implementation went with an outline. That matches the Material guidelines.

Bronstrom (approved):
Like the choices that were made to the structure and the Contact Dialog box looks pretty slick!

PR #582, opened 2022-08-23
Update to Google Analytics 4

nuttercd:
Currently react-ga does not support GA4. We can use the react-ga4
library for the time being while waiting for react-ga to be
updated. While the react-ga4 has not been updated in some
time, I have not found an alternative. React-ga4 does have 
some attention on react-ga's issue https://github.com/react-ga/react-ga/issues/493.


kylelaker (approved):
Okay cool! This seems to be working in Dev so I am good with this!

kylelaker:
@nuttercd please ping me internally once this has been merged.

package.json
```diff
@@ -26,6 +26,7 @@
     "react-dom": "^17.0.2",
     "react-error-boundary": "^3.1.4",
     "react-markdown": "^8.0.3",
+    "react-ga4": "^1.4.1",
```
tuckerzp, line 29: Why is this  `1.4.1` and the other `package.json` has it as `1.4.0`. This was not needed here before.

tuckerzp (changes requested):

tuckerzp (approved):

PR #583, opened 2022-08-23
Migrate tests to run on NodeJS 16

tuckerzp:
This is the last workflow that includes support for NodeJS 14. Because we operate primarily in a browser, we only really need Node for building. We moved the deployment pipeline to 16, Node 16 is now LTS, and this is a natural follow on to #521.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -41,10 +41,7 @@ jobs:
           # Lint YAML files
           yamllint --format github --config-file .yaml-lint.yml .
   build_and_test:
-    strategy:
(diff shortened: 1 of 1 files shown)


kylelaker (approved):

kylelaker:
Hmm this break sthe branch protection rules. We were previously only requiring 14 to pass. We probably should update that to 16.

folksgl:
There's no context in the git commit, PR description, code... @tuckerzp can you make sure the _why_ of this change is captured in the history here somewhere?

folksgl (approved):

PR #586, opened 2022-08-25
Temporarily fix npm dependency updates

tuckerzp:
Due to breaking changes tracked both on `tough-cookie` issue 246 & `jsdom`
 issue 3419, our tests are failing. This temporarily sets tough-cookie
to be version 4.0.0 so tests can pass and we can update other
dependencies. We will want to revert this change whenever their issues
are resolved.

See #581 for more context.


tuckerzp:
> I am also okay with just not merging the dependency update PR for a few days

Sounds good! I was mostly just playing around with it to see if tests passed and thought I might as well make a pr. I do not feel strongly at all about whether or not we merge this in.

PR #587, opened 2022-08-25
Display SSP Inventory Items

### LINKED ISSUE #224: Display SSP Inventory Items
As a user of the OSCAL Viewer, I want to see the Inventory Items of a System Security Plan ([example](https://github.com/GSA/fedramp-automation/blob/b895ed3ffbe28c36fb8448cb6af54990dc42edf9/dist/content/templates/ssp/json/FedRAMP-SSP-OSCAL-Template.json#L1495)) so that I can understand the managed inventory instances of the system.

### Acceptance Criteria
- An Inventory Items table with one row per Inventory Item is displayed in System Implementation if present in the OSCAL SSP
- The [known prop names](https://pages.nist.gov/OSCAL/reference/latest/system-security-plan/json-reference/#/system-security-plan/system-implementation/inventory-items) are displayed as columns in a form similar to the [FedRAMP Inventory Workbook](https://www.fedramp.gov/assets/resources/templates/SSP-A13-FedRAMP-Integrated-Inventory-Workbook-Template.xlsx)
  - In this iteration it would be sufficient if unknown/custom props were not displayed
- Any Responsible Parties defined are displayed in the table
- Any Implemented Components are displayed in the table
- Any remarks are displayed in the table
- It is not necessary to display links in this iteration

### Implementation Notes
The `inventory-items` array can be quite large, so some type of scroll or dropdown UI element should be used so that not all of them take up too much real estate at one time.

Bronstrom:
To understand the managed inventory instances of the system, this implements the [Inventory Items of a System Security Plan](https://pages.nist.gov/OSCAL/reference/latest/system-security-plan/json-reference/#/system-security-plan/system-implementation/inventory-items).

## Testing

To test a more elaborate set of `inventory-items` view in the SSP: https://github.com/GSA/fedramp-automation/blob/b895ed3ffbe28c36fb8448cb6af54990dc42edf9/dist/content/templates/ssp/json/FedRAMP-SSP-OSCAL-Template.json#L1495

Closes #224


src/components/OSCALImplementedComponents.js
```diff
@@ -16,8 +16,12 @@ const ImplementedComponentsTable = styled(TableContainer)(
 );
 
 export default function OSCALImplementedComponents(props) {
-  const getComponent = (compUuid) =>
-    props.components?.find((component) => component.uuid === compUuid);
+  function getComponent(compUuid) {
+    if (Array.isArray(props.components)) {
+      return props.components?.find((component) => component.uuid === compUuid);
+    }
+    return props.components?.uuid === compUuid ? props.components : null;
```
kylelaker, line 23: Wait can this actually happen? Have `components` that isn't an Array? Can you share a sample document where this happens and a link to the part of the OSCAL spec?

Bronstrom: So, I think some of our test data is out of date and a component was provided by itself and not in a list. I was thinking this check could be done in case the spec was ever changed, although this may be silly to include as it looks like this isn't supported currently.

Bronstrom: I believe `systemImplementationTestData` and `exampleSystemImplementation` in `SystemData` are causing the test failures. It looks like `components` doesn't contain a list.

kylelaker: Okay yeah I think _our_ test data is bad here. Can you fix the test data?

Bronstrom: Test data is resolved now.

src/components/OSCALSystemImplementation.test.js
```diff
@@ -27,35 +27,41 @@ export default function testOSCALSystemImplementation(
...
     const result = within(component).getByText("software");
     expect(result).toBeVisible();
   });
 
   test(`${parentElementName} shows component version`, () => {
     renderer();
-    const component = screen.getByText("Example Component").closest("tr");
+    const component = screen
+      .getByLabelText("An example component.")
+      .closest("tr");
```
Bronstrom, line 64: With the addition of component titles appearing in Inventory Items, this text (the component title) cannot be found as easy with it displaying twice. I was thinking label would be the next best option.

Bronstrom:
Currently this PR addresses the acceptance criteria as described in a way I believe makes sense and is displayed in a similar manner to how `System Implementation` is currently displayed.

I have made the description as the header for each inventory item "list" with the remark being the sub header underneath it. Each row is arranged by one of the `props` (which from my understanding is an item of the list) including the `name` (renamed to Item), `value`, `responsible-parties`, `implemented-components`, and the prop `remarks`. From my understanding, the Responsible Parties and Implemented Components are consistent for each item in a `inventory-item` list. I have also made a couple style changes like setting a minimum height, so not too much info overwhelms a user on the screen at a time, and a couple row coloring changes. Here is an example of how it looks:

image

I have a couple questions at this point.

The `ResponsibleRoles` class can be used for `responsible-parties` when it has been used for primarily for `responsible-roles`. If I were to create a separate class for `responsible-parties`, the functionality would be practically the same. Should this be renamed to make it a more general class or keep it as is?

I also plan to update the oscal-demo-content repo with more data for `inventory-items`. Should I also be addressing this feature in storybook as well?

kylelaker:
I am going to page @brian-easyd to fact check my understanding of the spec. But `responsible-roles` is specific to Component Definitions (at least of the document types we support this kind of information for today) and `responsible-parties` is specific to System Security Plans' System Implementations. I _think_ the spec sorta covers the difference when reading it:

> A reference to a set of organizations or persons that have responsibility for performing a referenced role in the context of the containing object.

https://pages.nist.gov/OSCAL/reference/latest/system-security-plan/json-reference/#/system-security-plan/system-implementation/inventory-items/responsible-parties

vs

> A reference to one or more roles with responsibility for performing a function relative to the containing object.

https://pages.nist.gov/OSCAL/reference/latest/component-definition/json-reference/#/component-definition/components/responsible-roles

Responsible Parties also have a reference to the actually _party_ that will be fulfilling that role in the system. I agree that the structure otherwise looks pretty similar. One could envision that when we add _editing_ capabilities, you'd be able to have a dropdown where you chose the specific parties for Responsible Parties (which you wouldn't be able to do with Responsible Roles). Also recall that with React, we generally prefer composition over inheritance.

I don't have a specific preference for how this is implemented in code (whether you completely write a new component from scratch or whether you find a way to make to compose `OSCALResponsibleRole` with the additional code for parties).

My answer would probably change here if we were talking about actual data-focused object-oriented classes but we're focusing on the presentation. And so long as we remember these are distinct and with different purposes, I think we should use the implementation that works best.

src/components/OSCALInventoryItem.js
```diff
@@ -0,0 +1,90 @@
...
+              <TableRow>
+                <StyledTableCell>Item</StyledTableCell>
+                <StyledTableCell>Value</StyledTableCell>
+                <StyledTableCell>Responsible Parties</StyledTableCell>
+                <StyledTableCell>Implemented Components</StyledTableCell>
+                <StyledTableCell>Remarks</StyledTableCell>
+              </TableRow>
+            </TableHead>
+            <TableBody>
+              {Object.values(props.inventoryItem.props).map((item) => (
```
tuckerzp, line 57: Would this `map` ever return more than one item? It seems like you are mapping a single item in `SystemImplementation` then trying to map that item object. That is probably why you needed to use `Object.values`

Bronstrom: I originally thought `inventory-items` was 0-infity, but it's 0-1, making this mapping unnecessary. Will make this change.

src/components/OSCALInventoryItem.js
```diff
@@ -0,0 +1,90 @@
...
+      <Grid item xs={12}>
+        <TableContainer sx={{ maxHeight: "15em" }}>
+          <Table aria-label="Inventory Items" sx={{ height: "max-content" }}>
+            <TableHead>
+              <TableRow>
+                <StyledTableCell>Item</StyledTableCell>
+                <StyledTableCell>Value</StyledTableCell>
+                <StyledTableCell>Responsible Parties</StyledTableCell>
+                <StyledTableCell>Implemented Components</StyledTableCell>
+                <StyledTableCell>Remarks</StyledTableCell>
```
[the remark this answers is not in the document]
Bronstrom: So `inventory-items` are organized into separate lists. I was creating separate tables to organize each of these lists. This `TableHead` sits at the top of these lists, not each individual item. To avoid the user from having to scroll up the page to see what the table head is, I included it for each list. I wouldn't mind making this change, if this makes more sense. Although, there can be 1-infinity `inventory-items`, so in the case there a ton, the user would have to do a lot of scrolling.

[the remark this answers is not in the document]
tuckerzp: Looking at `Components` we have usually just head a `header` then a new row for each `item`. I do not think we need a new header for each inventory item.

src/components/OSCALInventoryItem.js
tuckerzp: https://github.com/EasyDynamics/oscal-react-library/blob/72ad88399e0d5defb60c95d6a656976984ba0c72/src/components/OSCALSystemImplementation.js#L120-L124

I understand that, I think you are accidently mapping twice. There is a single list `Inventory-Items`. Inside the list there are `Inventory-item` objects, not more lists.

tuckerzp: Actually wait never mind. I am getting confused by the multiple tables for each item I think. If there are multiple tables, then this would be right.

Bronstrom:
I've restructured most of how this looks now. I had misunderstood that each item in `inventory-items` was an individual inventory item, rather I had organized these thinking the inventory item was a grouping for props (which I mistook for being the items). 

I've decided to keep most of the same structure, just now the rows are grouped by the inventory item (as it should be) and the props have their own column. This seems to much better fit the acceptance criteria reading it over.

I kept the same styling for the most part, but this can be changed back to the default. I think the blue background for the header helps differentiate itself from the inventory item rows beneath it and the gray and white alternating rows separate the items more visually. These styling changes could be implemented elsewhere where we have large tables.

image

src/components/OSCALProperties.js
```diff
@@ -0,0 +1,42 @@
...
+const PropertiesTable = styled(TableContainer)(
+  ({ theme }) => `
+  margin-top: ${theme.spacing(2)};
+  display: flex;
+  flex-direction: column;
+  text-transform: capitalize;
+`
+);
+
+export default function OSCALProperties(props) {
```
tuckerzp, line 18: Is this really `OSCALProperties`? Like does this encapsulate what it means everywhere we have `Properties`? I do not think that every component needs its own file. This seems very specific to `SystemImplementation`. I'd argue the same for `InventoryItems`. 

This also has a lot of overlap with #590. We should choose one way to do it and then just fix it in the other pr.

src/components/OSCALSystemImplementation.js
```diff
@@ -112,6 +113,11 @@ export default function OSCALSystemImplementation(props) {
               </TableContainer>
             </Grid>
           </Grid>
+          <OSCALInventoryItems
+            inventoryItems={props.systemImplementation["inventory-items"]}
+            parties={props.parties}
+            components={props.components}
+          />
```
tuckerzp, line 120: Is there a reason `InventoryItems` are both in `SSP` and `SystemImplementation`? I feel it makes more sense to just have it be part of `SystemImplementation` as that is also what the spec has.

src/components/OSCALImplementedComponents.js
```diff
@@ -0,0 +1,45 @@
...
+const ImplementedComponentsTable = styled(TableContainer)(
+  ({ theme }) => `
+  margin-top: ${theme.spacing(2)};
+  display: flex;
+  flex-direction: column;
+  text-transform: capitalize;
+`
+);
+
+export default function OSCALImplementedComponents(props) {
```
tuckerzp, line 18: Again I do not think that this is truly an encapsulation of `OSCALImplementedComponents`. It is `ImplementedComponents` in terms of what it means in `SystemImplementation`.

src/components/OSCALImplementedComponents.js
```diff
@@ -0,0 +1,45 @@
...
+  const getComponent = (compUuid) =>
+    props.components?.find((component) => component.uuid === compUuid);
+
+  return (
+    <ImplementedComponentsTable>
+      <Table size="small">
+        <TableBody>
+          {props.implementedComponents &&
+            Object.entries(props.implementedComponents).map(([key, comp]) => (
```
tuckerzp, line 27: I think you should be able to map `implementedComponents` without using `Objects.entires`. It should be an array. Same goes for a lot of places we are doing `Object.entries`.

src/components/OSCALProperties.js
Bronstrom: `props` is defined exactly the same throughout OSCAL. Some examples [profile/metadata/revisions/props](https://pages.nist.gov/OSCAL/reference/latest/profile/json-reference/#/profile/metadata/revisions/props), [plan-of-action-and-milestones/metadata/props](https://pages.nist.gov/OSCAL/reference/latest/plan-of-action-and-milestones/json-reference/#/plan-of-action-and-milestones/metadata/props), [system-security-plan/system-implementation/users/props](https://pages.nist.gov/OSCAL/reference/latest/system-security-plan/json-reference/#/system-security-plan/system-implementation/users/props). I think it would make sense to create a reusable component.

src/components/OSCALSystemImplementation.js
Bronstrom: Where does this appear in `SSP`? It should only be in `SystemImplementation`.

tuckerzp: I think it was a weird viewing thing on my end. You are right.

src/components/OSCALImplementedComponents.js
Bronstrom: I was thinking this would be reusable in a different context in the Assessment Layer, but it seems like `implemented-components` all fall underneath `local-definitions/inventory-items`. I think this can just be brought into `OSCALInventoryItems`.

src/components/OSCALImplementedComponents.js
Bronstrom: So just something like this:

```suggestion
            props.implementedComponents.map(([key, comp]) => (
```

tuckerzp: Most objects have a `uuid`. You should not need to use `key` I do not think. You should be able to just use `comp.uuid` as a key. Keep in mind that `comp.uuid` is an object, so you will need to use it as a `String` for the key.

tuckerzp:
```suggestion
            props.implementedComponents.map((comp) => (
```

src/components/OSCALProperties.js
Bronstrom: In many of these contexts we will. Currently, we create tables for `system-characteristic/props` and `component/props` in our application, where this can be reused. Maybe not necessarily everywhere. Perhaps this be made into `OSCALPropertiesTable`.

Bronstrom:
This issue has been migrated over to https://github.com/EasyDynamics/oscal-react-library/pull/590.

PR #594, opened 2022-08-30
Add aria label to Contact Button

tuckerzp:
In order to increase accessibility and give our Cypress tests something to look for, we added an aria label.


tuckerzp:
>  I don't love this workaround

I completely agree. It also seemed like the best way forward without needing to do some gross checking of `html` `classes`.

Bronstrom (approved):

Issue #88, opened 2021-07-02
Improved Catalog Control Group Navigation


rgauss:
As a user involved in the governance, risk, and compliance of systems and components, I want to be able to easily navigate different control groups and the controls within them in a catalog so that I can quickly get to and understand the details of those controls.

### Acceptance Criteria
- [x] The control groups in a catalog can be accessed in a left-hand navigation index
- [x] Clicking on a control group displays that group's controls
- [x] Individual controls are displayed in collapsible panel, initially closed

### Implementation Notes
* The [`Drawer Component`](https://mui.com/material-ui/react-drawer/#persistent-drawer) may be a good solution to displaying the navigation. We will need to make sure the naming is distinct to the what is currently called `DrawerSelector`.

Mockup:
![catalog-v2-initial](https://user-images.githubusercontent.com/168671/124276808-49dc4a00-db12-11eb-9570-5b039f0bb92d.jpg)

tuckerzp:
To clarify a question I originally had, this must be done before https://github.com/EasyDynamics/oscal-react-library/issues/176. Our initial implementation will display a control group's controls by using the state. We will later on use anchor links to clean this up a bit in the future issue.

PR #600, opened 2022-09-01
Add support for anchor links

### LINKED ISSUE #176: Support for Anchor Links
As a user of the OSCAL Viewer/Editor, I want to be able to quickly navigate to sections of OSCAL documents, including specific controls and control implementations, using anchor links so that I don't have to scroll through often large pages and can bookmark/share specific sections.

### Implementation Notes (and Original Submission)
Within the Catalog Viewer, it would be really helpful if I could link directly to a control within the catalog; for example, if navigating to https://oscal-viewer.msd.easydynamics.com/catalog?url=https://raw.githubusercontent.com/usnistgov/oscal-content/master/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_catalog.json#SI-3 were to automatically result in "System and Information Integrity" expanded with "SI-3 Malicious Code Protection" focused in the viewport. Perhaps each control could also have a way to copy a link to it (for example, the way that you can do so for headings on the [Microsoft Docs](https://docs.microsoft.com/en-us/windows/wsl/install-on-server#extract-and-install-a-linux-distribution)).

The link should use the `#`/fragment in the URL rather than a querystring parameter for determining where in the page to navigate to, in order to support browsers' native capabilities to do so. This could either be handled by adding an `id` attribute to the headings or placing an `<a>` tag with a name directly above the `<hX>` tags.

I am not sure whether this feature request cleanly maps to the other viewer types.

Bronstrom:
With the addition of anchor links to OSCAL Viewer/Editor users can quickly navigate between controls and sections of an OSCAL document.

In order to implement this, the class: `OSCALAnchorLinkHeader` was created to create an "anchored" header. This class sets the headers id to a specific id value when provided (or converts the header title), and makes the header hoverable to make an anchor link icon appear. Our anchor links work very similarily to heading in [Microsoft Docs](https://docs.microsoft.com/en-us/windows/wsl/install-on-server#extract-and-install-a-linux-distribution). The Loader was also updated to scroll directly to each "anchored" header, when a hash is specified in the browser URL and the hash matches an "anchored" header. This occurs specifically on hash updates and after reloads.

### Testing
-  Anchor links are usually based on the section title lower cased and separated by "-".
- An example of this is: `/system-security-plan/?url=https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/main/system-security-plans/ssp-example.json#authorization-boundary`

Closes #176


src/components/OSCALAnchorLinkHeader.js
```diff
@@ -0,0 +1,39 @@
...
+      {props.title}
+      {!hover ? (
+        ""
+      ) : (
+        <Link to={{ hash: `#${editString(props.title)}` }}>
+          <Fade in={hover}>
+            <LinkIcon />
+          </Fade>
+        </Link>
+      )}
```
kylelaker, line 36: Can you clarify for me how this is different from

```suggestion
      {!hover && (
        <Link to={{ hash: `#${editString(props.title)}` }}>
          <Fade in={hover}>
            <LinkIcon />
          </Fade>
        </Link>
      )}
```

src/components/OSCALLoader.js
```diff
@@ -233,6 +235,19 @@ export default function OSCALLoader(props) {
...
+  // Handle anchor link change in window url
+  useEffect(() => {
+    // TODO: Handle GET, when the hash hasn't changed & initial navigation to page.
+    // This outputs a 404 error at this point in the console & doesn't take the user
+    // to the desired section of page based on hash.
+    const { hash } = window.location;
+    // Determine a hash exists and grab it
+    const hashStr = hash && document.getElementById(hash.substring(1));
+    if (hashStr) {
+      hashStr.scrollIntoView({ behavior: "smooth" });
```
kylelaker, line 247: Can we maybe call this something like `elementToNavigateTo` (that's probably too long) or something like that? `hashStr` doesn't really explain what this is correctly. Even like `matchingElement` might work.

src/components/OSCALSystemImplementationInventoryItems.js
```diff
@@ -80,7 +81,10 @@ export default function OSCALOSCALSystemImplementationTableTitleInventoryItems(
   return (
     <>
       <OSCALSystemImplementationTableTitle variant="h6" id="tableTitle">
-        Inventory Items
+        <OSCALAnchorLinkHeader
+          title="Inventory Items"
+          style={{ "padding-top": "1em", "padding-bottom": "1em" }}
+        />
```
kylelaker, line 87: If we're consistently applying this style, can we just do that for `OSCALAnchorLinkHeader`?

kylelaker (commented):
Very early thoughts

src/components/OSCALAnchorLinkHeader.js
Bronstrom: There really isn't much difference, what you're suggesting is much cleaner. Originally to make sure that there was appearing after the header text, I had some test text, and ended up just making it an empty string. The ternary operator isn't needed anymore. Will make the change.

src/components/OSCALSystemImplementationInventoryItems.js
Bronstrom: This padding style will be something consistent for only the `system-implementation` header styling, which will occur three times. I was trying to decide if this would be something I would want to make into a separate style that would be included in `OSCALTableStyles`.

src/components/OSCALLoader.js
```diff
@@ -140,7 +140,9 @@ export default function OSCALLoader(props) {
   };
 
   const handleUrlChange = (value) => {
-    setOscalUrl(value);
+    let tempValue;
+    if (value.contains("#")) tempValue = value.split("#").pop();
+    setOscalUrl(tempValue);
```
kylelaker, line 145:
```suggestion
    if (value.contains("#")) {
      setOscalUrl(value.split("#").pop());
    }
    setOscalUrl(undefined);
```

Is it actually intentional that this can be set to `undefined` in so may situations? Also should we _really_ be pulling this by splitting on `"#"`? Or should we use the `URL()` constructor and [`hash`](https://developer.mozilla.org/en-US/docs/Web/API/URL/hash)

src/components/OSCALLoader.js
```diff
@@ -233,6 +235,19 @@ export default function OSCALLoader(props) {
     }
   }, [props.isRestMode]);
 
+  // Handle anchor link change in window url
+  useEffect(() => {
+    // TODO: Handle GET, when the hash hasn't changed & initial navigation to page.
+    // This outputs a 404 error at this point in the console & doesn't take the user
+    // to the desired section of page based on hash.
```
kylelaker, line 242: I don't think this is just a generic TODO, it feels like it's in-scope for the initial implementation of this feature. Can you elaborate a bit on what's going wrong? With real examples?

src/components/OSCALSystemImplementationInventoryItems.js
kylelaker: I definitely want it de-duplicated somehow

src/components/OSCALLoader.js
Bronstrom: This was a note to self. This is definitely something that needs to be handled in this issue.

src/components/OSCALLoader.js
Bronstrom: I'll admit this part is a bit rough and unfinished. I was investigating why there was a 404 error in the console when refreshing the page with a provided hash.

The idea here is that if the URL contains a hash it is removed from the URL, otherwise the URL would be handled normally. In the case of a URL containing a hash, we still want to keep track of the hash to navigate to after the page loads, so it is stored in `tempValue`. I was planning to migrate some of the code written in the `useEffect()` below into it's own constant, to be called at the end of this function with `tempValue` as it's parameter.

As you mentioned, I think `hash` would make more sense to handle and exclude here.

tuckerzp:
I am taking this out of review since the functionality is mostly there. We need #88 to be in before we can be sure things work perfectly.

tuckerzp:
@kylelaker 

I have realized that how we deal with scrolling to section headers will not work with our controls listed in the catalog. We will need a more specialized version of `handleHash`.

As an initial implementation, I thought it may be helpful to use a regex to check if a hash is a control id. Here is an initial attempt: `[A-z][A-z]-[0-9]*(.[0-9]*)?`. I have not used regex in a while, so any suggestions to clean it up would be welcome.

Bronstrom:
After a few months I've returned back to this issue with an imperfect and hacky solution to help solve navigation to controls from an anchor hash. The main issue is that the controls need to be navigated to by selecting their control grouping once loaded, otherwise they are completely skipped over when using `getElementById()` due to the controls being hidden when their control grouping is not appearing.

One idea I had was to interact with the collapsible states of the control groupings, but it seems this would get fairly messy with the control groupings being buried deep within the components and not easily interactable during the loading stage. I ended up choosing to adjust the IDs for the control grouping tabs and interacting with the DOM by using the `click()` function - which is definitely not the most ideal. Unfortunately, it appears `click()` causes this error: `Warning: unstable_flushDiscreteUpdates: Cannot flush updates when React is already rendering.`

Let me know if anyone has a better idea for solving this.

Bronstrom:
The console error has been removed now by avoiding any mutation with the DOM. I've added hooks handling opening and closing states within `OSCALCatalogGroups` and `OSCALCatalogGroup` which are triggered by the hash being changed within the browser search bar. I've also added an additional function to `OSCALLinkUtils` which determines if the hash is a control, as this was used in the components mentioned and `OSCALLoader`.

Double checking @tuckerzp's [list in the linked issue](https://github.com/EasyDynamics/oscal-react-library/issues/176#issuecomment-1249612630), all anchor links on the page should be covered, with "Control & Sub Controls `control-id`" and "Controls Groups `control-groups`" completed latest. I think the last part that needs to be handled for this issue are the tests, which make most sense to be Cypress tests in my opinion. Test cases include that the url contains the correct hash after an anchor link is selected and that a hash will navigate to a part of the page when provided in the url.

src/components/OSCALControl.js
```diff
@@ -101,16 +102,18 @@ export default function OSCALControl(props) {
...
-              variant="h6"
-              component="h2"
-              style={props.childLevel ? { fontSize: "1.1rem" } : undefined}
-            >
-              <span style={{ textTransform: "uppercase" }}>
-                {props.control.id}
-              </span>{" "}
-              {props.control.title} {modificationDisplay}
-            </Typography>
+            <OSCALAnchorLinkHeader value={props.control.id}>
```
tuckerzp, line 105: Could this work?

```suggestion
            <OSCALAnchorLinkHeader value={props.control.id} isControl>
```

Bronstrom: I like the thought, although this will not solve the issue of whether the provided fragment in the URL is a control or not. I have a feeling we will still need some sort of check that is done at a high level to determine whether the format follows a certain control scheme. Right now it just fits NIST SP-800-53 Rev 5. 

I had thought about making the fragment include "control" to help make the control check easier, although "#control-ac-1" doesn't look as clean in my opinion. Furthermore, we would need to validate the fragment isn't "control-implementation", "control-groups", or any other fragment starting with "control".

Bronstrom:
I've gone ahead and added useCallback()s to useEffect()s editted directly in this PR, although it seems the linter warnings persist and is upset by `window.location.hash` being used with a hook. This is the warning that appears when running the linter for a given file with `window.location.hash` in a hook:

image

I am a bit stumped at this point as it seems like we will need to capture `window.location.hash` at least once to know when the fragment is being changed in the url. Not sure if would be best to just ignore linter cases where this is used within a hook or keep searching for a better work around to capture the fragment.

I've also gone ahead and removed most occurrences of "hash", although not all. "hash" is the default term used to describe the fragment identifier of the url in native JS, such as mentioned several times above: `window.location.hash`. Otherwise, for variable and functions created containing "hash", this has been replaced with "fragment" instead.

tuckerzp (commented):
@Bronstrom thanks for your work on this! This has not been an easy thing to implement and I think it will be so cool once it is actually merged in. I had a couple comments.

I also am a bit confused by the linter warning. I don't think we should just let the warning occur though. Worst case, can we add a linter ignore there?

Also while playing around with the Catalog, I noticed some weird behavior. Let's say we have the url `http://localhost:3000/#si-14`. If I try to click and open up another control like `si-4`, the screen automatically scrolls back down to `si-14`. I feel like when you load the url, the screen should change but otherwise ignore the fragment.

kylelaker:
> I am a bit stumped at this point

I am confused -- does something break if you remove it from the dependency array? It seems the error tells you exactly what it wants you to do and why. `window.location.hash` does not result in re-rendering as an outer-scope variable and therefore serves no function in the array. Either remove it from the array or (if it is the only item in the array) remove the array.

Bronstrom:
> If I try to click and open up another control like si-4, the screen automatically scrolls back down to si-14. I feel like when you load the url, the screen should change but otherwise ignore the fragment.

I missed this. Sounds like this could be some unintended functionality when I was adding in sub-control check. I will look more into this.

> I am confused -- does something break if you remove it from the dependency array?

Unfortunately, yes. When removing `window.location.hash` from the hook array, the fragment is no longer captured in the url. Meaning whenever a fragment is provided or changed in the url, none of the intended functions are triggered. I don't like going this direction, but I was considering adding a linter ignore as @tuckerzp had mentioned.

kylelaker:
> I was considering adding a linter ignore as @tuckerzp had mentioned

Please do not. We have a few of these warning across the code base that we need to resolve. And we should evaluate all them them when we do. That includes re-evaluating this one at that time.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -61,3 +61,22 @@ export function guessExtensionFromHref(url) {
...
 }
+
+/**
+ * Finds if a fragment contains a control group.
+ * NOTE: Control groupings will contain two letters, followed by a "-" and a positive integer
+ *
+ * @param frag A given fragment.
+ * @returns A string providing the control group.
+ */
+export function determineControlGroupFromFragment(frag) {
```
kylelaker, line 72:
```suggestion
export function determineControlGroupFromFragment(fragment) {
```

There's no need to shorten this.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -61,3 +61,22 @@ export function guessExtensionFromHref(url) {
   }
   return detectedExtension;
 }
+
+/**
+ * Finds if a fragment contains a control group.
+ * NOTE: Control groupings will contain two letters, followed by a "-" and a positive integer
```
kylelaker, line 67: Will they? This feels _very_ SP 800-53 specific.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -61,3 +61,22 @@ export function guessExtensionFromHref(url) {
...
+ */
+export function determineControlGroupFromFragment(frag) {
+  // Create array from all tab control grouping elements
+  const controlGroupList = Array.from(
+    document.querySelectorAll('[id^="vertical-tab-"]')
+  );
+  // Grab group id if found
+  return controlGroupList
+    ?.find((group) => frag.includes(group.id.split("vertical-tab-").pop()))
+    ?.id?.split("vertical-tab-")
```
kylelaker, line 80:
```suggestion
    ?.id
    ?.split("vertical-tab-")
```

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -61,3 +61,22 @@ export function guessExtensionFromHref(url) {
   }
   return detectedExtension;
 }
+
+/**
+ * Finds if a fragment contains a control group.
+ * NOTE: Control groupings will contain two letters, followed by a "-" and a positive integer
+ *
+ * @param frag A given fragment.
+ * @returns A string providing the control group.
```
kylelaker, line 70: What does this mean... "A string providing the control group"

src/components/OSCALCatalogGroup.js
```diff
@@ -30,20 +32,58 @@ const StyledControlDescriptionWrapper = styled("div")`
...
 function CollapseableListItem(props) {
   const [open, setOpen] = React.useState(false);
+  const [listItemOpened, setListItemOpened] = React.useState(false);
 
   const handleClick = () => {
     setOpen(!open);
   };
 
+  const handleFragment = useCallback(() => {
+    if (!listItemOpened) {
```
kylelaker, line 42: We can un-indent a bunch of this code by doing:

```
if (listItemOpened) {
  return;
}
// REST OF THE CODE
```

src/components/OSCALCatalogGroup.js
```diff
@@ -30,20 +32,58 @@ const StyledControlDescriptionWrapper = styled("div")`
...
+      const { hash } = window.location;
+      // Ensure fragment exists and grab element associated
+      const controlFragment = hash?.substring(1);
+      const controlGroupingFragment =
+        determineControlGroupFromFragment(controlFragment);
+      // Locate the element with the provided fragment and scroll to the item
+      if (controlGroupingFragment) {
```
kylelaker, line 49: Same comment. We don't have to nest so deeply

src/components/OSCALCatalogGroup.js
```diff
@@ -30,20 +32,58 @@ const StyledControlDescriptionWrapper = styled("div")`
...
   const [open, setOpen] = React.useState(false);
+  const [listItemOpened, setListItemOpened] = React.useState(false);
 
   const handleClick = () => {
     setOpen(!open);
   };
 
+  const handleFragment = useCallback(() => {
+    if (!listItemOpened) {
+      const { hash } = window.location;
```
kylelaker, line 43: Can we use the `useLocation` hook?

src/components/OSCALCatalogGroup.js
```diff
@@ -30,20 +32,58 @@ const StyledControlDescriptionWrapper = styled("div")`
...
+      // Ensure fragment exists and grab element associated
+      const controlFragment = hash?.substring(1);
+      const controlGroupingFragment =
+        determineControlGroupFromFragment(controlFragment);
+      // Locate the element with the provided fragment and scroll to the item
+      if (controlGroupingFragment) {
+        // Find control list state and open collapsable item
+        if (controlFragment.includes(props.itemText?.props?.value)) {
```
kylelaker, line 51: I'd really rather we not fetch the `props` of a passed Component... This feels very hacky and is going to become rather complicated when we need to start adding type information I fear.

Is there a better way we can represent this?

src/components/OSCALCatalogGroup.js
```diff
@@ -30,20 +32,58 @@ const StyledControlDescriptionWrapper = styled("div")`
...
+          setOpen(true);
+        }
+        const elementWithFragment = document.getElementById(controlFragment);
+        if (elementWithFragment) {
+          elementWithFragment.scrollIntoView({ behavior: "smooth" });
+        }
```
kylelaker, line 57:
```suggestion
        elementWithFragment?.scrollIntoView?.({ behavior: "smooth" });
```

src/components/OSCALCatalogGroups.js
```diff
@@ -52,32 +53,64 @@ const ComponentTab = styled(Tab)(({ theme }) => ({
...
+    const controlGroupingFragment = determineControlGroupFromFragment(hash);
+    // Determine higher-level control or sub-control exists within catalog group
+    const catalogControl =
+      props?.groups
+        ?.find((group) => group.id === controlGroupingFragment)
+        ?.controls?.find((control) => control.id === hash?.substring(1)) ||
+      props?.groups
+        ?.find((group) => group.id === controlGroupingFragment)
+        ?.controls?.find((control) => hash?.substring(1)?.includes(control.id))
+        ?.controls?.find((subcontrol) => subcontrol.id === hash?.substring(1));
```
kylelaker, line 87: I don't think this will work for arbitrarily-nested control depths, correct?

src/components/OSCALCatalogGroups.js
```diff
@@ -52,32 +53,64 @@ const ComponentTab = styled(Tab)(({ theme }) => ({
...
 
 TabPanel.propTypes = {
-  index: PropTypes.number.isRequired,
-  value: PropTypes.number.isRequired,
+  value: PropTypes.string.isRequired,
 };
 
 export default function OSCALCatalogGroups(props) {
-  const [value, setValue] = React.useState(0);
+  const [value, setValue] = React.useState(props?.groups[0]?.id);
```
kylelaker, line 68: There must be a better name for this than `value`? If we're refactoring a bunch of this anyway...

kylelaker (changes requested):

src/components/oscal-utils/OSCALLinkUtils.js
Bronstrom: My bad. I should have rewritten this comment after making the below function support all control schemes. Will fix this.

src/components/oscal-utils/OSCALLinkUtils.js
Bronstrom: The linter for some reason favors it to be all on one line. I do prefer it to look like this instead.

src/components/oscal-utils/OSCALLinkUtils.js
Bronstrom: Do I need to go more in depth on what a control group is or rephrase this returns comment? I'm just saying: "the function returns a string that is a control group."

src/components/OSCALCatalogGroups.js
Bronstrom: I totally agree. When looking tracing the code at first, I was confused what `value` was referring to. From my understanding this is the current control grouping tab that is open. Perhaps this is renamed to something like "openControlGroupingTab" or simply "openTab".

src/components/oscal-utils/OSCALLinkUtils.js
Bronstrom: I'll add in the word "title" - or honestly "id" probably fits better. Is word choice of "providing" not clear? I've commonly written my return statements like this, including other places in this file.

tuckerzp: I'd update the other ones too unless Kyle disagrees. I think it just sounds a bit awkward. I think the string just is `the control group id` right? The string doesn't "provide" us anything.

I realize this is a bit pedantic, but it just sounds weird 😄

kylelaker: "providing" is very confusing -- "Control group ID" makes much more sense. And @tuckerzp is correct that we should put the type in braces. So:

```
@param {string} the control group ID
```

is probably most correct. Please correct this occurrence and any others.

Bronstrom: I want to make sure I get this right. `@param` we add braces too around the type, but `@returns` we don't, right? I'm noticing most returns we don't explicitly say the type on

tuckerzp: I personally prefer stating the type for the return when we can. I would be hesitant to base decisions on some of our past coding practices.

kylelaker: It is something I want to start doing and to improve as we go forward. Just because we haven't _always_ done the best/correct thing does not mean we should avoid iterative improvement. https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

Bronstrom: Created an issue to revise our previous JSDoc: https://github.com/EasyDynamics/oscal-react-library/issues/724

src/components/OSCALAnchorLinkHeader.js
```diff
@@ -0,0 +1,54 @@
...
+                }),
+              ]}
+              aria-label={`${
+                props.value || editString(props.children)
+              } anchor link`}
+            />
+          </Fade>
+        </Link>
+      )}
+    </Stack>
```
kylelaker, line 52:
```suggestion
  const linkId = props.value || editString(props.children);
  
  return (
    <Stack
      direction="row"
      gap={1}
      alignItems="center"
      onMouseEnter={onEnter}
      onMouseLeave={onLeave}
      id={linkId}
    >
      {props.children}
      {hover && (
        // Navigate to a specified fragment or use the child text
        <Link
          reloadDocument
          to={`#${linkId}`}
        >
          <Fade in={hover}>
            <LinkIcon
              sx={[
                (theme) => ({
                  width: "0.75em",
                  height: "0.75em",
                  color: theme.palette.grey[500],
                }),
              ]}
              aria-label={`${linkId} anchor link`}
            />
          </Fade>
        </Link>
      )}
    </Stack>
```

src/components/OSCALCatalogGroup.js
Bronstrom: I like this suggestion as it feels like the more "React approach" to the problem. This will remove all linter errors, however, React Testing Library is not very happy. I'm not entirely sure if this is a better choice to go at this point (over `window.location`) as I haven't dug too deep into the test failures, but it seems like the main issue is that `useLocation()` must be used in the context of a route. I've committed my changes as this is one of the more major changes I've made but can revert it if need be.

Bronstrom:
> Also while playing around with the Catalog, I noticed some weird behavior. Let's say we have the url `http://localhost:3000/#si-14`. If I try to click and open up another control like `si-4`, the screen automatically scrolls back down to `si-14`. I feel like when you load the url, the screen should change but otherwise ignore the fragment.

Looking more into @tuckerzp's introduced test results, I'm noticing this is becoming more difficult to eliminate than I was predicting, and I spent a good chunk of this morning and yesterday, not getting very far. I'm going to describe as best as I can at a high level and not get too much into the weeds in explaining the architecture approach I'm thinking of and let me know if this sounds like the right path.

So, the main issue is that we need a way to track if the fragment has already been dealt with between the several component layers of Catalog, CatalogGroups, and controls. I'm thinking this interaction is necessary, as the controls/sub-controls seem to re-render whenever you close/open a higher-level control or select a different grouping and then return back to the grouping with a control matching the current fragment. Currently the controls are still navigated too (high-level control uncollapses and scrolls right to it). The intended functionality is that this scrolling and opening only happens once the user has entered a new fragment into the URL.

Right now I'm thinking some state should be created at a higher level (like CatalogGroups) and then once the control has been opened and navigated to the child should trigger a state change. It seems like the best method of doing this in React is to create a callback function and pass it to the child component(s). Currently I'm having trouble doing so. Most of the time I'm getting `Uncaught TypeError: "x" is not a function` at the lowest level of the component chain (the control/sub-control) when trying to implement the callback function. Maybe I'm not completing the chain of passing the callback function down as there is a lot of component layers making it fairly complex (and the code doesn't seem the cleanest in my opinion).

I'm wondering if this sounds like a decent approach to the problem still or I should try something else out.

src/components/OSCALCatalogGroup.js
Bronstrom: I ended up just refactoring a portion of the code in order to acquire the fragment in the context of a route (in App) and just passed down the fragment's state to `OSCALLoader` for other components to handle. This seems to resolve all cases of: `[Error: useLocation() may be used only in the context of a <Router> component`, while also avoiding linter warnings that were previously added in this PR.

src/components/OSCALCatalogGroup.js
```diff
@@ -130,13 +130,25 @@ function OSCALCatalogControlListItem(props) {
 
 function OSCALCatalogGroupList(props) {
   return (
-    <CollapseableListItem itemText={props.group.title}>
+    <CollapseableListItem
+      itemText={props.group.title}
+      controlId={props.control?.id}
+      urlFragment={props.urlFragment}
```
tuckerzp, line 143: I would love if we could begin destructuring more of these `props.___` things. At the very least, can we do that where we are using `urlFragment`?

Bronstrom:
> a new state should be created at a higher level (like CatalogGroups) and then once the control has been opened and navigated to the child component should trigger a state change.

After rewriting this a few times, I've finally got this to work how I wanted it to. I might have been missing a point I was passing the callback function, causing the previous attempts to fail.

> * A different control grouping is selected and then return back to the previous control grouping

This bug has been resolved and gets me a lot closer to solving the other bug. There is a little more clean up that will need to be done at the `OSCALControl`/`OSCALControlGrouping` levels, but it shouldn't be too difficult to solve now.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -61,3 +61,24 @@ export function guessExtensionFromHref(url) {
...
+  }
+  // Create array from all tab control grouping elements
+  const controlGroupList = Array.from(
+    document.querySelectorAll('[id^="vertical-tab-"]')
+  );
+  // Grab group id if found
+  return controlGroupList
+    ?.find((group) => fragment.includes(group.id.split("vertical-tab-").pop()))
+    ?.id?.split("vertical-tab-")
+    ?.pop();
```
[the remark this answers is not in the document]
Bronstrom: Yes, this makes the assumption the group ID contains the group ID. I will plan to include the control ID and the group ID in the URL going forward.

src/components/OSCALAnchorLinkHeader.js
```diff
@@ -0,0 +1,52 @@
...
+  const onEnter = () => {
+    setHover(true);
+  };
+  const onLeave = () => {
+    setHover(false);
+  };
+
+  const editString = (str) =>
+    str.props?.children.replaceAll(" ", "-")?.toLowerCase() ||
+    str.replaceAll(" ", "-").toLowerCase();
```
kylelaker, line 19: What is this trying to do? What is `str`? Why does a `str` potentially have props (it's clearly not a string). What does "edit string" mean? Is this a string meant to be edited or just something that edits strings?

Please name this something more meaningful and give some context to the expected types for this.

src/components/OSCALAnchorLinkHeader.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const linkId = value || editString(children);
```
kylelaker, line 21: What happens if both of these are an empty string?

src/components/OSCALCatalogGroup.js
```diff
@@ -36,51 +38,111 @@ const StyledControlDescriptionWrapper = styled("div")`
...
+  const {
+    urlFragment,
+    control,
+    itemText,
+    children,
+    listItemOpened,
+    setListItemOpened,
+  } = props;
   const [open, setOpen] = React.useState(false);
+  const [itemNavigatedTo, setItemNavigatedTo] = React.useState(false);
```
kylelaker, line 51: For booleans, we should use the `is` convention.

```suggestion
  const [isItemNavigatedTo, setIsItemNavigatedTo] = React.useState(false);
```

We already have a few warnings around this because of `withdrawn`.

src/components/OSCALCatalogGroup.js
```diff
@@ -91,28 +153,71 @@ function OSCALCatalogControlListItem(props) {
...
     <OSCALControlList>
-      {props.group.groups?.map((innerGroup) => (
-        <OSCALCatalogGroupList group={innerGroup} key={innerGroup.title} />
+      {group.groups?.map((innerGroup) => (
+        <OSCALCatalogGroupList
+          group={innerGroup}
+          key={innerGroup.title}
+          urlFragment={urlFragment}
+          controlListItemOpened={controlListItemOpened}
+          setControlListItemOpened={setControlListItemOpened}
```
kylelaker, line 210: These both need to have `is` added if they're booleans too.

src/components/OSCALCatalogGroups.js
```diff
@@ -52,32 +53,69 @@ const ComponentTab = styled(Tab)(({ theme }) => ({
   textTransform: "none",
 }));
 
-function a11yProps(index, title) {
+function a11yProps(groupId) {
```
kylelaker, line 56: Why do we no longer use the title?

src/components/OSCALCatalogGroups.js
```diff
@@ -52,32 +53,69 @@ const ComponentTab = styled(Tab)(({ theme }) => ({
...
+      return;
+    }
+    // Confirm catalog tab group can be grabbed
+    const elementWithFragment = document.getElementById(
+      `vertical-tab-${controlGroupingFragment}`
+    );
+    if (elementWithFragment) {
+      setOpenTab(controlGroupingFragment);
+    }
+  }, [urlFragment, groups]);
```
kylelaker, line 104: How is this different from the code in `OSCALCatalogGroup.js`? It seems like there's a lot of repetition. Can the code here be reusable? Or at least broken into chunks or composed?

src/components/OSCALLoader.js
```diff
@@ -219,7 +245,11 @@ export default function OSCALLoader(props) {
         "",
         "",
         `/${props.oscalObjectType.jsonRootName}/${
-          props.isRestMode ? `${oscalObjectUuid}` : ""
+          props.isRestMode
+            ? `${oscalObjectUuid}${
+                props.urlFragment ? `#${props.urlFragment}` : ""
+              }`
```
kylelaker, line 251: This is getting very nasty. Please calculate this value _before_ the `window.history.pushState` call and then pass that variable here.

```
const fragment = props.urlFragment ? `#${props.urlFragment}` : "";
const document = props.isRestMode ? oscalObjectUuid : "";
const path = `/${props.oscalObjectType.jsonRootName}/${document}${fragment}`;

File: src/components/OSCALLoader.js [hunk 28]
```diff
@@ -233,6 +263,11 @@ export default function OSCALLoader(props) {
     }
   }, [props.isRestMode]);
 
+  // Handle anchor link change in window url
+  useEffect(() => {
+    handleFragment();
+  }, [handleFragment]);
```
kylelaker, line 269: Why not just have `handleFragment()`'s body here? I don't see where it gets used/passed anywhere else. And we can just change the triggers... Basically, why not just change the `useCallback` above to a `useEffect`?

src/components/OSCALMetadata.js
```diff
@@ -595,11 +604,33 @@ export function OSCALMetadataLocation(props) {
...
+    const controlFragment = urlFragment !== "" ? urlFragment : null;
+    // Expand metadata accordion section if control fragment matches title
+    if (controlFragment === title.toLowerCase()) {
+      setExpand(true);
+    }
+  }, [urlFragment, title]);
+
+  useEffect(() => {
+    handleFragment();
+  }, [handleFragment]);
```
kylelaker, line 626: Same here actually. I am confused why we can't just do:


```suggestion
  useEffect(() => {
    // Grab fragment identifier following hash character if fragment exists in location
    const controlFragment = urlFragment !== "" ? urlFragment : null;
    // Expand metadata accordion section if control fragment matches title
    if (controlFragment === title.toLowerCase()) {
      setExpand(true);
    }
  }, [urlFragment, title]);
```

kylelaker (changes requested):

src/components/oscal-utils/OSCALLinkUtils.js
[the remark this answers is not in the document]
Bronstrom: Fragments are now written like `{group.id}/{control.id}/{subcontrol.id}/{subcontrol.id}/...`, or like `at/at-2/at-2.2` in SP 800-53, in order to support more control schemes.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -61,3 +61,52 @@ export function guessExtensionFromHref(url) {
...
+  }
+  // Create array from all tab control grouping elements
+  const controlGroupList = Array.from(
+    document.querySelectorAll('[id^="vertical-tab-"]')
+  );
+  // Grab group id if found
+  return controlGroupList
+    ?.find(
+      (group) =>
+        fragment.split("/")[0] === group.id.split("vertical-tab-").pop()
```
kylelaker, line 83:
```suggestion
        fragment.split("/")?.[0] === group.id.split("vertical-tab-").pop()
```

src/components/OSCALCatalogGroup.js
Bronstrom: Gotcha, didn't realize this was the convention of writing booleans. Will keep this in mind.

src/components/OSCALAnchorLinkHeader.js
Bronstrom: Good point, `str` is the string that is being turned into an anchor links ID. This function specifically turns text placed within the `OSCALAnchorLinkHeader`, which is `children` and typically a header, into the ID, by ensuring all text is lowercased and is hyphen separated. I noticed a couple places where `children.props.children` were best to handle some of the headers. Perhaps this function could be renamed to "conformLinkIdText"?

src/components/OSCALCatalogGroups.js
kylelaker: My concern here is that while `group.title` _is_ required, `group.id` is not. Do we still provide enough information to be useful when the ID is not provided?

Bronstrom: Looking back at the spec now, I'm realizing `group.title` is the only required field (`[1]`) underneath `group`. Ideally `group.id` will be handled here, but I think the `group.title` will be an okay substitute when the `group.id` isn't provided. I will have to tweak this case at the `OSCALCatlogGroups` level in order to work properly with fragments.

src/components/OSCALLoader.js
Bronstrom: In an attempt to get rid of the linter warnings when using `window.location.hash` to aquire the hash over `useLocation()`, I had tried `useCallback()` but I remember receiving a warning saying I needed to use the function in conjunction with a `useEffect()`. I don't believe `useCallback()` is necessary with this problem solved, so I'll just shift it back to the way it was before.

src/components/OSCALCatalogGroups.js
Bronstrom: This should now function as stated above. If `group.id` is not provided, `group.title will` now act as an identifier (and is formatted with hyphens and lowercased). For example, say the Awareness and Training control group for a SP500-83 catalog doesn't have a group.id: instead it's title (Awareness and Training) is used and the fragment becomes `awareness-and-training/at-2/at-2.1` when clicking the "at-2.1" sub-control anchor link.

src/components/OSCALCatalogGroups.js
Bronstrom: A lot of the section has been changed in order to support fragment changes to how the group is captured. Some of this may look similar such as grabbing an element and opening a tab however these involve different states and require a different method of finding the controls. I've done a little bit of refactoring here and made a reusable function for fragment validation.

src/components/OSCALMetadata.js
```diff
@@ -597,11 +607,29 @@ export function OSCALMetadataLocation(props) {
...
-  const { title, children, summary } = props;
+  const { title, children, summary, urlFragment } = props;
+
+  const [isExpanded, setIsExpanded] = React.useState(false);
+
+  const handleExpand = () => {
+    setIsExpanded(!isExpanded);
+  };
+
+  useEffect(() => {
```
tuckerzp, line 618: I am not how much it adds to link the individual metadata segments. It feels a bit excessive. Maybe just do `Metadata`? Or even none of it since it is always the top of a page. 

Regardless of whether or not we keep it, this needs to work and probably have tests added if they haven't been already.

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -58,3 +58,76 @@ export function guessExtensionFromHref(url) {
...
+/**
+ * Finds if a fragment contains a control group by searching the control tabs
+ *
+ * @param {string} fragment A given fragment
+ * @returns {string} The control group id
+ */
+export function determineControlGroupFromFragment(fragment) {
+  if (fragment === null || fragment === "" || fragment === undefined) {
+    return null;
+  }
```
tuckerzp, line 71: Can we not just do
```suggestion
  if (!fragment) {
    return null;
  }
```

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -58,3 +58,76 @@ export function guessExtensionFromHref(url) {
...
+ * Add (push) a control/sub-control to fragmentPrefix.
+ *
+ * @param {string} fragmentPrefix The beginning of a fragment
+ * @param {string} controlId The identification for a control
+ * @returns {string} fragmentPrefix with an added control
+ */
+export function appendToFragmentPrefix(fragmentPrefix, controlId) {
+  return fragmentPrefix && fragmentPrefix !== ""
+    ? `${fragmentPrefix}/${controlId}`
+    : controlId;
```
tuckerzp, line 96: Same thing here. I am pretty sure `""` is falsy. 

```suggestion
  return fragmentPrefix
    ? `${fragmentPrefix}/${controlId}`
    : controlId;
```

src/components/oscal-utils/OSCALLinkUtils.js
```diff
@@ -58,3 +58,76 @@ export function guessExtensionFromHref(url) {
...
+
+/**
+ * Validate a fragment.
+ *
+ * @param {string} fragment A given fragment
+ * @returns {boolean} State of it being a valid fragment
+ */
+export function isValidFragment(fragment) {
+  return fragment && fragment !== "";
+}
```
tuckerzp, line 133: This seems to just be checking that a fragment isn't falsy right? I think this function can either be made more generic, or removed for a simple boolean check.

tuckerzp (commented):
Just gave this a quick glance over.  This is looking really good and I appreciate your hard work! I do have a few comments. 

I noticed a lot of places where we have created functions just checking that something is/ is not `null` or `""`. From my understanding, this can be done just with a boolean check as both values are falsy. I think it would clean up a bit of the indirection we have happening here. 

It is a bit late for it now, but I think more conversation on what is meaningful to add a anchor link to would have been useful. I don't want to add a link to everything just because we can.

src/components/oscal-utils/OSCALLinkUtils.js
kylelaker: Yep -- I think `!!fragment` would accomplish the same goal if we need boolean coercion.

src/components/OSCALCatalogGroups.js
```diff
@@ -52,32 +56,73 @@ const ComponentTab = styled(Tab)(({ theme }) => ({
...
   };
 }
 
 TabPanel.propTypes = {
-  index: PropTypes.number.isRequired,
-  value: PropTypes.number.isRequired,
+  value: PropTypes.string.isRequired,
 };
 
+function determineControlExists(groups, controlLayers, rootLayer) {
```
tuckerzp, line 70: Can we add some js doc to this function? It seems to do a lot and admittedly it is not super obvious what it means to determine a control exists. 

Based on the naming, it seems like the return should be a boolean but it returns what I am guessing is the top level control?

src/components/OSCALMetadata.js
```diff
@@ -597,11 +607,29 @@ export function OSCALMetadataLocation(props) {
...
 
 function OSCALMetadataFieldArea(props) {
-  const { title, children, summary } = props;
+  const { title, children, summary, urlFragment } = props;
+
+  const [isExpanded, setIsExpanded] = React.useState(false);
+
+  const handleExpand = () => {
+    setIsExpanded(!isExpanded);
+  };
```
kylelaker, line 616: This seems a bit dangerous. Why doesn't this function like others where we explicitly have different handlers to open/close? Additonally, we probably want to instead do:

```suggestion
  const handleExpand = () => {
    setIsExpanded((isCurrentlyExpanded) => !isCurrentlyExpanded);
  };
```

https://react.dev/reference/react/useState#setstate-parameters

src/components/oscal-utils/OSCALLinkUtils.js
Bronstrom: You're right on `!fragment` handling both `fragment === null ` & `fragment === undefined`. I've tested out all cases where I do any empty string validation and replacing it as so, however the application doesn't function the correct way. I've been looking into this and it appears that empty strings are handled differently than `null` and `undefined`. In order to truly validate an empty string, we will need to check a variable with the equality operator like `fragment === ""` or `fragment.length === 0`. This implementation and other areas of the code base uses empty strings, which makes these statements needed here.

src/components/OSCALMetadata.js
tuckerzp: > All metadata content functions currently this way

I understand _how_ it works, I just don't know if it is useful to link to something that:

1. Is at the top of the page
2. Something I don't imagine wanting to link to a ton

I mostly want to avoid using anchor links just because we can. I am curious what others think about this. 

> All anchor links tests will be handled in Cypress, as clicking and handling the url is required for testing.

Yeah exactly. Has this been done yet?

src/components/oscal-utils/OSCALLinkUtils.js
tuckerzp: > empty strings are handled differently than `null` and `undefined`

Can you provide more details here? I think something else may be going on. 

```bash
[~]$ node
Welcome to Node.js v19.8.1.
Type ".help" for more information.
> !""
true
> !undefined
true
> !null
true
> 

// Setting fragment to an empty string
> fragment = ""
''
> !fragment
true
> 
```

src/components/OSCALMetadata.js
Bronstrom:
```suggestion
  useEffect(() => {
```

> 1. Is at the top of the page

I get that, although this will send the user exactly to the write place and opens the content for them. I think the less clicks the better in my opinion. In the case of someone directing them to a section.

> 2. Something I don't imagine wanting to link to a ton

I imagine in the future we will be linking to specific parties, and roles, and other pieces of metadata (as these lists will be massive), but right now this doesn't seem all to necessary as we don't have any sections with exhaustive lists. This handles the state needed to get this started.

src/components/OSCALMetadata.js
Bronstrom: I based this implementation off of `OSCALCatalogGroup`'s`CollapsibleListItem`'s `handleClick()` which only has one handler as well. I believe the reason for this is due to the fact that [accordian opening/closing ("expanding/collapsing")](https://mui.com/material-ui/api/accordion/#props:~:text=increase%20of%20height.-,expanded,state%20of%20the%20accordion.,-square) is handled pretty much the same way (clicking the title to expand/collapse).

src/components/oscal-utils/OSCALLinkUtils.js
Bronstrom: Nevermind, I'm wrong. I think I was placing a `!` or two in front of `fragment` and that messed with the state of some things.

I also got stuck on some Stackover comments and short articles speaking in terms of `null`, `undetermined` and empty string, only can you use `myVar === ""` or `myVar.length === 0` for empty string validation - `null` and `undefined` doesn't work cannot be validated by these equality operator methods.

kylelaker (approved):
I am content with the general format and structure of this. My approval is based on the assumption that the remaining comments are addressed and that we get that small test failure cleaned up.

Thank you for your work on this. @tuckerzp please keep an eye on this and provide the final review.

tuckerzp (approved):

Bronstrom:
Multi-level groups are now supported! We may need to do some tweaking to this later, once we have some real content to some more fleshed out test data, but from the test data I was able to add into the content of our editor this works properly.

Bronstrom:
> Multi-level groups are now supported! We may need to do some tweaking to this later, once we have some real test data content with fleshed out groupings, but from the test data I was able to input into the content of our editor this works properly.

It was decided we would do some more stress testing on this before this change is merged.

kylelaker (approved):
I am going to move forward with this to reduce the number of times we have to resolve merge conflicts because of #722. Thank you @Bronstrom for your work.

We expect two follow up PRs: one to handle back matter resource links and another to fix up some of the support for layers of groups.

PR #602, opened 2022-09-02
Improve Catalog Control Group Navigation

### LINKED ISSUE #88: Improved Catalog Control Group Navigation

tuckerzp:
This uses `Tabs` from material ui to make a vertical menu to navigate catalog control groups. It also extracts sub groups to its own function.

closes #88


src/components/OSCALControlPart.js
```diff
@@ -92,7 +92,7 @@ export default function OSCALControlPart(props) {
             componentId={props.componentId}
             modificationAlters={props.modificationAlters}
             modificationSetParameters={props.modificationSetParameters}
-            key={part.id}
+            key={part.id ? part.id : part.name}
```
kylelaker, line 95:
```suggestion
            key={part.id ?? part.name}
```

Or we should just always use the `name`? It's required by the spec while `id` is optional

src/components/OSCALCatalogGroup.js
```diff
@@ -18,57 +18,54 @@ export const OSCALControlList = styled(List)`
...
+
+export default function OSCALCatalogGroup(props) {
+  return (
+    <div>
+      <Box display="flex" justifyContent="flex-end">
+        <OSCALControlParamLegend />
+      </Box>
+      <OSCALControlList>
+        {props.group.groups?.map((innerGroup) => (
+          <OSCALCatalogGroup group={innerGroup} key={innerGroup.id} />
```
kylelaker, line 63: `id` may not be present on `group`s. That's what the `groupKey` function was for. We could just use `group.title`. But `groupKey` gave us a shorter number instead of a super long string.

src/components/OSCALCatalogGroups.js
```diff
@@ -0,0 +1,91 @@
...
+      role="tabpanel"
+      hidden={value !== index}
+      id={`vertical-tabpanel-${index}`}
+      aria-labelledby={`vertical-tab-${index}`}
+      {...other}
+    >
+      {value === index && <Box sx={{ p: 3 }}>{children}</Box>}
+    </div>
+  );
+}
```
kylelaker, line 31: Why not just use https://mui.com/material-ui/api/tab-panel/?

src/components/OSCALCatalogGroups.js
```diff
@@ -0,0 +1,91 @@
...
+  const [value, setValue] = React.useState(0);
+
+  const handleChange = (event, newValue) => {
+    setValue(newValue);
+  };
+
+  return (
+    <OSCALSection>
+      <OSCALSectionHeader>Control Groups</OSCALSectionHeader>
+      <Paper sx={{ flexGrow: 1, bgcolor: "background.paper", display: "flex" }}>
```
kylelaker, line 68:
```suggestion
      <Paper sx={{ flexGrow: 1, display: "flex" }}>
```

Do we really need to set the background color?

kylelaker (commented):
I think we're pretty significantly missing a design for groups in groups in groups. And we're possibly changing the behavior of a control in a control.

The names of these components are also starting to imply that maybe we're losing a little insight into the spec. Remember that it's also possible to have controls without groups (in additions to just layers of groups all the way down).

src/components/OSCALCatalogGroups.js
tuckerzp: Most of the examples https://mui.com/material-ui/react-tabs/ do not use that `tab-panel`. They always use a function like that. I can look into using the already made component though.

tuckerzp:
> I think we're pretty significantly missing a design for groups in groups in groups

I do not think we have ever tested that so I created https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/6e72780dcb869bf9d6ee6774bb6634a6c4b15635/catalogs/NIST_SP-800-53_rev4_catalog.json. Things do get very broken. Although I also do not like how the `OSCALControlParamLegend` is displayed for each group on our current viewer. I will rework this to accommodate multiple groups.

src/components/OSCALControlPart.js
tuckerzp: It does not seem that the `rev4 catalog` actually follows the rule that names are unique. At the very least just using the name causes a key warning.

src/components/OSCALCatalogGroup.js
```diff
@@ -18,57 +18,98 @@ export const OSCALControlList = styled(List)`
...
+  const { control } = props;
+
+  return (
+    <CollapseableListItem
+      itemText={`${control.id.toUpperCase()} ${control.title}`}
+    >
+      <OSCALControl
+        control={control}
+        childLevel={0}
+        key={`control-${control.id}`}
```
kylelaker, line 56: Why not just?
```suggestion
        key={control.id}
```

src/components/OSCALCatalogGroups.js
```diff
@@ -0,0 +1,96 @@
...
+                    key={group.title}
+                    label={group.title}
+                    {...a11yProps(index)}
+                  />
+                ))}
+              </ComponentTabs>
+            </Grid>
+            <Grid item sm={10}>
+              {props.groups?.map((group, index) => (
+                <TabPanel key={group.title} value={value} index={index}>
```
kylelaker, line 86: This is a place where we use `group.title` for key. Let's be consistent around the `key` we use for groups

src/components/OSCALCatalogGroups.js
```diff
@@ -0,0 +1,102 @@
...
+                orientation="vertical"
+                variant="scrollable"
+                value={value}
+              >
+                {props.groups?.map((group, index) => (
+                  <Tab
+                    key={group.title}
+                    label={group.title}
+                    {...a11yProps(index, group.title)}
+                  />
```
kylelaker, line 86: Can you try to follow https://mui.com/material-ui/react-tabs/#customization to make it more like the appearance in the ticket. Specifically, I think I'd like to try to:

 1. Set `text-transform: none` so that we don't uppercase everything
 2. Try to use slightly better color choices
 3. Try and to `text-alight: left` or something else reasonable (this may require padding/margin)

src/components/OSCALCatalogGroups.js
```diff
@@ -0,0 +1,102 @@
...
+            <Grid item sm={2}>
+              <ComponentTabs
+                onChange={handleChange}
+                orientation="vertical"
+                variant="scrollable"
+                value={value}
+              >
+                {props.groups?.map((group, index) => (
+                  <Tab
+                    key={group.title}
```
kylelaker, line 83: Here's another `key={group.title}` without using `groupKey`

kylelaker (commented):
Really good progress here! Thanks! I think there are just a few more things I'd like to try to implement before merging this.

src/components/OSCALControlPart.js
kylelaker: @brian-easyd Can you clarify whether the intent of objects of type `token` (such as [Catalog > Control > Part > name](https://pages.nist.gov/OSCAL/reference/latest/catalog/json-reference/#/catalog/controls/parts/name)) are intended to be unique within a document or within a more limited scope?

kylelaker:
@tuckerzp Do you think that you can provide a summary of the status of this PR? Is it ready for review again or are there additional things that need to be completed?

hreineck (dismissed):
Thanks for your work on this! Just tested it locally and the new UI works great.

src/components/OSCALCatalogGroup.js
```diff
@@ -1,74 +1,118 @@
...
+  if (group.id) {
+    return group.id;
+  }
+  let hash = 7;
+  // eslint-disable-next-line no-restricted-syntax
+  for (const char of group.title) {
+    hash = 31 * hash + char.charCodeAt(0);
+  }
+  return hash;
+}
```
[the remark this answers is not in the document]
tuckerzp: Title is very long and this gives a smaller key based on the title. The React [documentation](https://reactjs.org/docs/lists-and-keys.html) recommends to never use index as a key.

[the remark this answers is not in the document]
tuckerzp: I will switch it to just use `title`. I do not think it matters too much if key is long after thinking about it more.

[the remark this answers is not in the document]
tuckerzp: It seems `groupKey()` was added [here](https://github.com/EasyDynamics/oscal-react-library/commit/2418ffdd340f09f87d0265116aacfd1be5e491f4). I agree it may just be unneeded complication. I assume that our examples may not have always have unique `titles`.

tuckerzp:
@kylelaker 
> provide a summary of the status of this PR?

I think functionality wise, everything is good to go. It is mostly stylistic issues brought up I need to address. 

> Try and to text-alight: left or something else reasonable (this may require padding/margin)

I know that the issue has this happening, but I am now not sure if that would be best. I just realized that after clicking on a tab and then using the up/down arrow keys, the currently selected box gets a really neat highlight feature. It is highlighted on the center of the tab, so I feel like having the text centered makes sense. 

> Any reason we can't use a counter? Or set the key to equal the title itself?

I got rid of `groupKey` completely.

> Have line in the middle go all the way down

The issue I am having here is that sometimes the tabs are longer and sometimes the content is longer. Currently the line is a `borderRight` of the tabs. I tried adding a `borderLeft` and it does fix the issue when the content is longer. The issue is that these two lines render slightly differently, meaning I cant use both.

nuttercd:
@tuckerzp This is really good work! I think the currently selected box highlighting feature is very cool.

nuttercd (approved):

zclarkEDC (approved):

PR #616, opened 2022-09-16
Rework profile catalog inheritance

### LINKED ISSUE #614: Clean up code in SSP Profiles/Catalog Inheritance component
`OSCALProfileCatalogInheritance` has a few places where code can either be cleaned up or updated to better match how we have dealt with similar issues in the codebase. 

### Places To Consider
- The `createTree` function adds unneeded complexity to `map` the profile/catalog inheritance into a `TreeView`. Instead a simple `map` function can accomplish the same thing in a way more consistent with how we have done similar things in the past. 

https://github.com/EasyDynamics/oscal-react-library/blob/b6e0c360533308a252a35c94513f91a625876e59/src/components/OSCALProfileCatalogInheritance.js#L21

- `TreeView` should not need to be in a list. We also usually use `OSCALSectionHeader` instead of `ListSubheader` for the section header. 

https://github.com/EasyDynamics/oscal-react-library/blob/b6e0c360533308a252a35c94513f91a625876e59/src/components/OSCALProfileCatalogInheritance.js#L62-L64

- The parent `Grid` component does not seem like it is needed.

https://github.com/EasyDynamics/oscal-react-library/blob/b6e0c360533308a252a35c94513f91a625876e59/src/components/OSCALProfileCatalogInheritance.js#L58

### Acceptance Criteria 
- Areas of code listed above are refactored as well as any other portions of the component that could be reworked. 
- `OSCALProfileCatalogInheritance.test.js` is updated to fit `describe/test` paradigm. 

### Implementation Notes
- [x] I plan to implement this myself.

tuckerzp:
This improves and simplifies the structure of the
`OSCALProfileCatalogInheritance` component. The `TreeItem`s are simplified
to leverage a recursive structure and to work generically between document
types that can import one another. Additionally, an emphasis is now placed
on the document _title_ instead of the document _type_ with the type moving
to a pill/chip component after the title (so the information is not lost).

Tests are added and improved to ensure that this remains working in the
future as well.

Closes #614

### DIFF
--- a/src/components/OSCALProfileCatalogInheritance.js
+++ b/src/components/OSCALProfileCatalogInheritance.js
@@ -1,86 +1,43 @@
-import React, { useState } from "react";
-import { styled } from "@mui/material/styles";
-import { TreeItem, TreeView } from "@mui/lab";
+import React from "react";
+import ChevronRightIcon from "@mui/icons-material/ChevronRight";
 import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
-import ExpandLessIcon from "@mui/icons-material/ExpandLess";
-import { Grid, IconButton, List, Paper, ListSubheader } from "@mui/material";
-
-const CatalogInheritancePaper = styled(Paper)(
-  ({ theme }) => `
-  margin-top: ${theme.spacing(2)};
-  margin-bottom: ${theme.spacing(2)};
-  position: relative;
-  overflow: auto;
-`
-);
-
-function generateLabel(title, type) {
-  return type === "profile" ? `Profile: ${title}` : `Catalog: ${title}`;
-}
-
-function createTree(tree, ids) {
-  if (tree.inherited == null) {
-    return null;
-  }
-
-  const children = [];
-
-  tree.inherited.forEach((childProfileOrCatalog) => {
-    ids.push(`${childProfileOrCatalog.uuid} #${ids.length + 1}`);
-    children.push(
-      <TreeItem
-        nodeId={`${childProfileOrCatalog.uuid} #${ids.length}`}
-        key={`${childProfileOrCatalog.uuid} #${ids.length}`}
-        label={generateLabel(
-          childProfileOrCatalog.title,
-          childProfileOrCatalog.type
-        )}
-      >
-        {createTree(childProfileOrCatalog, ids)}
-      </TreeItem>
-    );
-  });
-
-  return children;
-}
+import { TreeItem, TreeView } from "@mui/lab";
+import Card from "@mui/material/Card";
+import Stack from "@mui/material/Stack";
+import Chip from "@mui/material/Chip";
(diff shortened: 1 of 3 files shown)


src/components/OSCALProfileCatalogInheritance.js
```diff
@@ -1,86 +1,53 @@
...
+    return (
       <TreeItem
-        nodeId={`${childProfileOrCatalog.uuid} #${ids.length}`}
-        key={`${childProfileOrCatalog.uuid} #${ids.length}`}
-        label={generateLabel(
-          childProfileOrCatalog.title,
-          childProfileOrCatalog.type
-        )}
+        key={item.uuid}
+        nodeId={count.toString()}
```
kylelaker, line 27: Out of curiosity, can we just use the `item.uuid` here? Does `nodeId` need to be an integer-ish thing? Would be nice to pull `count` out of here.

kylelaker (commented):
havent looked in depth but i saw this and wanted to ask

src/components/OSCALProfileCatalogInheritance.js
tuckerzp: Good point. That makes a lot more sense.

src/components/OSCALProfileCatalogInheritance.js
```diff
@@ -1,86 +1,49 @@
...
-export default function OSCALProfileCatalogInheritance(props) {
-  const [expandedIds, setExpandedIds] = useState([]);
-  const ids = [];
-
-  const children = createTree(props.inheritedProfilesAndCatalogs, ids);
-
-  if (ids.length > 0 && expandedIds.length === 0) {
-    setExpandedIds(ids);
-  }
+  };
```
kylelaker, line 29: Okay now I am also kinda curious whether we can do something like:

```
const documentLabel = (
  <Typography>
  {item.title}
  <Chip label={item.type[0].toUpperCase() + item.type.slice(1) />
  </Typography>
);

return (
  <TreeItem
    key={item.uuid}
    nodeId={item.uuid}
    label={documentLabel}
  >
  {item.inherited?.map((inherited) => inheritedListItem(inherited))}
  </TreeItem>
);
```

Since the `children` would either be a list or `undefined` in this case, we don't really need the if statement for that. And the profile vs catalog thing can be solved by just semi-title-casing the document `type`. And moving it to `Chip` at the end of the text at the same time kinda cleans up the display a bit. And I think this reduced some duplication.

kylelaker (commented):
Okay I have one more thought on how to clean this up further.

kylelaker (approved):
This is an amazing improvement! I love how simple it is, the emphasis on the title instead of document type, and the improved tests. Thank you so much!

zclarkEDC (approved):

zclarkEDC (approved):

PR #639, opened 2022-12-08
Resolve dependency failures

Bronstrom:


Bronstrom:
Looking a little deeper into the dependency failures it seems like `eslint` is the culprit of the failures. I believe there may be an issue with the configuration cascading structure causing: ESLint to be unable to "determine the plugin "import" uniquely" due to `eslint-plugin-import` being imported twice, once in `oscal-react-library/node_modules` and secondly in `oscal-react-library/example/node_modules`.

Recent versions of ESLint load all config files (an `.eslintrc…` or `package.json`) starting from low level parent directories and chain up to the top project root directory. With the way our project is configured with `react-libraries`, I think we'll need to limit ESLint, which can be solved be halting it at a parent directory config file with `root: true`.

For more information, check the [eslint configuration-files doc](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#cascading-and-hierarchy).

kylelaker:
Thank you so much for your work on this. Please instead consider #640. That handles moving everything to `.eslintrc.json` as well as removing duplicate `plugins` configs from `package.json` and fixing the lint rules that are now enabled as part of this.

PR #640, opened 2022-12-13
Update dependencies and fix linting

kylelaker:
This updates all dependencies. In doing so, it highlighted that our config for
ESLint was weird with duplicate plugins and other issues. And it turns out that
quite a few rules haven't actually been running against our code. In order to
not have a series of commits that are failing on the `develop` branch and to
avoid temporarily adding a bunch of rule overrides, this also works on fixing
those lint rules.

**Please** review the commit that modifies a line to understand why it is being
changed before providing feedback on that change. Additionally, please focus all
efforts on dependency updates on this branch; hopefully the bot can take back over
once this is in.

### DIFF
--- a/src/components/OSCALSystemImplementationComponents.test.js
+++ b/src/components/OSCALSystemImplementationComponents.test.js
@@ -5,26 +5,35 @@ import { componentsTestData } from "../test-data/SystemData";
 import OSCALSystemImplementationComponents from "./OSCALSystemImplementationComponents";
 
 describe("OSCALSystemImplementationComponents", () => {
-  beforeEach(() => {
+  test("shows Components section title", () => {
     render(
       <OSCALSystemImplementationComponents
         components={componentsTestData}
         parties={responsiblePartiesTestData}
       />
     );
-  });
-
-  test("shows Components section title", () => {
     const result = screen.getByText("Components");
     expect(result).toBeVisible();
   });
 
   test(`shows component title`, () => {
+    render(
+      <OSCALSystemImplementationComponents
+        components={componentsTestData}
+        parties={responsiblePartiesTestData}
+      />
+    );
     const result = screen.getByText("Example Component");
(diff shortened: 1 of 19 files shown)


Bronstrom:
In order to resolve "Avoid direct Node access", I've had to rewrite the tests providing this error. There seems to be no way to acquire the parent element from a child element without using native HTML methods (like `closest`), which are not a part of the testing library). Instead, I've written these tests to acquire the parent elements first (the rows) and then the data `within` (`th` and `td` cells).

For the `OSCALSystemImplementation` tests, I don't particularly like how I had to specifying a specific row in an array (created by `getAllByRole`). Although this seems like the best option without having to tweak any of the current code. What would make this search easier, would be to make these implemented component/item rows unique by giving them a label.

src/components/OSCALSystemImplementationComponents.test.js
```diff
@@ -47,8 +47,17 @@ describe("OSCALSystemImplementationComponents", () => {
         parties={responsiblePartiesTestData}
       />
     );
-    const component = screen.getByText("Example Component").closest("tr");
-    const result = within(component).getByText("operational");
+    // Grab example party row beneath heading row
+    const exampleResponsiblePartiesRow = screen.getAllByRole("row")[1];
```
kylelaker, line 51: Hmm `[1]` feels like we're just doing the same thing in a different way. Can you explain why this is correct? I don't think it's wrong; it's just not immediately obvious to me why it's right.

Bronstrom: This is the part I was mentioning I don't like very much.  Following the previous written tests, we're trying to ensure that the example content appears in the same row. There is nothing unique separating the rows that appear on screen, so they need to both be acquired when using the testing library. In this array, the first row is the heading cells, and the second is the example component.

PR #660, opened 2023-02-02
Remove ignore rules from ESLint configuration

kylelaker:
In a recent pull request, we disabled many ESLint rules because they were finally being
detected. This adds corrections for many of them. This is largely a duplicate of #640;
however, in #640 there are some commits that cause rendering loops and this PR does not
include those (we will still need to address them but they can be merged separately).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -31,20 +31,9 @@
       "react/destructuring-assignment": [0],
       "react/jsx-props-no-spreading": "off",
       "prettier/prettier": "error",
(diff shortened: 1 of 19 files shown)


tuckerzp (approved):
Thank you for your work on this! I am happy to see some of these fixes finally getting in.

Bronstrom (approved):
Good to see these linting rules be resolved that we're failing for so long! Hopefully the commits with render loops won't be much of an issue going forward.

Issue #592, opened 2022-08-30
Support OSCAL Markup (Line and Multi-Line) in Hover-Over (Tooltips-like functionality)


tuckerzp:
Currently in places such as the [`Components`](https://github.com/EasyDynamics/oscal-react-library/blob/92b4818000368a707342941ce3752caa9236913b/src/components/OSCALSystemImplementation.js#L69-L73 ) in the `System Implementation`, we do not support `Markdown`. As per the spec, the `component.description` / `Tooltip title` needs to be ` markdown multi-line` & the `child` / `component.title` needs to be `markdown line`. This occurs in multiple places throughout the codebase. 

Trying to use our `MarkdownProse` components inside the `ToolTip` causes console warnings and test failures.

We need a hover-over capability that can handle complex presentations such as markup-multiline. This does not have to be tools tips specifically.

### Acceptance Criteria
- [ ] `Markdown` is supported and works according to the spec
- [ ] A `description` is displayed in a meaningful way. Preferably similar to how the tooltip works today. 
- [ ] This capability should be issued in in one or two places to demo. (Additional implementations are handled outside this issue.)

_Originally posted by @tuckerzp in https://github.com/EasyDynamics/oscal-react-library/pull/590#discussion_r957660892_

brian-easyd:
This task is about creating the capability and implementing it in one or two places to demo. Additional implementations are handled outside this issue.

PR #682, opened 2023-02-27
feat(metadata): Display role description

### LINKED ISSUE #408: Display Metadata Role Descriptions
### Description 
As a user of an OSCAL object, I want a display of the description of the "role" of each party in the metadata section so that I can understand what each party's responsibilities are.

### Acceptance Criteria
- If a party has a role with a description, both the role title and role description are displayed in the metadata React component
- The tests are updated to demonstrate the new functionality

### Implementation Notes
The OSCAL Viewer currently displays the `title` of a given role in `metadata/roles` -- the object may also have a `description` attribute, in which case it should be rendered in addition to the title.

tuckerzp:
`metadata.roles` may contain a description. The description
will now be included when you view the `Contact` information.

To test, load my temporary [example rev 5 catalog](http://localhost:3000/catalog/?url=https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/test-role-description/catalogs/NIST_SP-800-53_rev5_catalog.json). Then click the `Contact` button.

Closes #408

### DIFF
--- a/src/components/OSCALMetadata.js
+++ b/src/components/OSCALMetadata.js
@@ -334,15 +334,41 @@ export function OSCALMetadataParty(props) {
 }
 
 export function OSCALMetadataRole(props) {
+  const [open, setOpen] = React.useState(false);
+
+  const handleOpen = () => {
+    if (props.role.description) {
+      setOpen(true);
+    }
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
   return (
     <Card>
       <CardHeader
         title={props.role.title}
         avatar={<MetadataAvatar text={props.role.title.toUpperCase()} />}
-        subheader={
-          <OSCALMarkupMultiLine>{props.role.description}</OSCALMarkupMultiLine>
-        }
+        onClick={handleOpen}
       />
+      <CardActions>
+        <Dialog
+          open={open}
+          onClose={handleClose}
+          scroll="paper"
+          maxWidth="md"
+          fullWidth
+        >
+          <DialogTitle>{props.role.title}</DialogTitle>
+          <DialogContent dividers>
+            <OSCALMarkupMultiLine>
+              {props.role.description}
+            </OSCALMarkupMultiLine>
+          </DialogContent>
+        </Dialog>
+      </CardActions>
     </Card>
   );
 }
--- a/src/components/OSCALMetadata.js
+++ b/src/components/OSCALMetadata.js
@@ -353,6 +353,7 @@ export function OSCALMetadataRole(props) {
           variant="outlined"
           onClick={handleOpen}
           aria-label={`${props.role.title} details button`}
+          disabled={!props.role?.description}
         >
           <InfoIcon />
           Details
--- a/src/components/OSCALMetadata.js
+++ b/src/components/OSCALMetadata.js
@@ -200,18 +200,7 @@ export function OSCALMetadataPartyDialog(props) {
           <Stack direction="column">
             {props.party.name}
             {props.partyRolesText?.map((role) => (
-              <StyledTooltip
-                title={
-                  role.description ? (
-                    <Typography> {role.description} </Typography>
-                  ) : (
-                    ""
-                  )
-                }
-                key={role.title}
-              >
-                <Typography> {role.title} </Typography>
-              </StyledTooltip>
+              <Typography key={role.title}> {role.title} </Typography>
             ))}
           </Stack>
         </Stack>
--- a/src/components/OSCALMetadata.js
+++ b/src/components/OSCALMetadata.js
@@ -1,12 +1,10 @@
 import BusinessIcon from "@mui/icons-material/Business";
-import ContactPageIcon from "@mui/icons-material/ContactPage";
 import EmailIcon from "@mui/icons-material/Email";
 import HelpOutlineIcon from "@mui/icons-material/HelpOutline";
 import HomeIcon from "@mui/icons-material/Home";
 import MapIcon from "@mui/icons-material/Map";
 import PhoneIcon from "@mui/icons-material/Phone";
 import SmartphoneIcon from "@mui/icons-material/Smartphone";
-import { CardActionArea, CardContent } from "@mui/material";
 import Avatar from "@mui/material/Avatar";
 import Button from "@mui/material/Button";
 import Card from "@mui/material/Card";
@@ -26,6 +24,7 @@ import Stack from "@mui/material/Stack";
 import { styled } from "@mui/material/styles";
 import Typography from "@mui/material/Typography";
 import React from "react";
+import InfoIcon from "@mui/icons-material/Info";
 import OSCALEditableTextField from "./OSCALEditableTextField";
 import { OSCALMarkupMultiLine } from "./OSCALMarkupProse";
 
@@ -316,9 +315,9 @@ export function OSCALMetadataParty(props) {
           size="small"
           variant="outlined"
           onClick={handleOpen}
-          aria-label={`${props.party.name} contact button`}
+          aria-label={`${props.party.name} details button`}
         >
-          <ContactPageIcon />
+          <InfoIcon />
           Details
         </Button>
         <OSCALMetadataPartyDialog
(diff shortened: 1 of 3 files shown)


tuckerzp:
*Copied from teams*

This design (seen below) came from a few decisions/thoughts:
- I did not believe the `MetaDataParties` box really needed to be cluttered with descriptions. I think it looks really clean and should stay that way. Especially given we do not use description in any of our OSCAL data today, I decided it was not something that needs to be displayed prominently.
- I felt the next place you'd look for the description is in the `Contact` section. I usually think of getting more information when I click on something like that, so it felt natural to put the information there. 
- The current way we are displaying roles is just taking each role and making a comma separated list of titles. I think the role still fits at the top of the contact card. If it has a description, I thought the best way to display it would be `title - description` or something similar on each line. Further needing to click seemed clunky to me. Plus I was unsure how long a description could possibly be, so I wanted to have a fairly flexible way to deal with various lengths. I also made my own custom catalog to test at https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/test-role-description/catalogs/NIST_SP-800-53_rev5_catalog.json

image

kylelaker:
Because we're now displaying more than just the basic summary + contact info (we have basic summary + role details + contact info), I wonder if we should change the "Contact" button to "Details" (and also maybe make the entire card clickable)? Also, what was the reason for choosing `-` over enclosing in parenthesis or using a tooltip?

Some more thoughts:

1. Should we display _all_ roles with their title and description somewhere in a table? Roles are used outside of just Party info I believe; but even if I am mistaken and they aren't, we likely want to view all the Roles defined somewhere.
2. If we do that, then I definitely think this should be a tooltip/hover over the title to show the description (and you can do more lookups in the table later)
3. The comma separation and `-` delimiting feels naive and problematic. Are these characters disallowed in the descriptions? This becomes confusing if my `description` has a `-` and `,` and text that could be interpreted as the start of a new role title & description.
4. `description` is a `markup-multiline` field. How do you handle the markup here?

Bronstrom:
I do like the choice of using "Details" over "Contact" and tooltip ideas as suggested by @kylelaker. "Details" seem to more encapsulate the idea of "More Information". Most of the time I think of "Contact" bringing you directly to a contact page that allows you to directly send a message to an inbox. I can kinda see why you made the decision to use `-`, specifically in this case, as it seems like the description is defining the title. Although, it may be best to standardize to something other than this or `,` (which was used previously) to avoid cases where these are used in the title or description, also suggested by @kylelaker. I'm leaning most towards tooltip to avoid any cases of symbols being used.

I think you are making a good choice avoiding the unnecessary details at the general document level, and including the title description only in the pop-up window. I also have one other nitpicky detail, I would suggest making the title and description that grayish, subtext color used for the Parties title on the main document page. I think this creates better separation from the party heading.

src/components/OSCALMetadata.js
```diff
@@ -300,7 +307,7 @@ export function OSCALMetadataParty(props) {
       <CardHeader
         avatar={avatar}
         title={props.party.name}
-        subheader={props.partyRolesText}
+        subheader={props.partyRolesText.map((role) => role.title).join(", ")}
```
kylelaker, line 310:
```suggestion
        subheader={props.partyRolesText?.map((role) => role.title).join(", ")}
```

src/components/OSCALMetadata.js
```diff
@@ -197,7 +198,13 @@ export function OSCALMetadataPartyDialog(props) {
           {props.avatar}
           <Stack direction="column">
             {props.party.name}
-            <Typography variant="body2">{props.partyRolesText}</Typography>
+            {props.partyRolesText.map((role) => (
```
kylelaker, line 201:
```suggestion
            {props.partyRolesText?.map((role) => (
```

kylelaker (commented):
This should fix the failing test

tuckerzp:
> I'm leaning most towards tooltip to avoid any cases of symbols being used.

> description is a markup-multiline field. How do you handle the markup here?

I will push changes soon, but I feel like the tool tip looks really nice here. However, there is the issue of the `ToolTip` not being able to support our `MarkupProse` components https://github.com/EasyDynamics/oscal-react-library/issues/592. 

I think initially just using a `Typography` could be good here as placing `MarkupProse` components in tool tips is a common issue we will need to address. 

> Should we display all roles with their title and description somewhere in a table? Roles are used outside of just Party info I believe; but even if I am mistaken and they aren't, we likely want to view all the Roles defined somewhere.

Maybe there could be another section that looks similar to the `Parties` just without the extra `Details` option here? You would just hover over the card to get the details tool tip.

tuckerzp:
image

tuckerzp:
Taking this out of draft for now since the basic design is there.

tuckerzp:
> Should we display all roles with their title and description somewhere in a table? Roles are used outside of just Party info I believe; but even if I am mistaken and they aren't, we likely want to view all the Roles defined somewhere

image

tuckerzp:
I almost wonder if we should just drop the `tooltip`? I really like how just having a roles Card looks here. What does everyone think?

Bronstrom:
Oh wow, this is a whole redesign. I can see benefits to both approaches:
1) Tooltip approach: This nicely links the description and other role data when opening a party. If want more information you can hover over the role.
2) Separate Roles Grouping: This in my opinion seems more like a "legend" system, where you match up the party's role(s) with the "Roles" section.

At this point, I would almost say to support both.

I do have a couple of design nitpicks to the new approach. From my understanding it seems like "Parties" should be of greater focus over "Roles", but currently it seems like "Roles" and "Parties" are at the same level in the hierarchy, albeit the "Details" button. Roles following the same design choices as Parties (Logo to right, title on top and description beneath title). Here is a suggestion that could a little bit of a stretch to implement: Instead of having the logo, maybe we choose to use that color in the logo for the font color of the title and then we use the same font color for the roles color underneath each party. To me this seems to follow more of the "legend" approach it is going for, but also create more hierarchy between "Parties" from "Roles".

tuckerzp:
> Tooltip approach: This nicely links the description and other role data when opening a party. If want more information you can hover over the role.

The issue with that is `ToolTips` do not support our `MarkupProse`

Bronstrom:
Also, here is an up-to-date screenshot of metadata current design using Zach's catalog:

image

Bronstrom:
> The issue with that is `ToolTips` do not support our `MarkupProse`

Okay gotcha, that makes more sense for the redesign. I thought you had worked something out based on one of the previous screenshots.

kylelaker:
@tuckerzp is there any way we can get the cards to be approximately the same height? I don't love that their sizes are so different based on how much description there is.

tuckerzp:
> is there any way we can get the cards to be approximately the same height? I don't love that their sizes are so different based on how much description there is.

I feel like using a dialog box helps address both of your comments. The Role has a logo partly because I like the logo and partly because the `CardHeader` does weird things with just a `title` set. 

---
image

image

tuckerzp:
@kylelaker @Bronstrom I am going to take a break for the night but I think things are getting close.

image

I am now using a `CardActionArea` to open a `Dialog` when a role has a description. The actual content is just a `Typography` that we can edit to look nicer. I am having a couple issues that I will look into but would appreciate any help you can provide after looking over the code.

1. Using the `CardActionArea` makes the card *slightly* larger. This is extremely frustrating and I am unsure how to fix it. I will have to look into that, but do realize it is terrible.
2. Unlike the Party Dialog, clicking on the screen outside the Role dialog does not count as closing... I am confused on why that is the case as the two are pretty similar. I am wondering if not having a button is making my life more difficult, but I feel the button would be overkill.

kylelaker:
> 2\. Unlike the Party Dialog, clicking on the screen outside the Role dialog does not count as closing... I am confused on why that is the case as the two are pretty similar. I am wondering if not having a button is making my life more difficult, but I feel the button would be overkill.

Because 'Details' is the only action for Party, I think we should use the same UX "idiom" to open the details modal for both (so either button for both or make the whole thing clickable for both). This should also help with the visual consistency.

tuckerzp:
Kyle's suggestion helped me address both of my issues! I _think_ this design is at a finished state... Let me know what you all think. 


image

---

image

---

image

src/components/OSCALMetadata.js
```diff
@@ -325,12 +333,55 @@ export function OSCALMetadataParty(props) {
...
+  return (
+    <Card>
+      <CardHeader subheader={props.role.title} />
+      <CardActions>
+        <Button
+          size="small"
+          variant="outlined"
+          onClick={handleOpen}
+          aria-label={`${props.role.title} details button`}
+        >
```
kylelaker, line 356:
```suggestion
        <Button
          size="small"
          variant="outlined"
          onClick={handleOpen}
          aria-label={`${props.role.title} details button`}
          disabled={!props.role?.description}
        >
```

Maybe we can just not allow opening the dialog if there isn't a description? Seems silly to get a confusing modal without text

Bronstrom:
> I think we have more duplication between the parties/roles than I'd really like

I feel the same way about this, but I do think this has improved quite a bit from the original design and there is some distinction from Parties and Roles. I think disabling the "Details" button when the description isn't present works fine for now and better than having the card height differences. I definitely think you have satisfied the requirements for this issue.

Although, should we add tests for this issue?

Bronstrom (approved):
> Although, should we add tests for this issue?

Nevermind I missed them on my first lookup

PR #720, opened 2023-03-07
fix(catalog): Withdrawn "top" controls display correctly

### LINKED ISSUE #635: Withdrawn "top-level" controls are no longer identifiable
### Description

Previously, withdrawn controls in SP 800-53r5 were shown with a strikethrough (or maybe it was just lighter text?). After #602, they are no longer shown that way.

This impacts controls like AC-15 but not others, like AC-2.10.

### Environment

https://oscal-viewer.msd.easydynamics.com/

### Steps to Reproduce

1. Open https://oscal-viewer.msd.easydynamics.com/catalog?url=https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/main/catalogs/NIST_SP-800-53_rev5_catalog.json
2. View a control that was been withdrawn in Rev 5 (AC-15)

### Expected behavior

The item should not expand if there isn't any text. Withdrawn controls should have a unique appearance indicating their status.

### Actual Behavior

The title is shown and it expands; however there is neither any text or indication that it has been withdrawn

### Additional Notes

_Note_: I am not sure whether `{ "name": "status", "value": "withdrawn" }` is part of the specification or just a convention; on https://pages.nist.gov/OSCAL/reference/latest/catalog/json-reference/#/catalog/controls/props the only defined `name` is `marking` and I do not see any definitions for status at https://github.com/search?q=repo%3Ausnistgov%2FOSCAL%20withdrawn&type=code

Bronstrom:
This fixes the "state" of withdrawn "top-level"
controls. Instead of being a collapsible item
that shows a blank box when selected, the
control no longer opens and closes, as
indicated by no presence of an arrow
appearing to the far right of the item. The
item text is also styled to appear a light-gray
and has a strikethrough, like the "sub-level"
controls.

Resolves #635

### DIFF
--- a/src/components/OSCALCatalogGroup.js
+++ b/src/components/OSCALCatalogGroup.js
@@ -8,6 +8,7 @@ import ListItemText from "@mui/material/ListItemText";
 import { styled } from "@mui/material/styles";
 import React from "react";
 import OSCALControl from "./OSCALControl";
+import isWithdrawn from "./oscal-utils/OSCALCatalogUtils";
 
 export const OSCALControlList = styled(List)`
   padding-left: 2em;
@@ -52,11 +53,12 @@ function CollapseableListItem(props) {
 
 function OSCALCatalogControlListItem(props) {
   const { control } = props;
-
-  return (
-    <CollapseableListItem
-      itemText={`${control.id.toUpperCase()} ${control.title}`}
-    >
+  // Determine if control is withdrawn
+  const withdrawn = isWithdrawn(control);
+  const itemText = `${control.id.toUpperCase()} ${control.title}`;
+  // Make collapsible item when not a withdrawn control
+  return !withdrawn ? (
+    <CollapseableListItem itemText={itemText}>
       <OSCALControl
         showInList
         control={control}
@@ -64,6 +66,16 @@ function OSCALCatalogControlListItem(props) {
         key={control.id}
       />
     </CollapseableListItem>
+  ) : (
+    <StyledListItemPaper>
+      <StyledListItem>
+        <ListItemText
+          primary={itemText}
+          withdrawn={withdrawn}
+          style={{ textDecoration: "line-through", color: "#d4d4d4" }}
+        />
+      </StyledListItem>
+    </StyledListItemPaper>
   );
 }
 
--- a/src/components/OSCALCatalogGroup.js
+++ b/src/components/OSCALCatalogGroup.js
@@ -25,7 +25,7 @@ const StyledListItemPaper = styled(Paper)`
   border-radius: 0.5em;
 `;
 
-const StyledListItemText = styled(ListItemText)(({ theme }) => ({
+const WithdrawnListItemText = styled(ListItemText)(({ theme }) => ({
   textDecoration: "line-through",
   color: theme.palette.grey[400],
 }));
@@ -73,7 +73,7 @@ function OSCALCatalogControlListItem(props) {
   ) : (
     <StyledListItemPaper>
       <StyledListItem>
-        <StyledListItemText primary={itemText} withdrawn={withdrawn} />
+        <WithdrawnListItemText primary={itemText} withdrawn={withdrawn} />
       </StyledListItem>
     </StyledListItemPaper>
   );
--- a/src/components/OSCALCatalogGroup.js
+++ b/src/components/OSCALCatalogGroup.js
@@ -73,10 +73,7 @@ function OSCALCatalogControlListItem(props) {
   ) : (
     <StyledListItemPaper>
       <StyledListItem>
-        <StyledListItemText
-          primary={itemText}
-          withdrawn={withdrawn}
-        />
+        <StyledListItemText primary={itemText} withdrawn={withdrawn} />
       </StyledListItem>
     </StyledListItemPaper>
   );
--- a/src/components/OSCALCatalogGroup.js
+++ b/src/components/OSCALCatalogGroup.js
@@ -25,6 +25,11 @@ const StyledListItemPaper = styled(Paper)`
   border-radius: 0.5em;
 `;
 
+const StyledListItemText = styled(ListItemText)(({ theme }) => ({
+  textDecoration: "line-through",
+  color: theme.palette.grey[400],
+}));
+
 const StyledControlDescriptionWrapper = styled("div")`
   padding: 1em;
 `;
@@ -53,10 +58,9 @@ function CollapseableListItem(props) {
 
 function OSCALCatalogControlListItem(props) {
   const { control } = props;
-  // Determine if control is withdrawn
   const withdrawn = isWithdrawn(control);
   const itemText = `${control.id.toUpperCase()} ${control.title}`;
-  // Make collapsible item when not a withdrawn control
+
   return !withdrawn ? (
     <CollapseableListItem itemText={itemText}>
       <OSCALControl
@@ -69,10 +73,9 @@ function OSCALCatalogControlListItem(props) {
   ) : (
     <StyledListItemPaper>
       <StyledListItem>
-        <ListItemText
+        <StyledListItemText
           primary={itemText}
           withdrawn={withdrawn}
-          style={{ textDecoration: "line-through", color: "#d4d4d4" }}
         />
       </StyledListItem>
     </StyledListItemPaper>
--- a/src/components/oscal-utils/OSCALCatalogUtils.js
+++ b/src/components/oscal-utils/OSCALCatalogUtils.js
@@ -1,0 +1,15 @@
+/**
+ * Determines if a control is withdrawn.
+ *
+ * TODO: This is probably 800-53 specific and it should be made more
+ * generic to allow the library to work with other frameworks.
+ * https://github.com/EasyDynamics/oscal-react-library/issues/504
+ *
+ * @param {object} control
+ * @returns a boolean describing whether the control is withdrawn
+ */
+export default function isWithdrawn(control) {
+  return control?.props?.find(
+    (prop) => prop.name === "status" && prop.value === "withdrawn"
+  );
+}
--- a/src/components/oscal-utils/OSCALCatalogUtils.js
+++ b/src/components/oscal-utils/OSCALCatalogUtils.js
@@ -3,10 +3,6 @@ import matchingProp from "./OSCALPropUtils";
 /**
  * Determines if a control is withdrawn.
  *
- * TODO: This is probably 800-53 specific and it should be made more
- * generic to allow the library to work with other frameworks.
- * https://github.com/EasyDynamics/oscal-react-library/issues/504
- *
  * @param {object} control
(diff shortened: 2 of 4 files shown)


tuckerzp (commented):
This looks like a pretty good start so far! I had a small comment and a thought.

I will also post this for other's reviewing
image

src/components/oscal-utils/OSCALCatalogUtils.js
```diff
@@ -0,0 +1,15 @@
...
+ * https://github.com/EasyDynamics/oscal-react-library/issues/504
+ *
+ * @param {object} control
+ * @returns a boolean describing whether the control is withdrawn
+ */
+export default function isWithdrawn(control) {
+  return control?.props?.find(
+    (prop) => prop.name === "status" && prop.value === "withdrawn"
+  );
+}
```
kylelaker, line 15: Let's actually do namespace validation. Some of this code probably belongs in a different file but I am thinking of an API a bit like this (ignore the types if the resulting file has to be JS):


```ts
/**
 * When a namespace is not specified for a property, the assumption is that the
 * namespace is default namespace.
 */
const NIST_DEFAULT_NAMESPACE = "https://csrc.nist.gov/ns/oscal";

export interface PropFilter {
  /**
   * The namespace of the property.
   * 
   * @default https://csrc.nist.gov/ns/oscal
   */
  ns?: string;

  /**
   * The name of the prop to find.
   */
  name: string;

  /**
   * The acceptable value of the property.
   * 
   * Either this property or {@link filter} (but not both) must be specified.
   */
  value?: string;

  /**
   * A function to validate the property value.
   * 
   * Either this property or {@link value} (but not both) must be specified.
   */
  filter?: (value: string) => boolean;
}

/**
 * Return the given namespace, or if undefined, the default namespace.
 */
export function namespaceOf(ns: string | undefined): string {
  return ns ?? NIST_DEFAULT_NAMESPACE;
}

/**
 * Returns the first matching property in the given list.
 * 
 * @param props the list of properties to check
 * @param filter the attributes to match against
 */
export function matchingProp(props: Prop[], filter: PropFilter): Prop | undefined {
  const ns = namespaceof(filter.ns);
  return props.find((prop) =>
    namespaceOf(prop.ns) === namespaceOf(filter.namespace)
    && prop.name === filter.name
    && (!filter.value || prop.value === filter.value)
  );
}

export function isWithdrawn(control: Control): boolean {
  // By default controls are *not* withdrawn
  if (!control.props) {
    return false;
  }
  return !!matchingProp(control.props, { name: "status", value: "withdrawn });
}
```

Besides `isWithdrawn`, most of this probably deserves to exist more in an `oscal-utils/OSCALPropUtils.js`.

kylelaker (changes requested):

src/components/oscal-utils/OSCALCatalogUtils.js
kylelaker: (Eventually, we may want to support `class` in that filter interface but I do not want to deal with that today)

kylelaker:
I am also worried that the font color with the background could be an accessibility issue... I wonder if there's something we can do there (or maybe we wait and try to work on a high-contrast theme later?)

src/components/OSCALCatalogGroup.js
```diff
@@ -52,18 +53,29 @@ function CollapseableListItem(props) {
...
   const { control } = props;
-
-  return (
-    <CollapseableListItem
-      itemText={`${control.id.toUpperCase()} ${control.title}`}
-    >
+  // Determine if control is withdrawn
+  const withdrawn = isWithdrawn(control);
+  const itemText = `${control.id.toUpperCase()} ${control.title}`;
+  // Make collapsible item when not a withdrawn control
```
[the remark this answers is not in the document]
Bronstrom: Yeah, I can remove these - the top one I just copied and pasted for the utils function anyway. I had a lot of random thoughts I scrambled to write in the comments the other day. I think I got rid of most of these at the very least 😅

Bronstrom:
> I am also worried that the font color with the background could be an accessibility issue... I wonder if there's something we can do there (or maybe we wait and try to work on a high-contrast theme later?)

Where the other withdrawn controls appear (sub-level controls), a `theme.palette` color wasn't declared there and just a hex code for a gray color. I ended up just matching the top-level controls to this. It would probably be best we have these align to the `theme.palette`. In the past we have just declared something like `theme.palette.grey[50]` to produce darker/lighter shades of palette colors, which could work in this case. I believe "50" aligns to opacity.

src/components/oscal-utils/OSCALCatalogUtils.js
Bronstrom: I've gone ahead and added namespace validation for JS - types have been removed as well as the interface for now. There were a couple other variable corrections I had to make, but I was able to figure them out.

Bronstrom: I've separated `isWithdrawn` from the rest of the functions `oscal-utils/OSCALPropUtils.js` but followed the same logic as mentioned. I've just kept the `isWithdrawn` function in the Catalog Utils class I created. Let me know if something else works better here.

src/components/OSCALCatalogGroup.js
```diff
@@ -24,6 +25,11 @@ const StyledListItemPaper = styled(Paper)`
   border-radius: 0.5em;
 `;
 
+const StyledListItemText = styled(ListItemText)(({ theme }) => ({
```
kylelaker, line 28:
```suggestion
const WithdrawnListItemText = styled(ListItemText)(({ theme }) => ({
```

Let's specify _how_ it's styled and what it actually represents. Imagine some day we want to signify "important" controls (or some other made up descriptor) somehow -- those would be styled too but perhaps in a different way. So `StyledListItemText` doesn't really get the point across in the same way.

I know we've made this mistake elsewhere, but let's do better here.

src/components/oscal-utils/OSCALCatalogUtils.js
```diff
@@ -0,0 +1,19 @@
+import matchingProp from "./OSCALPropUtils";
+
+/**
+ * Determines if a control is withdrawn.
+ *
+ * TODO: This is probably 800-53 specific and it should be made more
+ * generic to allow the library to work with other frameworks.
+ * https://github.com/EasyDynamics/oscal-react-library/issues/504
```
kylelaker, line 8: This is now handled. `props[@name="status" and @value="withdrawn"]` is how to do this per the spec and we now handle the namespace field correctly for this.

src/components/oscal-utils/OSCALPropUtils.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+export default function matchingProp(props, filter) {
+  const ns = namespaceOf(filter.ns);
+  if ((!filter.value && !filter.filter) || (filter.value && filter.filter)) {
+    throw new Error("Exactly one of filter or value must be specified");
+  }
+
+  return props.find(
+    (prop) =>
+      namespaceOf(prop.ns) === ns &&
```
kylelaker, line 28:
```suggestion
  if ((!filter.value && !filter.filter) || (filter.value && filter.filter)) {
    throw new Error("Exactly one of filter or value must be specified");
  }

  return props.find(
    (prop) =>
      namespaceOf(prop.ns) === namespaceOf(filter.ns) &&
```

This makes the whole thing more consistent internally and the variable is never used elsewhere.

Bronstrom: I was wondering - that makes more sense now.

src/components/oscal-utils/OSCALPropUtils.js
```diff
@@ -0,0 +1,33 @@
...
+ * namespace is default namespace.
+ */
+const NIST_DEFAULT_NAMESPACE = "https://csrc.nist.gov/ns/oscal";
+
+/**
+ * Return the given namespace, or if undefined, the default namespace.
+ */
+export function namespaceOf(ns) {
+  return ns ?? NIST_DEFAULT_NAMESPACE;
+}
```
kylelaker, line 12:
```suggestion
/**
 * Return the given namespace, or if undefined, the default namespace.
 * 
 * @param {string | undefined} ns the namespace
 * @returns {string} the given namespace or the the default namespace
 */
export function namespaceOf(ns) {
  return ns ?? NIST_DEFAULT_NAMESPACE;
}
```

src/components/oscal-utils/OSCALPropUtils.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Returns the first matching property in the given list.
+ *
+ * @param props the list of properties to check
+ * @param filter the attributes to match against
+ */
+export default function matchingProp(props, filter) {
```
kylelaker, line 20:
```suggestion
/**
 * @callback PropertyFilterValueMatcher
 * @param {string} value - the value to check against
 * @returns {boolean} whether the given property matches
 */

/**
 * A filter to apply to properties to find a matching property.
 *
 * This allows matching based on the namespace (defaulting to the default namespace),
 * the name, and the value. Exactly one of `value` or `filter` must be supplied.
 *
 * @typedef PropertyFilter
 * @type {object}
 * @property {string|undefined} ns - the namespace to check in
 * @property {string| name - the name of the field to match
 * @property {string} value - the value of the field to match
 * @property {PropertyFilterValueMatcher} filter - a function to match the value against
 */

/**
 * Returns the first matching property in the given list.
 *
 * @param props {Array.<Object>} the list of properties to check
 * @param filter {PropertyFilter} the attributes to match against
 */
export default function matchingProp(props, filter) {
```

I think we can represent the types from before as JSDoc types if we want to... I don't know whether this is useful enough or not.

Bronstrom: Would this info make more sense to add once we add TypeScript? . . . or maybe this could be helpful to who implements TS here?

kylelaker: Fair yeah -- maybe we just add it then.

kylelaker (approved):
Good stuff!

src/components/oscal-utils/OSCALPropUtils.js
Bronstrom: Reviewing over it again, this documentation is valuable even now with JS and TS support is just around the corner.

src/components/oscal-utils/OSCALCatalogUtils.js
```diff
@@ -0,0 +1,15 @@
+import matchingProp from "./OSCALPropUtils";
+
+/**
+ * Determines if a control is withdrawn.
+ *
+ * @param {object} control
+ * @returns a boolean describing whether the control is withdrawn
```
tuckerzp, line 7: Can we also update this `@returns`?
```suggestion
 * @returns {boolean} true if control is withdrawn
```

tuckerzp (approved):
This looks great! Thanks for your work on this @Bronstrom

PR #733, opened 2023-03-20
feat: display remove control elements

### LINKED ISSUE #89: Display of Removed Control Elements in Profiles
As a user involved in the governance, risk, and compliance of systems and components, I want to browse an OSCAL Profile in a user interface and see its removed control elements so that I can better understand the control modifications within it.

### Acceptance Criteria
* The Profile viewer shows the removal of control elements of an OSCAL Profile

### Implementation Notes
* This may require a deeper investigation of the current [Profile resolution guidance from NIST](https://pages.nist.gov/OSCAL/concepts/processing/profile-resolution/)
* Perhaps a simple strikethrough font variant of the control element removed would be a sufficient visual indicator

### LINKED ISSUE #91: Support Include-All and Exclude Controls within Profile Viewer
As a user involved in the governance, risk, and compliance of systems and components, I want to browse an OSCAL Profile in a user interface and see the proper set of controls if [include-all](https://pages.nist.gov/OSCAL/reference/latest/profile/json-reference/#/profile/imports/include-all) or [exclude-controls](https://pages.nist.gov/OSCAL/reference/latest/profile/json-reference/#/profile/imports/exclude-controls) has been defined for a catalog import so that I can better understand the control modifications within it.

### Acceptance Criteria
* Any profile using `include-all` in a catalog import renders all controls in that catalog. unless excluded
* Any profile using `exclude-controls` in a catalog import should not render those excluded controls

### Implementation Notes
* This will likely require a close examination of the [processing guidance defined by NIST](https://pages.nist.gov/OSCAL/concepts/processing/).

mpemy:
This pull request is for the Display remove control elements in Profile.
For testing this implementation, you can use this FedRAMP profile.
https://gist.githubusercontent.com/mpemy/9ea8e9b285cc9198ded3fa46d4154850/raw/9969488e66064281e8394305d352ae73cb60d9ec/Correct_rev4_LI-SaaS-baseline_profile.json

Closes #89 
Closes #91


tuckerzp:
@mpemy can you provide a screenshot of this?

mpemy:
Another one
image

src/components/OSCALControlModification.js
```diff
@@ -42,13 +42,30 @@ function getAlterAddsOrRemovesDisplay(addsElements, addsLabel, controlPartId) {
         Name: {item.name}, Value: {item.value}
       </Typography>
     ));
+  let removedTypographies = null;
+  if (addsLabel === "Removes ") {
+    removedTypographies = addsElements
+      .flatMap((elt) => elt ?? [])
+      .map((y) => (
```
tuckerzp, line 49: I dont think we need to shorten `element` here. Can we also change `y` to something more descriptive like `item`?

src/components/OSCALControlModification.js
```diff
@@ -42,13 +42,30 @@ function getAlterAddsOrRemovesDisplay(addsElements, addsLabel, controlPartId) {
...
+        <Typography
+          color="textSecondary"
+          paragraph
+          variant="body1"
+          key={y.id}
+          style={{ textDecorationLine: "line-through" }}
+        >
+          Attribute: {Object.keys(y)[0]}, Value: {Object.values(y)[0]}
```
tuckerzp, line 57: Is there a reason that this is so different than how we do the `adds` on line 42?

https://github.com/EasyDynamics/oscal-react-library/blob/57b305bba845f4c6593549d57246f23248e8b8d8/src/components/OSCALControlModification.js#L42

I feel like it is a bit unclear what values we are actually displaying here. It may be that `adds` and `removes` are just very different in the `json`...

src/components/OSCALControlModification.js
```diff
@@ -42,13 +42,30 @@ function getAlterAddsOrRemovesDisplay(addsElements, addsLabel, controlPartId) {
         Name: {item.name}, Value: {item.value}
       </Typography>
     ));
+  let removedTypographies = null;
+  if (addsLabel === "Removes ") {
```
tuckerzp, line 46: If `addsLablel === "Removes "` is it really an **adds** label? Same for `addsElements`. 

The function documentation also describes the `addsLabel` as a boolean variable which makes this more confusing.

tuckerzp (commented):
Looking good so far! I had a few clarifying questions.

src/components/OSCALControlModification.js
mpemy: Thanks for your suggestions, I will adjust those labels accordingly.

src/components/OSCALControlModification.js
mpemy: This label should be string, I changed it to the more descriptive addsRemovesLabel.

src/components/OSCALControlModification.js
```diff
@@ -42,13 +46,33 @@ function getAlterAddsOrRemovesDisplay(addsElements, addsLabel, controlPartId) {
...
+      .map((item) => (
+        <Typography
+          color="textSecondary"
+          paragraph
+          variant="body1"
+          key={item.id}
+          style={{ textDecorationLine: "line-through" }}
+        >
+          Attribute: {Object.keys.length > 0 ? Object.keys(item)[0] : ""},
+          Value: {Object.values.length > 0 ? Object.values(item)[0] : ""}
```
tuckerzp, line 62: As far as design goes, I feel like this may not be the best strategy. What you are doing now is just displaying the `json` reference on what should be removed with a strike through. Based on my understanding, a `remove` element is a reference to the component to remove. 

> Perhaps a simple strike through font variant of the control element removed would be a sufficient visual indicator

I'd assumed based on Ray's comment you would use that remove reference in the `json` to find the component to remove. I'm not sure the best design from there and maybe this should have been discussed earlier. Some intial ideas could be:
- You may want to put a strike though on the element being referenced if it is currently being displayed. 
- Another potential option would be to list the information of the element being referenced with a strike through.

src/components/OSCALControlModification.js
```diff
@@ -19,11 +19,15 @@ const OSCALControlModificationsButton = styled(IconButton)(
  * Create a typography html object for addsOrRemovesLabel
  *
  * @param {Object} addsElements add or remove object to map into html
- * @param {String} addsLabel boolean variable, true if adding
+ * @param {String} addsRemovesLabel string variable
```
tuckerzp, line 22: I think saying the string variable `addsRemovesLabel` is a string variable is not super descriptive. Please update this.

I also am not sure I love that variable name either, but am blanking on a better one right now.

src/components/OSCALControlModification.js
```diff
@@ -19,11 +19,15 @@ const OSCALControlModificationsButton = styled(IconButton)(
...
  * @param {String} controlPartId Control part ID to match
  * @returns html object
  */
-function getAlterAddsOrRemovesDisplay(addsElements, addsLabel, controlPartId) {
+function getAlterAddsOrRemovesDisplay(
+  addsElements,
```
tuckerzp, line 27: Is this only the `adds` elements? 

I think you could change a lot of the `addsorremoves` to be `alters`. So `altersElements`, `altersLabel`, etc.

tuckerzp (changes requested):
I think we need to have some discussion on the overall design here. I'd need to look into the issue and actual `json` structure a bit more before I could be too prescriptive on best solution, but I am thinking we may need to change how we are displaying removed elements.

src/components/OSCALControlProse.js
```diff
@@ -169,7 +169,11 @@ function getTextSegment(text, key) {
   }
   return (
     <Typography component="span" key={key}>
-      <OSCALMarkupLine>{text}</OSCALMarkupLine>
+      {text.includes("\n") ? (
+        <OSCALMarkupMultiLine>{text}</OSCALMarkupMultiLine>
+      ) : (
+        <OSCALMarkupLine>{text}</OSCALMarkupLine>
+      )}
```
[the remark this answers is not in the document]
mpemy: OSCALMarkupLine will wrap everything and ignore "\n" because it is designed to only return a single line. That is why if the input contains "\n" we need to use OSCALMarkupMultiLine.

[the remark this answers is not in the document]
tuckerzp: I am thinking this is issues with branching... That change _should_ be in `develop` already. 

https://github.com/EasyDynamics/oscal-react-library/blob/3a96c641983ceef29cb004ff5c255c7f835ae44d/src/components/OSCALControlProse.js#L171-L177

src/components/OSCALControlModification.js
Bronstrom: I think this is a decent start overall to the design, but yes, I feel like the strikethrough isn't in the intended place.

> You may want to put a strike though on the element being referenced if it is currently being displayed.

I favor this approach most. I believe this is what Ray was intending in the issue but didn't explicitly say. I would still include the "Removes" section in the "Modifications" box (the json reference), but have the removed text be clear here (no strikethrough).

tuckerzp: > I would still include the "Removes" section in the "Modifications" box (the json reference), but have the removed text be clear here (no strikethrough).

I think something to consider if we do include that, is all `by-___` [properties](https://pages.nist.gov/OSCAL/reference/latest/complete/json-reference/#/profile/modify/alters/removes/by-name) can be used to identify multiple items to remove. I'd argue it is more useful to list everything we remove rather than just listing the criteria we are using to remove things. 

We even could display the criteria and then all items removed underneath it. Something like:

---

## Modifications

### Removes

#### Attribute: `by-id` Value: `ac-2_smt.c`

- item 1
- item 2

---

src/components/OSCALControlModification.js
kylelaker: Well we really have an object with multiple keys. In fact, I think the implementation of `adds` here is pretty basic and focuses purely on `props`; however using `adds` we can actually add anything. I thought we supported more but I really guess we don't.

The structure for an [`add`](https://pages.nist.gov/OSCAL/reference/latest/profile/json-reference/#/profile/modify/alters/removes) is actually very similar to that for [`remove`](https://pages.nist.gov/OSCAL/reference/latest/profile/json-reference/#/profile/modify/alters/removes).

So in that case, even `add` has more to support adding than what we're showing.

So we're supporting `removes` somewhat unequally here, by supporting removing more than just props. Except it's represented in a slightly confusing way because of unclear guidance in the ticket.

My expectation was that in supporting `removes` we would not just put a strike-through on text in the dialog; but that we would list `removes` where we cannot support strike-throughs (such as for props or links) and then we would use the strike-through for text in the area where we support viewing prose (so that you can see those modifications inline).

I am worried at this point that properly supporting this will require a more holistic rework of how we manage profiles, their resolution, and how the `OSCALControl` and other components operate.

src/components/OSCALControlModification.js
kylelaker: Yeah this string comes from https://github.com/EasyDynamics/oscal-react-library/blob/57b305bba845f4c6593549d57246f23248e8b8d8/src/components/OSCALControlModification.js#L145-L163 and it's an odd architectural decision...

src/components/OSCALControlModification.js
mpemy: I revised the code this a screenshot the new implementation
image

mpemy:
I pushed changes for my two issues #89 and #91 in this branch. I am sorry it may be confusing when reviewing, moving forward each issue will have its own branch.
As for issues #91 you will need a profile that has include-all and exclude-control.
The following fedRAMP profile has include-all and exclude-control
https://gist.githubusercontent.com/mpemy/b9bbae968fb0b700bc669e3c8288b223/raw/5e3fc1e6abfa284399870f0ac794e37dc41e3987/FedRAMP_rev4_LI-SaaS-baseline_profile.json
This other profile only has include-control and exclude-control.
https://gist.githubusercontent.com/mpemy/9ea8e9b285cc9198ded3fa46d4154850/raw/9969488e66064281e8394305d352ae73cb60d9ec/Correct_rev4_LI-SaaS-baseline_profile.json

src/components/OSCALControlProse.js
```diff
@@ -347,8 +349,39 @@ export function OSCALReplacedProseWithParameterLabel(props) {
...
+        {props.label}
+        {prose}
+        {props.modificationDisplay}
+      </Typography>
+    );
+  }
+
   return (
-    <Typography>
+    <Typography style={{ textDecorationLine: "line-through" }}>
```
tuckerzp, line 384: I think it is usually better to not do `css` things in line like this. 

Maybe something like what we do in MetaData?

https://github.com/EasyDynamics/oscal-react-library/blob/83159a59b03a1c1c53359ef0d9cb2165cf05b825/src/components/OSCALMetadata.js#L40-L43

src/components/OSCALControl.js
```diff
@@ -73,10 +73,18 @@ function ControlsList(props) {
...
+        !props.includeControlIds.includes(props.control.id)))
+  ) {
+    return null;
+  }
   if (
     !props.control ||
-    (props.includeControlIds &&
-      !props.includeControlIds.includes(props.control.id))
+    (props.excludeControlIds &&
+      props.excludeControlIds.includes(props.control.id))
```
tuckerzp, line 87: I don't love how complex these if statements are getting, but I also am not sure of a better solution.

tuckerzp (commented):
Looking over this, it seems like we only are dealing with things being removed `by-id`. Looking at the [remove outline](https://pages.nist.gov/OSCAL/reference/latest/profile/json-reference/#/profile/modify/alters/removes), it seems we can also remove by name, class, item-name, and ns. 

The acceptance criteria for #89 does not explicitly say we need to all of them, but it also doesnt say to just do `by-id`. Would this be a big change in this pr?  

Overall though this looks a lot better, I am starting to wonder if maybe we push refactoring some of this code to a separate issue. It seems like a lot of this needs to be reworked when we rework profile resolution.

src/components/OSCALControlProse.js
mpemy: Thanks for your feedback. I implemented by-id because I had profiles that have remove by-id, you are right that we can also remove by class, ns, ..., etc. The logic should be the same, but the only issue is to have data (profiles) with those edge cases to test the new implementation.

mpemy: @tuckerzp, I will adjust this line of code <Typography style={{ textDecorationLine: "line-through" }}> and use css as you suggested. Thanks for the example.

src/components/OSCALProfile.js
```diff
@@ -58,11 +58,23 @@ export default function OSCALProfile(props) {
...
+    includeAll = true;
+  }
   // Flatten controls and IDs into single key, value structure
   const includeControlIds = props.profile.imports
     .flatMap((imp) => imp["include-controls"])
     .flatMap((includeControl) => includeControl["with-ids"]);
 
+  const excludeControlIds = props.profile.imports
+    .flatMap((imp) => imp["exclude-controls"] ?? [])
+    .flatMap((excludeControl) => excludeControl["with-ids"] ?? []);
```
kylelaker, line 76: Very clean! Can you actually apply this `?? []` "safety" to the above lines (currently 70-72) dealing with `flatMap`ing the includes?

kylelaker:
It will be necessary to pull in the changes from the `develop` branch and resolve the merge conflicts. I am happy to help with that as I think they conflict with the work I just did in #746

src/components/OSCALProfile.js
mpemy: Thanks @kylelaker.  I will apply ?? [] to line 70 - 72.

src/components/OSCALControlProse.js
```diff
@@ -359,21 +363,39 @@ export function OSCALReplacedProseWithParameterLabel(props) {
...
+    ?.find((item) => item["control-id"] === controlId)
+    ?.removes?.find((object) => object["by-item-name"] === controlPartId);
+
+  if (
+    (removeByIds !== null && removeByIds !== undefined) ||
+    (removeByNames !== null && removeByNames !== undefined) ||
+    (removeByNS !== null && removeByNS !== undefined) ||
+    (removeByClass !== null && removeByClass !== undefined) ||
+    (removeByItemNames !== null && removeByItemNames !== undefined)
+  ) {
```
kylelaker, line 388:
```suggestion
  if (removeByIds || removeByNames || removeByNS || removeByClass || removeByItemNames) {
```

`null` and `undefined` are both falsey, so we can shorten this.

mpemy: Thanks @kylelaker for this suggestion. I will commit your suggestion.

kylelaker (approved):
Thanks for all the work on this and for the iterations from review! I think at this point we're good to go ahead and merge. I still want to make sure that in the future we handle profile resolution more holistically; however, we have a separate ticket to implement that.

src/components/OSCALProfile.js
tuckerzp: @mpemy have you tried this?

tuckerzp (commented):
@mpemy I had a few comments on things that were most likely lost in the sea of comments. 

Since a lot of this will need refactoring, I am ok with the fact that we did not add unit tests for this. Would you be able to create an issue to test your added functionality at some point?

tuckerzp:
@mpemy can you send a link to that issue in chat? I will move forward with merging once that is done. I just don't want to forget to add tests for this later.

tuckerzp (approved):

PR #746, opened 2023-03-24
fix(control): controls display id instead of label

kylelaker:
Currently, controls are often displayed with their `id` in uppercase instead of using the `label` property. This means that the display is not necessarily always what a document author may intend (even if it perhaps is close).

The label is displayed matching the surrounding content. If the ID is used, it falls back to displaying using the ID (in our standard grey color).

This also fixes an issue where, for prose, the label was squished up against the text without a space.

#### Before

image

#### After

image

#### Before

image

#### After

image

### DIFF
--- a/src/components/OSCALCatalogGroups.test.js
+++ b/src/components/OSCALCatalogGroups.test.js
@@ -35,6 +35,22 @@ const testGroups = [
   },
 ];
 
+function getTextByChildren(text) {
+  function result(_content, node) {
+    const hasText = (checkNode) => checkNode.textContent === text;
+    const nodeHasText = hasText(node);
+    // This is necessary because we are providing a query function to override how
+    // text search is performed.
+    // eslint-disable-next-line testing-library/no-node-access
+    const childrenDontHaveText = Array.from(node.children).every(
+      (child) => !hasText(child)
+    );
+
+    return nodeHasText && childrenDontHaveText;
+  }
+  return result;
+}
+
 describe("OSCALCatalogGroup", () => {
   test("displays param legend", () => {
     render(<OSCALCatalogGroups groups={testGroups} />);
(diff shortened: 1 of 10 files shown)


src/components/OSCALProfile.test.js
```diff
@@ -24,7 +24,7 @@ function testOSCALProfile(parentElementName, renderer) {
   jest.setTimeout(6000);
   test(`${parentElementName} displays controls`, async () => {
     renderer();
-    const result = await screen.findByText("ac-1", {
+    const result = await screen.findByText("AC-1", {
```
kylelaker, line 27: This changes because we now properly use the _label_ which does specify the value in uppercase.

src/components/OSCALCatalogGroups.test.js
```diff
@@ -35,6 +35,22 @@ const testGroups = [
   },
 ];
 
+function getTextByChildren(text) {
```
kylelaker, line 38: This is necessary because we're now potentially pulling text across multiple `<span>`s.

src/components/OSCALControlProse.js
```diff
@@ -488,7 +486,7 @@ export function OSCALReplacedProseWithByComponentParameterValue(props) {
           <Link underline="hover" href={`#${props.label}`}>
             {props.label}
           </Link>
-        </StyledTooltip>
+        </StyledTooltip>{" "}
```
kylelaker, line 489: This fixes the issue with the label being "squished" up against the actual text. Same for the changes above where `{props.label}` and `{prose}` are now on the same line.

kylelaker (commented):
@tuckerzp

src/components/OSCALCatalogGroups.test.js
```diff
@@ -58,7 +74,7 @@ describe("OSCALCatalogGroup", () => {
     fireEvent.click(expand2);
 
     const expand3 = screen.getByText(
-      "CONTROL-ID Access Control Policy and Procedures"
+      getTextByChildren("control-id Access Control Policy and Procedures")
```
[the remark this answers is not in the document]
Bronstrom: Wow, so smart!

Issue #743, opened 2023-03-23
Unable to expand withdrawn controls


kylelaker:
### Description

A control can itself have controls within it (for example, SP 800-53 uses this to represent control enhancements). A control may also be withdrawn (represented by a prop where `@name="status"` and `@value="withdrawn" in the appropriate namespace).

A user may want to be able to view the details (prose and enhancements) of a withdrawn control; however, in the application today, the ability to expand a control is disabled when a control is withdrawn. This makes viewing these details impossible. In SP 800-53rev5, these details seem to often not be present; however, other catalogs may choose to still provide prose and other information for withdrawn controls.

### Environment

oscal-viewer.msd.easydynamics.com

### Steps to Reproduce

*A catalog to reproduce will be provided soon*

### Expected behavior

I should be able to expand the detail view of a withdrawn control if there's something to see there.

### Actual Behavior

It is never possible to expand a withdrawn control.

### Additional Notes

We get input from @brian-easyd to find out what it means if a "parent" control is withdrawn but the inner controls are not labeled as withdrawn. For example:

```json
{
  "controls": [
    {
      "id": "repro-A",
      "title": "Sample Withdrawn",
      "props": [
        { "name": "status", "value": "withdrawn" }
      ],
      "controls": [ {"id": "repro-A.1", "title": "Parent is withdrawn but this isn't labeled as such" } ]
    }
  ]
}
```

Should `repro-A.1` have strikethrough? Or should it appear normally?

brian-easyd:
Ensure consistent styling of text for withdrawn content.

If a parent control is withdrawn, all child controls are considered withdrawn as well.

brian-easyd:
Display of withdrawn controls should include the same content as a normal control. It should be formatted differently and display some kind of "Withdrawn" label.

kylelaker:
Closed by #720

PR #758, opened 2023-04-07
fix(catalog): Support multi-level group & high-level withdrawn control anchor links

Bronstrom:
This PR supports anchor links for multiple-levels of control groups.

## Testing

Navigate to the Catalog Viewer and reload the following catalog(s): 
* https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/brad/test-catalog-multi-layered-groups/catalogs/multi_layered_group_test_catalog.json. Test out any of multi-level catalog groupings.
* https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/test-content/catalogs/basic-test-catalog.json. Test out "Test Group E" and ensure the anchor links open, scroll, and navigate properly.

Relates to #176


tuckerzp:
Is there sample data or a test to ensure this works? I remember some talk on this being tricky, but this also seems like something we'd prefer to not just trust it works. 

At the very least discussion on why this can't be done would be good here. In the past, we have made dummy data to at least show a feature works even if we don't push it to our actual sample data.

kylelaker (approved):

tuckerzp (changes requested):
I think some changes in your previous pr to groups may have broken

https://oscal-viewer.msd.easydynamics.com/catalog/?url=https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/test-content/catalogs/basic-test-catalog.json

Bronstrom:
Doing some additional testing on this issue, I believe this is now in proper working order to fully support multi-level groups.

> I think some changes in your previous pr to groups may have broken how we handle groups within groups.

Yes, when ensuring that a `group.title` is used when `group.id` isn't present I missed one place, and it would cause the application to fail right away when no `group.id` is provided by the first highest-level group. I've corrected this.

> https://oscal-viewer.msd.easydynamics.com/catalog/?url=https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/test-content/catalogs/basic-test-catalog.json

Hey, this is actually a pretty good example to test multi-layered groups. Thanks for sharing this! I'll go ahead and place it in the description under "Testing".

Bronstrom:
I've tried to stress test these groups and they seem to be working as intended. I also noticed that high-level withdrawn controls had some state issues, so I've made a fix to them as well.

Please refer to the test data provided above, when testing.

kylelaker:
> I also noticed that high-level withdrawn controls had some state issues in a similar vein, so I've made a fix to them as well

Are you talking about #743 or another issue?

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
```diff
@@ -42,16 +42,38 @@ const StyledControlDescriptionWrapper = styled("div")`
...
 
+/**
+ * Validates fragment and determines if top fragment item matches a list ID.
+ *
+ * @param {*} urlFragment The fragment currently being handled
+ * @param {*} previousHandledFragment The previous handled fragment
+ * @param {*} fragmentSuffix String with current item and sub items
+ * @param {*} listId The group/control ID
+ * @returns
+ */
```
tuckerzp, line 53: I don't want to hold this issue more than necessary, so I won't ask for TypeScript, but I also really hope that some of these parameters can't just be anything. Can we be more specific?

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
```diff
@@ -113,7 +113,7 @@ export function shiftFragmentSuffix(fragmentSuffix: string): string | undefined
  */
 export function conformLinkIdText(linkText: any): string {
```
tuckerzp, line 114: I would love if this didn't have to be `any`. The word `linkText` seems to imply this should be a `string`. Even the jsdoc says it is a string. But then we do `linkText.props.children`... 

Is it a small change to just make this a `string` and deal with those consequences?

packages/oscal-react-library/src/components/OSCALCatalogGroups.js
```diff
@@ -112,9 +119,12 @@ export default function OSCALCatalogGroups(props) {
     }
     // Confirm catalog tab group can be grabbed
     if (document.getElementById(`vertical-tab-${rootLayer}`)) {
+      console.log("Heyo");
```
tuckerzp, line 122:
```suggestion
```

packages/oscal-react-library/src/components/OSCALCatalogGroups.js
```diff
@@ -81,27 +81,34 @@ function determineControlExists(groups, controlLayers, rootLayer) {
...
-  // Ensure lowest/deepest control exists
+  // Ensure lowest/deepest control exists within controls/groups
   for (let i = 1; i < controlLayers.length && upperLayer; i += 1) {
-    upperLayer = upperLayer?.controls?.find((control) => control.id === controlLayers[i]);
+    upperLayer =
+      upperLayer?.controls?.find((control) => control.id === controlLayers[i]) ||
+      upperLayer?.groups?.find(
+        (group) =>
+          group.id === controlLayers[i] || conformLinkIdText(group.title) === controlLayers[i]
+      );
```
tuckerzp, line 91: Is it possible to split this out a bit? There is a lot happening here. Maybe have like:

```JavaScript
const find_1 = ...
const find_2 = ...

upperLayer = find_1 || find_2
```

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
```diff
@@ -133,59 +216,85 @@ function OSCALCatalogControlListItem(props) {
...
-  ) : (
-    <StyledListItemPaper>
-      <StyledListItem>
-        <WithdrawnListItemText primary={itemText} withdrawn={withdrawn} />
-      </StyledListItem>
-    </StyledListItemPaper>
   );
 }
 
 function OSCALCatalogGroupList(props) {
```
tuckerzp, line 226: Again, this would be a really nice place for an `interface`. Can we at least add some `jsdoc`?

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
```diff
@@ -64,32 +86,71 @@ function CollapsibleListItem(props) {
...
+        onEntered={() => !group && setListItemOpened(true)}
+        unmountOnExit
+      >
         <StyledControlDescriptionWrapper>{children}</StyledControlDescriptionWrapper>
       </Collapse>
     </StyledListItemPaper>
   );
 }
 
+function WithdrawnListItem(props) {
```
tuckerzp, line 127: Same thing here. `jsdoc` will make switching to TypeScript a lot less painful.

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
```diff
@@ -133,59 +216,85 @@ function OSCALCatalogControlListItem(props) {
...
     urlFragment,
     fragmentPrefix,
     fragmentSuffix,
     isControlListItemOpened,
     setIsControlListItemOpened,
+    previousHandledFragment,
+    setPreviousHandledFragment,
   } = props;
-
+  const [isListItemNavigatedTo, isSetListItemNavigatedTo] = React.useState(false);
```
tuckerzp, line 237:
```suggestion
  const [isListItemNavigatedTo, setIsListItemNavigatedTo] = React.useState(false);
```

Is this not more standard?

tuckerzp (changes requested):
@Bronstrom I really appreacite all your hard work! These are some fairly complex changes, so my comments are mostly just about clarity. I think a bit of jsdoc and renaming could make things a bit easier to understand here.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
tuckerzp: Only commenting on this since it was touched in this pr.

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
Bronstrom: Good point, I rushed this JSDoc and missed some details.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
Bronstrom: Currently `linkText` can be anything to almost everything including button components and styling type things. I think if we restrict this to a string, some anchor links functionality will be slimmed down a bit. I see 2 ways of solving this problem:
1) Rename linkText to something like linkElement and fix the `@param` to `any`
2) Make this only a `string` type and remove unnecessary logic here and double check other places in the codebase that use this function. ` value` will need to be added to most AnchorLinks throughout the application.

packages/oscal-react-library/src/components/OSCALCatalogGroups.js
Bronstrom: You're telling me we don't need a "Heyo" logged every time a Catalog group tab is matched to part of a fragment? 😂

packages/oscal-react-library/src/components/OSCALCatalogGroups.js
Bronstrom: Definitely better clarity written that way. I don't this would have much impact on the runtime either.

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
Bronstrom: I'm wondering to what extent we want to add JSDocs? Are we adding them just to our utility functions or are we including OSCAL components in these now?

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
Bronstrom: This goes with the question as stated previously on JSDoc support. I consider this to be like an OSCAL component, just not under the OSCAL name, similar to `CollapsibleListItem`.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
kylelaker: We can maybe do `string | ReactNode` for the type? Not 100% sure but let's try it?

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
tuckerzp: > I've added JSDoc to all OSCALCatalogGroup sub-components.

@Bronstrom For clarity, I was mostly looking for the things under `props` to be documented. `props` contains a lot of fields and, as someone reviewing, it is a bit difficult to know what they are. 

As an example, this is what we  to do in `OSCALBackMatter.tx`. Each `prop` being passed needs an associated type and ideally a description. Having those types and descriptions now in some type of comment, would really help a lot in both reviewing this code and eventually switching to `.tsx`.

https://github.com/EasyDynamics/oscal-react-library/blob/6dc2a5370d5d8cdc7b865bc84d83106f3f25923f/packages/oscal-react-library/src/components/OSCALBackMatter.tsx#L87-L97

If it is a big thing to add, we can skip for now.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
tuckerzp: Oh I see why that would not work, but that seems very unsafe. I guess we can stick with `any` for now as I can't think of a cleaner solution at the moment.

packages/oscal-react-library/src/components/OSCALCatalogGroup.js
Bronstrom: I've added documentation for the props. I can't directly place the documentation above each destructured property (the documentation is just above the `const`) and it's not going to look as nice as the JSDoc above a function, but I think it should give a better idea of the types for TS implementation.

tuckerzp: > not going to look as nice as the JSDoc

Yeah I agree it looks a bit nasty :smile:. Hopefully it will motivate us to switch to file to `.tsx` soon! Thank you for humoring me.

tuckerzp (approved):

Issue #488, opened 2022-07-09
Perform OSCAL schema validation prior to displaying


kylelaker:
### Description

As an author of an OSCAL document, I want to know if there are schema validation errors in it that prevent it from being displayed properly (or at all), so that I know if information presented in the viewer/editor is accurate.

### Acceptance Criteria

 - The user is informed on the page if the OSCAL document does not pass schema validation; and that information is presented using typical UI idioms for presenting warnings
 - The viewer/editor continues to perform best-effort presentation of the document, even if validation fails (as we do today)
 - If we do fail to display the document, the error message on the page displays the validation errors

### Proposed Solution

Validators such as `ajv` or [`usnistgov/oscal-cli`](https://github.com/usnistgov/oscal-cli) exist that can perform validation. We use `ajv` on [EasyDynamics/oscal-demo-content](https://github.com/EasyDynamics/oscal-demo-content) today (though perhaps that would be a better candidate to use `oscal-cli` while `ajv` may fit into our JS/TS world better).

We should perhaps provide a Yellow/Warning "Yikes" text (though, without "Yikes") informing a user if their OSCAL does not pass schema validation.

### Additional Notes

In some scenarios, an invalid JSON object may be passed to our viewer that theoretically we could parse and display but that isn't technically proper OSCAL (and theoretically, the inverse is true, we may fail to display valid OSCAL). 

### Implementation

- [ ] I can (or plan to) submit a pull request to implement this
- [ ] Implementing this may result in a breaking change

kylelaker:
Implemented in #830 and earlier Loader refactoring.

PR #765, opened 2023-04-07
feat: automatically generate OSCAL types from JSON Schema

kylelaker:
This adds a new package `@easydynamics/oscal-types` that will convert
the underlying OSCAL JSON schema to TypeScipt type definitions. It also
replaces two existing instances of type information in `oscal-react-library`
with these new types: Properties and BackMater within `oscal-utils`.

The types are represented _literally_. This includes things like
`media-type` rather than making it the friendlier `mediaType`.
Additionally, to encourage good hygiene, all types are represented as
`readonly`.


kylelaker:
To do:

- [ ] Automatically update the types in a GitHub workflow
- [ ] Add a publishing step (to GitHub and to npm)
- [ ] Ensure there aren't other low-hanging fruit in the OSCAL React Library to adopt types

packages/oscal-react-library/src/components/OSCALLoader.js
```diff
@@ -81,16 +82,16 @@ export default function OSCALLoader(props) {
     fetch(newOscalUrl)
       .then((response) => {
         if (!response.ok) throw new Error(response.status);
-        else return response.json();
+        else return response.text();
       }, handleError)
       .then((result) => {
-        if (!unmounted.current) {
+        if (!unmounted.current && result) {
+          const oscalObj = Convert.toOscal(result);
```
kylelaker, line 89: This is probably the most significant change in this PR. Functionally, it ensures that we are validating every input document as _structurally_ valid OSCAL upon loading it.

This should be phenomenal for safety across the code base, even in places where we're still using JavaScript. The downside here is that we may reject documents that are "mostly" correct.

Note the emphasis on _structural_. Like all the TypeScript changes, this is focused on purely the structure of the objects (do fields exist) not the constraints (must meet particular string formats).

This may resolve #488.

scripts/generate-types.ts
```diff
@@ -0,0 +1,358 @@
...
+  const schemaInput = new JSONSchemaInput(undefined);
+  const inputData = new InputData();
+  for (const schema of schemas) {
+    schemaInput.addSource({ name: schema.typeName, schema: schema.schemaContent });
+  }
+  inputData.addInput(schemaInput);
+  return quicktype({
+    inputData,
+    lang: new JsiiCompatTypeScriptTargetLanguage(),
+    rendererOptions: { 'nice-property-names': 'false' },
```
kylelaker, line 296: We can later add `{ "converters": "all-objects" }` (default is `"top-level"`) if we want to support converters for literally every object in the stack (which we may need later).

scripts/generate-types.ts
```diff
@@ -0,0 +1,358 @@
...
+  // however, doing that and emitting it to a single document currently results in a lot
+  // of duplicate types beign emitted (for example, one `Metadata` for each root document
+  // type). Instead, we'll use the complete schema. This generates a very large file but
+  // at least it makes things a little bit cleaner.
+  const oscalSchemas = [
+    {
+      name: 'OSCAL',
+      file: 'complete'
+    },
+  ]
```
kylelaker, line 317:
```suggestion
  ];
```

scripts/generate-types.ts
```diff
@@ -0,0 +1,358 @@
...
+    console.error(err);
+  }
+}
+
+async function main() {
+  const args = process.argv.slice(2);
+  if (args.length !== 2) {
+    console.error("Usage: generate-types.ts <OSCAL-VERSION> <SCHEMA-DIR> <PACKAGE-PATH>")
+  }
+  const schemaFile = args[0]
```
kylelaker, line 343:
```suggestion
  const schemaFile = args[0];
```

packages/oscal-react-library/src/components/OSCALLoader.js
tuckerzp, line 89: I think you need to have some error handling here to make the error in the promise get caught and handled. Right now when validation fails, it spins forever. 

For example, try loading https://raw.githubusercontent.com/usnistgov/OSCAL/main/json/schema/oscal_assessment-plan_schema.json as a document.

kylelaker: Oof -- good idea testing something that is actually invalid. I probably should have considered that code path :D

Bronstrom (approved):

PR #768, opened 2023-04-12
feat(back-matter): edit backmatter resources

### LINKED ISSUE #687: Support editing basic back matter resource data
### Description

As a catalog author, I want to edit the title, description, and other basic metadata of back matter resources

### Acceptance Criteria

- `title` (`markup-line`) and `description` (`markup-multiline`) fields are properly displayed within the library (this should already be completed but let's fix-up any issues and avoid regressions)
- `title` and `description` fields for back matter resources can be edited and saved within the library
- When saved, changes to the `title` and `description` field are sent to the REST API following the [specification](https://editor.swagger.io/?url=https://raw.githubusercontent.com/EasyDynamics/oscal-rest/develop/openapi.yaml) using the best available method

### Proposed Solution

It is likely best to either (a) add new routes to the API specification to support back matter editing; or (b) use the `PATCH /{documentType}/{id}` routes

### Additional Notes

_No response_

### Implementation

tuckerzp:
This adds basic editing functionality for backmatter resource
titles and descriptions.

Our original editing functionality made some assumptions that made
implementing this tricky. Namely, we assumed that the text field we are
updating is the direct child of the field we are sending a PATCH request
for. [^1]

As you can see in the code, the way we handle editing is a bit hacky and
naive. We most likely want to revisit this and redo a lot of that code.
This would not be an insignificant change, hence it being pushed for
now. These changes make editing at least work for now so that our other
editing tasks are no longer blocked.

closes #687 

[^1]: For example, `metadata.title` updating `metadata` vs `backmatter.resources[i].title` updating all of `backmatter`.


tuckerzp:
Tests are failing because I changed how description is displayed. Should be an easy fix, I will wait until we discuss approach before I fix it.

packages/oscal-react-library/src/components/OSCALBackMatter.js
```diff
@@ -91,34 +65,82 @@ export default function OSCALBackMatter(props) {
...
+                      : null
+                  }
+                  editedValue={props.backMatter.resources}
+                  editedValueId={resource.uuid}
+                  onFieldSave={props.onFieldSave}
+                  partialRestData={
+                    props.isEditable
+                      ? {
+                          [Object.keys(props.partialRestData)[0]]: {
+                            uuid: props.partialRestData[Object.keys(props.partialRestData)[0]].uuid,
```
kylelaker, line 85: Can you clarify why we need to pass the UUID in this body? it should not be necessary in the body of the PATCH?

packages/oscal-react-library/src/components/OSCALBackMatter.js
```diff
@@ -91,34 +65,82 @@ export default function OSCALBackMatter(props) {
...
+                      ? [Object.keys(props.partialRestData)[0], "back-matter", "resources"]
+                      : null
+                  }
+                  editedValue={props.backMatter.resources}
+                  editedValueId={resource.uuid}
+                  onFieldSave={props.onFieldSave}
+                  partialRestData={
+                    props.isEditable
+                      ? {
+                          [Object.keys(props.partialRestData)[0]]: {
```
kylelaker, line 84: Can we maybe extract this to a function?

```ts
function getObjectRootKey(object: any): string {
  const keys = Object.keys(object);
  if (keys.length !== 1) {
    throw new Error(`The given object does not have exactly one root key ${JSON.stringify(object)`);
  }
  return keys[0];
}
```

And then we can do:

```ts
const objectKey = getRootKey(props.partialRestData);
```
and

```
partialRestData = {
  [objectKey]: {
    uuid: props.partialRestData[objectKey].uuid,
    "back-matter": {
      resources: props.backMatter.resources,
    },
  },
}

**Comment by kylelaker** [comment]
Great stuff! This isn't quite as thorough of a rework as I'd like to have seen overall but it gets things working for this task. What we still need is a written description of how to use this framework; can you please write some documentation?

I don't care whether it's docs on those lower-level functions or docs in Markdown written somewhere but we need documentation on how this works because we have a ton of upcoming editing tasks.

File: packages/oscal-react-library/src/components/OSCALEditableFieldActions.js [hunk 3]
```diff
@@ -9,6 +9,12 @@ export function getElementLabel(editedField) {
   return editedField.toString().replace(/,/g, "-");
 }
 
+export function updateListItem(list, uuid, field, value) {
+  const updatedItem = list.find((item) => item.uuid === uuid);
+  updatedItem[field] = value;
+  return list;
+}
```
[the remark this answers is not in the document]
tuckerzp: It does need to be exported, it has been updated with types.

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,178 @@
...
+);
+
+function TitleDisplay(props: any): ReactElement {
+  const { children, uuid } = props;
+  return (
+    <OSCALAnchorLinkHeader value={uuid}>
+      <Typography variant="subtitle1">{children}</Typography>
+    </OSCALAnchorLinkHeader>
+  );
+}
```
kylelaker, line 34:
```suggestion
interface TitleDisplayProps {
  uuid: string;
  children: any | any[];
}

function TitleDisplay(props: TitleDisplayProps): ReactElement {
  const { children, uuid } = props;
  return (
    <OSCALAnchorLinkHeader value={uuid}>
      <Typography variant="subtitle1">{children}</Typography>
    </OSCALAnchorLinkHeader>
  );
}
```

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,178 @@
...
+  return (
+    <StyledTooltip title={<OSCALMarkupLine>{props.resource.citation.text}</OSCALMarkupLine>}>
+      <FormatQuoteIcon
+        color="primary"
+        fontSize="small"
+        titleAccess={`${props.resource.title}-citation`}
+      />
+    </StyledTooltip>
+  );
+}
```
kylelaker, line 55: This code actually is problematic because if `props.resource` is undefined, we are still trying to access its `title` attribute in the `titleAccess` property of `FormatQuoteIcon`... Can we just make `resource` required like this? If not, we'll have to come up with something else to do with `props.resource` is undefined.

```suggestion
interface CitationDisplayProps {
  readonly resource: Resource;
}

function CitationDisplay(props: CitationDisplayProps): ReactElement {
  if (!props.resource.citation?.text) {
    return (
      <FormatQuoteIcon
        color="disabled"
        fontSize="small"
        titleAccess={`${props.resource.title}-citation`}
      />
    );
  }
  return (
    <StyledTooltip title={<OSCALMarkupLine>{props.resource.citation.text}</OSCALMarkupLine>}>
      <FormatQuoteIcon
        color="primary"
        fontSize="small"
        titleAccess={`${props.resource.title}-citation`}
      />
    </StyledTooltip>
  );
}
```

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
kylelaker, line 34:
```suggestion
interface TitleDisplayProps {
  uuid: string;
  children: any | any[];
}

function TitleDisplay(props: TitleDisplayProps): ReactElement {
  const { children, uuid } = props;
  return (
    <OSCALAnchorLinkHeader value={uuid}>
      <Typography variant="subtitle1">{children}</Typography>
    </OSCALAnchorLinkHeader>
  );
}
```

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+function getObjectRootKey(object: any): string {
+  if (!object) {
+    return "";
+  }
```
kylelaker, line 60:
```suggestion
function getObjectRootKey(object: unknown): string {
  if (!object) {
    return "";
  }
```

Let's maybe use `unknown` here if we need to handle `undefined`, `null`, and other falsey values? You might need to change this to `if (!object || typeof object !== 'object')` or something like that.

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+export default function OSCALBackMatter(props: OSCALBackMatterProps): ReactElement {
+  const getMediaType = (rlink: ResourceLink) =>
+    rlink["media-type"] ||
+    guessExtensionFromHref(getAbsoluteUrl(rlink.href, props.parentUrl) ?? "");
+
+  const objectKey = getObjectRootKey(props.partialRestData);
+
+  const backMatterDisplay = (resource: Resource) => (
+    <Grid item xs={3} key={resource.uuid}>
```
kylelaker, line 87: For composability, rather than being a nested function, this almost feels like it should be its own component in the root of the document...


```tsx
interface EditableFieldProps {
  isEditable?: boolean;
  onFieldSave?: (...args: any[]) => void;
  partialRestData?: Record<string, any>;
}

interface BackMatterResourceProps extends EditableFieldProps {
  resource: Resource;
}

function BackMatterResource(props: BackMatterResourceProps): ReactElement {
  //... the body of this function
}
```

packages/oscal-react-library/src/components/OSCALEditableFieldActions.test.tsx
```diff
@@ -23,7 +23,7 @@ test("OSCALEditableFieldActions loads", () => {
   );
 });
 
-export default function testOSCALEditableFieldActions(parentElementName, renderer) {
+export default function testOSCALEditableFieldActions(parentElementName: any, renderer: any) {
```
kylelaker, line 26:
```suggestion
export default function testOSCALEditableFieldActions(parentElementName: string, renderer: any) {
```

I hope we can at least scope this first parameter to a string?

packages/oscal-react-library/.eslintrc.json
```diff
@@ -63,8 +63,11 @@
       {
         "devDependencies": [
           "**/*.test.js",
+          "**/*.test.tsx",
           "**/setupTests.js",
-          "**/TestUtils.js"
+          "**/TestUtils.js",
+          "jest.config.ts",
+          "rollup.config.js"
```
kylelaker, line 70: Good catch! This result is much less noise when editing.

kylelaker (commented):
Hey! I know that some of these asks go a little beyond your ticket but I am hopeful that we can do a little more cleanup while we're in here to improve the type checking and composability of some of these components. I'd even like to see us be able to `export` `BackMatterDisplay` when this is done.

If some of these are too complicated to do now, please be very explicit about the errors that `tsc` or `eslint` gives you and we can track them as tech debt to fix later.

Thank you so much for all your work on this so far! I'd approve as-is but I think it'd be good to use this to set a good example for what we're looking for in some of our refactoring as we incrementally adopt TypeScript. Especially since there's at least one bug in this code today that wasn't so obvious until now.

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
tuckerzp: I have tried to do this and it is not much cleaner. I think this is becuase you can't really edit a resource title without it being generally aware of all other resources. It may be that I am just missing something, so I will push it and see what everyone thinks.

packages/oscal-react-library/src/components/OSCALBackMatter.js
tuckerzp: The `uuid` is oddly required for the rest call.

tuckerzp:
As a note to reviewers, the diff has gotten a bit nasty here. While there is really not much change, there was enough with switching to `.tsx` that some files are not recognized as the same file.

Feel free to comment on specific lines or I can provide a bit more context on the major changes if that is helpful. Hopefully between the pr comment and some of my commit messages, what is happening is clear enough.

packages/oscal-react-library/docs/adding-editing.md
```diff
@@ -0,0 +1,28 @@
...
+- `value`: The value of the field that is changed on the front end.
+
+### Example PRs
+
+- [Editing backmatter resources](https://github.com/EasyDynamics/oscal-react-library/pull/768)
+
+[^1]: For example, "title" or "description".
+[^2]: This is only included if editing a value would update a list of objects.
+For example, editing a backmatter resource title would have an
+`editedValue` of `backmatter.resources` but a value of `resource.title`.
```
kylelaker, line 28: Do you mind breaking this into a separate branch/PR. I have thoughts here but I don't want it to hold up review on the task itself.

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+interface OSCALBackMatterProps extends EditableFieldProps {
+  backMatter: BackMatter | undefined;
+  parentUrl: string;
+  isEditable: boolean;
+}
+
+interface BackMatterResourceProps extends OSCALBackMatterProps {
+  resource: Resource;
+}
```
kylelaker, line 88: Can you add comments to these items? And export the `OSCALBackMatterProps` interface? (We should try to export the props interface for any components that we're exporting)

packages/oscal-react-library/src/components/OSCALEditableTextField.tsx
```diff
@@ -66,14 +85,38 @@ function textFieldWithEditableActions(props, reference, inEditState, setInEditSt
...
+  /**
+   * The value that is edited?
+   */
+  editedValue?: any;
+  editedValueId?: string;
+  value?: string;
+  size?: number;
+  textFieldSize?: string;
+  typographyVariant?: TypographyVariant;
+}
```
kylelaker, line 110: Fix comments

packages/oscal-react-library/docs/adding-editing.md
tuckerzp: @kylelaker After addressing your comments below, I do not see much value in this file anymore. I think the interface documentation does a better job at describing these parameters. 

It has been deleted and, unless you disagree, I don't think there is much reason to make it into a new pr.

packages/oscal-react-library/src/components/OSCALEditableTextField.tsx
tuckerzp: Fixed them! The bottom 3 are hopefully self explanatory enough.

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,210 @@
...
+                      ? {
+                          [objectKey]: {
+                            uuid: props.partialRestData?.[objectKey].uuid,
+                            "back-matter": props.backMatter,
+                          },
+                        }
+                      : null
+                  }
+                  value={resource.title}
+                />
```
tuckerzp, line 129: Actual change here

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,210 @@
...
+                ? {
+                    [objectKey]: {
+                      uuid: props.partialRestData?.[objectKey].uuid,
+                      "back-matter": props.backMatter,
+                    },
+                  }
+                : null
+            }
+            value={resource.description}
+          />
```
tuckerzp, line 156: And here

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -137,7 +137,7 @@ function BackMatterResource(props: BackMatterResourceProps) {
           </Grid>
           <OSCALEditableTextField
             fieldName="description"
-            canEdit={props.isEditable}
+            isEditable={!!props.isEditable}
```
kylelaker, line 140:
```suggestion
            isEditable={props.isEditable}
```

packages/oscal-react-library/src/components/OSCALBackMatter.test.tsx
```diff
@@ -9,60 +8,91 @@ import {
...
-export default function testOSCALBackMatter(parentElementName, renderer) {
-  test(`${parentElementName} displays resource title`, () => {
-    renderer();
+describe("OSCAL Backmatter", () => {
+  test(`displays resource title`, () => {
+    render(
+      <OSCALBackMatter
+        backMatter={backMatterTestData}
+        parentUrl={parentUrlTestData}
+        isEditable={false}
```
kylelaker, line 17:
```suggestion
        backMatter={backMatterTestData}
        parentUrl={parentUrlTestData}
```

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,210 @@
...
+    <OSCALSection>
+      <Card>
+        <CardContent>
+          <Grid container spacing={2}>
+            <Grid item xs={12}>
+              <OSCALAnchorLinkHeader>
+                <OSCALSectionHeader>Back Matter</OSCALSectionHeader>
+              </OSCALAnchorLinkHeader>
+            </Grid>
+            <Grid item xs={7}>
```
kylelaker, line 189: why 7?

packages/oscal-react-library/src/components/OSCALBackMatter.test.tsx
```diff
@@ -9,56 +8,50 @@ import {
...
-function backMatterRenderer() {
-  render(<OSCALBackMatter backMatter={backMatterTestData} parentUrl={parentUrlTestData} />);
-}
-
-export default function testOSCALBackMatter(parentElementName, renderer) {
-  test(`${parentElementName} displays resource title`, () => {
-    renderer();
+describe("OSCAL Backmatter", () => {
+  test("displays resource title", () => {
+    render(<OSCALBackMatter backMatter={backMatterTestData} parentUrl={parentUrlTestData} />);
```
Bronstrom, line 13: This Backmatter test data is duplicated throughout the tests. Is there any reason to avoid having a renderer function?

packages/oscal-react-library/src/components/OSCALEditableTextField.tsx
```diff
@@ -66,14 +85,74 @@ function textFieldWithEditableActions(props, reference, inEditState, setInEditSt
...
+  const [inEditState, setInEditState] = useState(false);
 
-  return props.canEdit ? (
+  return props.isEditable ? (
     textFieldWithEditableActions(props, reference, inEditState, setInEditState)
   ) : (
-    <Grid item className={props.className}>
+    <Grid item>
       <Typography variant={props.typographyVariant}>{props.value}</Typography>
     </Grid>
```
Bronstrom, line 157: I'm wondering how much value this `Grid` adds. Would removing the `Grid` and just having `props.value` within the `Typography` support the same cause? Or could we use a `Stack` instead?

Bronstrom (commented):
Looking good so far!

packages/oscal-react-library/src/components/OSCALBackMatter.test.tsx
kylelaker: I agree 100%. This is a small enough difference that I think it's valuable to just make the call. Additional indirection is not useful here and is actually less helpful for debugging.

I am very much opposed to these styles of "renderer()" functions.

What is the alternative? Calling `renderer()` over and over? It's still one line repeated many times. Not sure the juice is worth the squeeze.

In fact, I'd argue that our tests _should not_ be using the same data and the same parameters _every_ time; and that would discourage some magic function that solves the problem of repetition. We should solve it by it not existing in the first place

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,207 @@
...
+  ({ theme }) => `
+    margin-top: ${theme.spacing(1)};
+    display: flex;
+    flex-direction: column;
+`
+);
+
+interface TitleDisplayProps {
+  uuid: string;
+  children: React.ReactNode;
```
Bronstrom, line 30: I like the comments and examples provided with other interface properties, such as with `OSCALBackMatterProps` in this file. Could we add similar documentation here?

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const { children, uuid } = props;
+  return (
+    <OSCALAnchorLinkHeader value={uuid}>
+      <Typography variant="subtitle1">{children}</Typography>
+    </OSCALAnchorLinkHeader>
+  );
+}
+
+interface CitationDisplayProps {
+  readonly resource: Resource;
```
Bronstrom, line 43: Although there is only a single property in this interface, I would suggest still providing documentation here.

Bronstrom (changes requested):
Overall, this looks great! I love all the refactoring choices made to improve the overall structure of editing and the implementations made with TypeScript.

Although, it appears the JSON editor window is blank. Maybe during the refactoring process some properties were lost and not passed to `OSCALEditableTextField`?

image

Bronstrom:
Also, one other difference I noticed when comparing backmatter resources between this branch and develop, is the `DescriptionIcon` (looks like a document in the right corner) is missing.

Before:

image

After:

image

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
tuckerzp: I was wondering about that, but Resource is a documented field on `oscal-types` so I did not document the field.

tuckerzp:
> Although, it appears the JSON editor window is blank. Maybe during the refactoring process some properties were lost and not passed to OSCALEditableTextField?

I think #776 fixes that. `develop` just has not been merged in. From my understanding, the `EditableTextField` and the `Drawer` do not really interact. Should be good now!

> the DescriptionIcon (looks like a document in the right corner) is missing.

Yeah it did not make sense to edit a little icon button. So `description` is now a text field.

Bronstrom (approved):

PR #776, opened 2023-04-13
fix(editor): JSON drawer was not displaying data

tuckerzp:


kylelaker (approved):
Great catch! This will fix the tests on the all-in-one container stuff.

Issue #718, opened 2023-03-07
Display properties for metadata, back matter, and controls


Bronstrom:
### Description

As a governance, risk, and compliance author, I would like to see a generic display of properties for metadata, backmatter and controls.

### Acceptance Criteria

* The following `props` fields are properly displayed within the React library for Metadata (`metadata/props`), Backmatter (`backmatter/resources/props`), and Controls (`controls/props`):
   * `name` [1]
   * `ns` [0 or 1]
   * `value` [1]
   * `class` [0 or 1]
   * `remarks` [0 or 1]

### Proposed Solution

The data will be displayed in a table in a modal that can be launched by clicking an icon on the field(s).

### Additional Notes

`markings` will be handled separately. For more information see: https://pages.nist.gov/OSCAL/reference/latest/complete/json-reference/#/profile/metadata/props

### Implementation

- [ ] I can (or plan to) submit a pull request to implement this
- [ ] Implementing this may result in a breaking change

brian-easyd:
For `@ns`, properties should be separated into NIST OSCAL core properties and third-party OSCAL extensions. 

Empty ns and the explicit NIST OSAL name spaces are identical and should be listed first.

Other namespaces should be grouped (sorted by @ns first and @name second)

Bronstrom:
I've started to make a basic implementation of a reusable `OSCALProperties` component to set up the generic representation of properties. Here is a quick mock-up of how I plan the design to look:

![OSCAL_properties_support](https://user-images.githubusercontent.com/35196021/233143752-1f0f1e7c-3a4f-45f5-82a5-873db97f7f49.png)

@brian-easyd Do the tables and headings look correct?  I also plan to sort `@ns` by: empty `ns` > NIST `ns` > 3rd party `ns`.

tuckerzp:
Pushing to Sprint 46

brian-easyd:
It is possible that other NIST departments will have their own OSCAL extensions separate from the core syntax, which means it is possible to have other @ns values with "nist.gov" in them that are *not* the default syntax. 

So yes, we should treat no `ns`, empty `ns=""` (I am not sure whether or not this is even valid) and explicit `ns="https://csrc.nist.gov/ns/oscal"` (case sensitive) as equivalent. Any other value for `ns` is an OSCAL extension.

kylelaker:
@Bronstrom I reopened this until we have the story book stories.

tuckerzp:
[...]

PR #786, opened 2023-04-20
feat: add generic display of properties

Bronstrom:
This creates a generic OSCALProperties component and adds it to back matter, controls, and metadata.

Closes https://github.com/EasyDynamics/oscal-react-library/issues/718

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALBackMatter.tsx
+++ b/packages/oscal-react-library/src/components/OSCALBackMatter.tsx
@@ -16,6 +16,7 @@ import OSCALAnchorLinkHeader from "./OSCALAnchorLinkHeader";
 import OSCALEditableTextField, { EditableFieldProps } from "./OSCALEditableTextField";
 import { Resource, ResourceLink, BackMatter } from "@easydynamics/oscal-types";
 import { ReactElement } from "react";
+import OSCALProperties from "./OSCALProperties";
 
 export const OSCALBackMatterCard = styled(Card)(
   ({ theme }) => `
@@ -100,13 +101,12 @@ interface BackMatterResourceProps extends OSCALBackMatterProps {
 }
 
 function BackMatterResource(props: BackMatterResourceProps) {
-  const { resource } = props;
+  const { resource, parentUrl, partialRestData, isEditable, backMatter, onFieldSave } = props;
 
   const getMediaType = (rlink: ResourceLink) =>
-    rlink["media-type"] ||
-    guessExtensionFromHref(getAbsoluteUrl(rlink.href, props.parentUrl) ?? "");
+    rlink["media-type"] || guessExtensionFromHref(getAbsoluteUrl(rlink.href, parentUrl) ?? "");
 
-  const objectKey = getObjectRootKey(props.partialRestData);
+  const objectKey = getObjectRootKey(partialRestData);
 
   return (
     <Grid item xs={3} key={resource.uuid}>
@@ -115,19 +115,23 @@ function BackMatterResource(props: BackMatterResourceProps) {
           <Grid container spacing={0}>
             <Grid item xs={11}>
               <TitleDisplay uuid={resource.uuid}>
+                <OSCALProperties
+                  properties={resource?.props}
+                  title={resource?.title ?? resource?.uuid}
+                />
                 <OSCALEditableTextField
                   fieldName="title"
-                  isEditable={props.isEditable}
-                  editedField={props.isEditable ? [objectKey, "back-matter", "resources"] : null}
-                  editedValue={props.backMatter?.resources}
+                  isEditable={isEditable}
+                  editedField={isEditable ? [objectKey, "back-matter", "resources"] : null}
+                  editedValue={backMatter?.resources}
                   editedValueId={resource.uuid}
-                  onFieldSave={props.onFieldSave}
+                  onFieldSave={onFieldSave}
                   partialRestData={
-                    props.isEditable
+                    isEditable
                       ? {
                           [objectKey]: {
-                            uuid: props.partialRestData?.[objectKey].uuid,
-                            "back-matter": props.backMatter,
+                            uuid: partialRestData?.[objectKey].uuid,
+                            "back-matter": backMatter,
                           },
                         }
                       : null
@@ -144,17 +148,17 @@ function BackMatterResource(props: BackMatterResourceProps) {
           </Grid>
           <OSCALEditableTextField
             fieldName="description"
-            isEditable={props.isEditable}
-            editedField={props.isEditable ? [objectKey, "back-matter", "resources"] : null}
-            editedValue={props.backMatter?.resources}
+            isEditable={isEditable}
+            editedField={isEditable ? [objectKey, "back-matter", "resources"] : null}
+            editedValue={backMatter?.resources}
             editedValueId={resource.uuid}
-            onFieldSave={props.onFieldSave}
+            onFieldSave={onFieldSave}
             partialRestData={
-              props.isEditable
+              isEditable
                 ? {
                     [objectKey]: {
-                      uuid: props.partialRestData?.[objectKey].uuid,
-                      "back-matter": props.backMatter,
+                      uuid: partialRestData?.[objectKey].uuid,
+                      "back-matter": backMatter,
                     },
                   }
                 : null
@@ -169,7 +173,7 @@ function BackMatterResource(props: BackMatterResourceProps) {
                 label={getMediaType(rlink)}
                 component="a"
                 role="button"
-                href={getAbsoluteUrl(rlink.href, props.parentUrl)}
+                href={getAbsoluteUrl(rlink.href, parentUrl)}
                 target="_blank"
                 variant="outlined"
                 clickable
--- a/packages/oscal-react-library/src/components/OSCALControl.js
+++ b/packages/oscal-react-library/src/components/OSCALControl.js
@@ -11,6 +11,7 @@ import OSCALAnchorLinkHeader from "./OSCALAnchorLinkHeader";
 import isWithdrawn from "./oscal-utils/OSCALCatalogUtils";
 import { propWithName } from "./oscal-utils/OSCALPropUtils";
 import { appendToFragmentPrefix } from "./oscal-utils/OSCALLinkUtils";
+import OSCALProperties from "./OSCALProperties";
 
 const OSCALControlCard = styled(Card, {
   // https://github.com/mui/material-ui/blob/c34935814b81870ca325099cdf41a1025a85d4b5/packages/mui-system/src/createStyled.js#L56
@@ -44,6 +45,15 @@ function ControlsList(props) {
   } = props;
   return (
     <div>
+      <OSCALProperties
+        properties={control.props}
+        title={
+          <>
+            <OSCALControlLabel id={control.id} label={control.label} component="span" />
+            {` ${control.title} Discussion`}
+          </>
+        }
+      />
       {control.parts?.map((part, index) => (
         <OSCALControlPart
           componentId={componentId}
--- a/packages/oscal-react-library/src/components/OSCALProperties.tsx
+++ b/packages/oscal-react-library/src/components/OSCALProperties.tsx
@@ -1,0 +1,144 @@
+import HomeIcon from "@mui/icons-material/Home";
+import Button from "@mui/material/Button";
+import Dialog from "@mui/material/Dialog";
+import DialogContent from "@mui/material/DialogContent";
+import DialogTitle from "@mui/material/DialogTitle";
+import Stack from "@mui/material/Stack";
+import { styled } from "@mui/material/styles";
+import React from "react";
+import Table from "@mui/material/Table";
+import TableBody from "@mui/material/TableBody";
+import TableCell from "@mui/material/TableCell";
+import TableContainer from "@mui/material/TableContainer";
+import TableRow from "@mui/material/TableRow";
+import StyledTooltip from "./OSCALStyledTooltip";
+import {
+  OSCALSystemImplementationTableTitle,
+  StyledHeaderTableCell,
+  StyledTableRow,
+  StyledTableHead,
+} from "./OSCALSystemImplementationTableStyles";
+
+const OSCALPropertiesButton = styled(Button)(
+  ({ theme }) => `
+    color: ${theme.palette.primary.main};
+    margin-top: 1em;
+    margin-bottom: 0.5em;
+  `
+);
+
+function OSCALPropertiesCard(props: any) {
+  const { title, children } = props;
+
+  const [open, setOpen] = React.useState(false);
+
+  const handleOpen = () => {
+    setOpen(true);
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  return (
+    <>
+      <OSCALPropertiesButton
+        size="small"
+        variant="outlined"
+        onClick={handleOpen}
+        aria-label={`${title} properties button`}
+        startIcon={<HomeIcon />}
+      >
+        Properties
+      </OSCALPropertiesButton>
+      <Dialog
+        open={open}
+        onClose={handleClose}
+        scroll="paper"
+        aria-labelledby="scroll-dialog-title"
+        aria-describedby="scroll-dialog-description"
+        maxWidth="md"
+        fullWidth
+      >
+        {children}
+      </Dialog>
+    </>
+  );
+}
+
+function isNistNamespace(namespace: string) {
+  return !namespace || namespace.includes("nist.gov");
+}
+
+function getThirdPartyNamespaces(properties: any) {
+  let items: string[] = [""];
+
+  properties
+    ?.filter((property: any) => !isNistNamespace(property?.ns))
+    .map((property: any) => {
+      items.push(property?.ns);
+    });
+
+  return items;
+}
+
+function OSCALPropertiesTable(properties: any, namespace: string) {
+  return (
+    <DialogContent dividers>
+      <OSCALSystemImplementationTableTitle variant="h6" id={`${namespace}-table-title`}>
+        {namespace}
+      </OSCALSystemImplementationTableTitle>
+      <TableContainer sx={{ maxHeight: "25em" }}>
+        <Table aria-label="Components" sx={{ height: "max-content" }}>
+          <StyledTableHead>
+            <TableRow>
+              <StyledHeaderTableCell>Name</StyledHeaderTableCell>
+              <StyledHeaderTableCell>Class</StyledHeaderTableCell>
+              <StyledHeaderTableCell>Value</StyledHeaderTableCell>
+            </TableRow>
+          </StyledTableHead>
+          <TableBody>
+            {properties
+              ?.filter((property: any) =>
+                namespace === "NIST OSCAL"
+                  ? isNistNamespace(property.ns)
+                  : property?.ns === namespace
+              )
+              .map((property: any) => (
+                <StyledTooltip title={property?.remarks ?? ""}>
+                  <StyledTableRow key={property?.uuid}>
+                    <TableCell>{property?.name}</TableCell>
+                    <TableCell>{property?.class}</TableCell>
+                    <TableCell>{property?.value}</TableCell>
+                  </StyledTableRow>
+                </StyledTooltip>
+              ))}
+          </TableBody>
+        </Table>
+      </TableContainer>
+    </DialogContent>
+  );
+}
+
+export default function OSCALProperties(props: any) {
+  const { properties, title } = props;
+  const thirdPartyNamespaces = getThirdPartyNamespaces(properties);
+
+  return properties ? (
+    <OSCALPropertiesCard title={title}>
+      <DialogTitle id="scroll-dialog-title">
+        <Stack direction="column">{title} Properties</Stack>
+      </DialogTitle>
+      {/* Handle NIST properties */}
+      <OSCALPropertiesTable properties={properties} namespace={"NIST OSCAL"} />
+      {
+        /* Handle 3rd party properties */
+        thirdPartyNamespaces
+          ?.filter((namespace) => namespace !== "")
+          .map((namespace: any) => (
+            <OSCALPropertiesTable properties={properties} namespace={namespace} />
+          ))
+      }
+    </OSCALPropertiesCard>
+  ) : null;
+}
--- a/packages/oscal-react-library/src/components/OSCALMetadata.js
+++ b/packages/oscal-react-library/src/components/OSCALMetadata.js
@@ -39,6 +39,7 @@ import { propWithName } from "./oscal-utils/OSCALPropUtils";
 import OSCALEditableTextField from "./OSCALEditableTextField";
 import OSCALAnchorLinkHeader from "./OSCALAnchorLinkHeader";
 import { OSCALMarkupLine, OSCALMarkupMultiLine } from "./OSCALMarkupProse";
+import OSCALProperties from "./OSCALProperties";
 
 const OSCALMetadataSectionInfoHeader = styled(Typography)`
   display: flex;
@@ -644,6 +645,9 @@ export default function OSCALMetadata(props) {
                 onFieldSave={props.onFieldSave}
               />
             </OSCALMetadataSection>
+            <OSCALMetadataSection>
+              <OSCALProperties properties={props.metadata.props} title={props.metadata.title} />
+            </ OSCALMetadataSection>
             <OSCALMarkupMultiLine>{props.metadata.remarks}</OSCALMarkupMultiLine>
             <OSCALMetadataKeywords
               keywords={propWithName(props.metadata.props, "keywords")?.value}


packages/oscal-react-library/src/components/OSCALControl.js
```diff
@@ -44,6 +45,15 @@ function ControlsList(props) {
   } = props;
   return (
     <div>
+      <OSCALProperties
+        properties={control.props}
+        title={
+          <>
+            <OSCALControlLabel id={control.id} label={control.label} component="span" />
+            {` ${control.title} Discussion`}
```
kylelaker, line 53:
```suggestion
            {` ${control.title} Properties`}
```

?

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,146 @@
...
+      >
+        {children}
+      </Dialog>
+    </>
+  );
+}
+
+function isNistNamespace(namespace: string) {
+  return !namespace || namespace.includes("nist.gov");
+}
```
kylelaker, line 71: Consider trying to use something from `OSCALPropUtils` (exporting something that exists or adding a new function)

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,146 @@
...
+    ?.filter((property: any) => !isNistNamespace(property?.ns))
+    .map((property: any) => {
+      items.push(property?.ns);
+    });
+
+  return items;
+}
+
+function OSCALPropertiesTable(props: any) {
+  const { properties, namespace } = props;
```
kylelaker, line 86:
```suggestion

interface OSCALPropertiesTableProps {
  properties: Property[]; // Import from "@easydynamics/oscal-types"
  namespace?: string;
}

function OSCALPropertiesTable(props: any) {
  const { properties, namespace } = props;
```

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <TableRow>
+              <StyledHeaderTableCell>Name</StyledHeaderTableCell>
+              <StyledHeaderTableCell>Class</StyledHeaderTableCell>
+              <StyledHeaderTableCell>Value</StyledHeaderTableCell>
+            </TableRow>
+          </StyledTableHead>
+          <TableBody>
+            {properties
+              ?.filter((property: any) =>
+                namespace === "NIST OSCAL"
```
kylelaker, line 105: Make this a constant if we're going to reuse it -- especially bc i don't love this value but I can't think of something better right now.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <Stack direction="column">{title} Properties</Stack>
+      </DialogTitle>
+      {/* Handle NIST properties */}
+      <OSCALPropertiesTable properties={properties} namespace={"NIST OSCAL"} />
+      {
+        /* Handle 3rd party properties */
+        thirdPartyNamespaces
+          ?.filter((namespace) => namespace !== "")
+          .map((namespace: any) => (
+            <OSCALPropertiesTable properties={properties} namespace={namespace} />
```
kylelaker, line 141:
```suggestion
            <OSCALPropertiesTable properties={properties} namespace={namespace} key={namespace} />
```

kylelaker (commented):
Very early thoughts. There are a few design things not written here that we should talk through. I will try to summarize those and provide more info soon.

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: This is a little different than what is in `OSCALPropUtils`, more of a validation check rather than providing a NIST namespace. I've gone ahead and added this to `OSCALPropUtils`.

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: I agree, my plan was to check-in on what our NIST namespace should be called.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * Title of the element the properties resides under
+   */
+  title: string;
+  /**
+   * Children to display in the modal
+   */
+  children: React.ReactNode;
+}
+
+const OSCALPropertiesDialog = (props: OSCALPropertiesDialogProps): ReactElement => {
```
kylelaker, line 42:
```suggestion
const OSCALPropertiesDialog: React.FC<OSCALPropertiesDialogProps> = (props) => {
```

packages/oscal-react-library/src/components/oscal-utils/OSCALPropUtils.ts
```diff
@@ -5,6 +5,7 @@ import type { Property } from "@easydynamics/oscal-types";
  * namespace is default namespace.
  */
 const NIST_DEFAULT_NAMESPACE = "https://csrc.nist.gov/ns/oscal";
+export const NIST_NAMESPACE_DOMAIN = "nist.gov";
```
kylelaker, line 8: I don't want to make assumptions based on domain name -- let's remove this

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <OSCALPropertiesTable properties={properties} namespace={NIST_NAMESPACE_DOMAIN} key={NIST_NAMESPACE_DOMAIN} />
+      {
+        /* Handle 3rd party properties */
+        thirdPartyNamespaces
+          ?.filter((namespace) => namespace !== "")
+          .map((namespace: any) => (
+            <OSCALPropertiesTable properties={properties} namespace={namespace} key={namespace} />
+          ))
+      }
+    </OSCALPropertiesDialog>
```
kylelaker, line 183: Let's try to sort to put the `https://csrc.nist.gov/ns/oscal` namespace at the top; however, otherwise it should be treated the exact same as other namespaces. Including using the _value_ (not some other string) in the display.

tuckerzp (approved):
Just had a quick look over things as they are. I had a few comments. 

I also am not sure I understand why `namespaceOf` can't be used instead of `isNistNamespace`. I didn't have a ton of time to think it over, so maybe I am missing something.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * Contains properties elements
+   */
+  properties: any;
+  /**
+   * Title of the element the properties resides under
+   */
+  title: string;
+}
+
+export const OSCALProperties = (props: OSCALPropertiesProps) => {
```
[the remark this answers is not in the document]
tuckerzp: To clarify a bit, I think it is best to make things that are objects from OSCAL (like Properties) be the simplest possible component so in theory it can be used elsewhere.

Bronstrom:
> Just had a quick look over things as they are. I had a few comments.
> 
> I also am not sure I understand why `namespaceOf` can't be used instead of `isNistNamespace`. I didn't have a ton of time to think it over, so maybe I am missing something.

`namespaceOf` returns a namespace or the default NIST namespace when there is no namespace provided. `isNistNamespace` validates the namespace being a NIST namespace, which is true when `ns` is empty or an empty string. This validation check is done a few times within `OSCALProperties` and I thought it made sense to be a function. This could be turned into `namespaceOf(ns) === NIST_DEFAULT_NAMESPACE` if we wanted to use that function inside of this one.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,185 @@
...
+   */
+  namespace?: string;
+}
+
+const OSCALPropertiesTable = (props: OSCALPropertiesTableProps): ReactElement => {
+  const { properties, namespace } = props;
+
+  return (
+    <>
+      <OSCALSystemImplementationTableTitle variant="h6" id={`${namespace}-table-title`}>
```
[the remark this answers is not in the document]
Bronstrom: I agree, I might hold off on this to the Global Styles implementation. Unless this feel urgent.

packages/oscal-react-library/src/components/OSCALProperties.tsx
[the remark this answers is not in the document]
Bronstrom: This may make more sense now with there no longer being a default export, which would have followed the naming scheme of the file. Although, what we should also think about is the sheer amount of data that will appear on screen to avoid overwhelming the user - the test catalog I've attached above may give a better idea of why we would want to commonly separate properties into a modal rather than displaying long screens of these properties (especially if there are many 3rd party namespaces in a `props` list). Are there any places in particular you would picture the table existing over just the button?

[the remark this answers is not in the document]
tuckerzp: I don't see a use of it currently, but I think it would be annyoing to switch this around if there ever is a use case. 

Take [OSCAL Revisions](https://github.com/EasyDynamics/oscal-react-library/blob/develop/packages/oscal-react-library/src/components/OSCALRevision.tsx) as an example. Today, we do not actually use the `OSCALRevisions` component outside of `OSCALRevisionsButton`. I do not think the dialog button necessarily encapsulates a list of revisions, it just allows for you to view Revisons. 

I would love to see a similar pattern here. I think making a very basic component that we can then provide wrappers to is a more maintainable pattern overall. 

I was thinking you could do something like:

- `OSCALPropertiesDialog` as the dialog that opens the properties table
- `OSCALProperties` as a table
- `OSCALProperty` as a row within that table

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  properties
+    ?.filter((property: any) => !isNistNamespace(property?.ns))
+    .map((property: any) => {
+      if (!items.includes(property?.ns)) {
+        items.push(property?.ns);
+      }
+    });
+
+  return items;
```
kylelaker, line 49:
```suggestion
  return [...new Set(
    properties
      ?.filter((prop) => namespaceOf(property.ns) !== NIST_DEFAULT_NAMESPACE)
      .map((prop) => prop.ns)
      .filter((prop): prop is string => !!prop.ns)
  )]
```

packages/oscal-react-library/src/components/oscal-utils/OSCALPropUtils.ts
```diff
@@ -16,6 +16,16 @@ export function namespaceOf(ns: string | undefined): string {
...
+/**
+ * Validates if the namespace is a NIST namespace.
+ *
+ * @param namespace the namespace
+ * @returns a boolean describing whether the namespace is
+ */
+export function isNistNamespace(ns: string | undefined): boolean {
+  return !ns || ns === NIST_DEFAULT_NAMESPACE;
+}
+
```
kylelaker, line 28:
```suggestion
```

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,188 @@
...
+          <TableBody>
+            {properties
+              ?.filter((property: any) =>
+                namespace === NIST_DEFAULT_NAMESPACE
+                  ? isNistNamespace(property?.ns)
+                  : property?.ns === namespace
+              )
+              .map((property: any) => (
+                <OSCALProperty property={property} />
+              ))}
```
kylelaker, line 111:
```suggestion
              ?.filter((property) =>
                namespace === NIST_DEFAULT_NAMESPACE
                  ? isNistNamespace(property?.ns)
                  : property?.ns === namespace
              )
              .map((property) => (
                <OSCALProperty property={property} />
              ))}
```

tuckerzp:
I discussed this briefly with @Bronstrom, but I wonder if filtering out non nist namespaces to a separate list to only then put it in more or less the same list is the best approach.

To me, this feels more like we need to `sort`. Anything that is considered to be in the default namespace (found preferably using `namespaceOf` and a boolean check) would be put at the top of the list while everything else can just remain in the same order. 

I honestly am unsure if this cleans up the code a ton, but conceptually it makes more sense to me in terms of what we are actually trying to do. I also think it will make the return statement of the component a lot cleaner.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -30,21 +30,40 @@ const OSCALPropertiesButton = styled(Button)(
 );
 
 /**
- * Accumulates all instances of third-party namespaces into a list.
+ * Sorts the namespaces by NIST namespaces first, orders third-party namespaces alphabetically,
+ * then orders by name.
  *
  * @param {Property[]} properties A list of properties
- * @returns {string[]} A list of third-party namespaces
+ * @returns {Property[] | undefined} A list of third-party namespaces
```
kylelaker: Just move the types to TS and remove from JSDoc

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -30,21 +30,40 @@ const OSCALPropertiesButton = styled(Button)(
...
-    .map((property: any) => {
-      if (!items.includes(property?.ns)) {
-        items.push(property?.ns);
-      }
-    });
+  properties?.map((property: any) => {
+    if (!items.includes(namespaceOf(property?.ns))) {
+      items.push(namespaceOf(property?.ns));
+    }
+  });
```
kylelaker, line 66: If you're going to do this over the suggestion I had made, then this needs to be `forEach` not a `map`. You're `map`ing to void/undefined. Which is very ungood.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -101,11 +120,7 @@ const OSCALProperties = (props: OSCALPropertiesProps): ReactElement => {
           </StyledTableHead>
           <TableBody>
             {properties
-              ?.filter((property: any) =>
-                namespace === NIST_DEFAULT_NAMESPACE
-                  ? isNistNamespace(property?.ns)
-                  : property?.ns === namespace
-              )
+              ?.filter((property: any) => namespaceOf(property?.ns) === namespace)
```
kylelaker, line 123: I do not want any `: any`s as part of this -- we know these are `Property`s and we should make sure that is able to inferred in all these places. If it can't we're doing something wrong somewhere else.

There are a lot of these so I am just going to comment it here instead of everywhere.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -30,21 +30,40 @@ const OSCALPropertiesButton = styled(Button)(
...
+    namespaceOf(propA.ns) > namespaceOf(propB.ns) ? 1 : -1
+  );
+  const nistNamespaceFirstProperties = namespaceSortedProperties?.sort((propA, propB) =>
+    isNistNamespace(propA.ns) && !isNistNamespace(propB.ns) ? 1 : -1
+  );
+  const nameSortedProperties = nistNamespaceFirstProperties?.sort((propA, propB) =>
+    namespaceOf(propA.ns) === namespaceOf(propB.ns) && propA?.name > propB?.name ? 1 : -1
+  );
+
+  return nameSortedProperties;
```
kylelaker, line 50: This is... complicated. And I still _really_ do not like `isNistNamespace`'s existence.

Bronstrom:
Yesterday, @tuckerzp and I discussed making some the `filter`s into `sort`s. `sort`s  have proven to look a lot cleaner than using `filter`s, which really should be used if we're removing some of the data rather than ordering it. I've turned the third parties namespace function into one that captures all namespaces in a list, in addition to creating a function to sort properties in the correct order. This ordering is first by the NIST namespace, then 3rd Party extension namespaces, then by name. Three sorts seemed to achieve this well and have been split apart into separate constants for readability.

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: I thought that helped with readability of the sorts. Do you have a better suggestion? Also, didn't realize you don't like `isNistNamespace` - thought it would be helpful to create a reusable function if we're making the validation more than once.

kylelaker:
> `sort`s have proven to look a lot cleaner than using `filter`s, which really should be used if we're removing some of the data rather than ordering it

I think this is a matter of preference not proof 🤣 I think the chained sorting that is required adds complexity and definitely means that we have to test the heck out of that sorting code.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,179 @@
...
+            properties={sortProperties(nist)}
+            namespace={NIST_DEFAULT_NAMESPACE}
+            key={NIST_DEFAULT_NAMESPACE}
+          />
+          {
+            /* Handle 3rd party properties */
+            Object.entries(thirdParties)
+              .sort(([key1, props1], [key2, props2]) => key1.localeCompare(key2))
+              .map(([key, props]) => (
+                <OSCALProperties properties={sortProperties(props)} namespace={key} key={key} />
```
kylelaker, line 172: If it's going to be broken into a different function I'd rather see this as

```ts

function byName(a: Property, b: Property) {
  return a.name.localeCompare(b.name);
}

<OSCALProperties properties={props.sort(byName)} namespace={key} key={key} />
```

This way it's more scriptive _how_ they're being sorted.

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
```diff
@@ -215,6 +215,10 @@ function BackMatterResource(props: BackMatterResourceProps) {
           action={<CitationDisplay resource={resource} />}
         />
         <CardContent>
+          <OSCALPropertiesDialog
+            properties={resource?.props}
+            title={resource?.title ?? resource?.uuid}
```
kylelaker, line 220: Let's not use a UUID in the title... I'd rather see us do something like `<NotSpecifiedTypography component="span">Resource</NotSpecifiedTypography>`

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+interface OSCALPropertyProps {
+  /**
+   * Contains property elements
+   */
+  property: Property;
+}
+
+const OSCALProperty = (props: OSCALPropertyProps): ReactElement => {
+  const { property } = props;
```
tuckerzp, line 52: Can we update all the components in this file to do this?

```suggestion
const OSCALProperty: React.FC<OSCALPropertyProps> = ({ property }) => { 
```

packages/oscal-react-library/src/components/OSCALControl.js
```diff
@@ -44,6 +45,15 @@ function ControlsList(props) {
...
     <div>
+      <OSCALPropertiesDialog
+        properties={control.props}
+        title={
+          <>
+            <OSCALControlLabel id={control.id} label={control.label} component="span" />
+            {` ${control.title}`}
+          </>
+        }
+      />
```
tuckerzp, line 56: Is there a cleaner way to deal with this? Is there a reason each component using an `OSCALPropertiesDialog` would need to create there own title? Would having this just be the `children` be a cleaner solution?

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
Bronstrom: I added a fallback title within `Properties`, maybe we would want to make it more generic within Properties. Maybe it could just be "Not Specified Properties"?

packages/oscal-react-library/src/components/OSCALControl.js
tuckerzp: I personally would like if we stopped putting huge components as a prop. At the very least lets do something like

```TypeScript
const controlLabel =  (
  <>
    <OSCALControlLabel id={control.id} label={control.label} component="span" />
      {` ${control.title}`}
  </>
);

return (
  <OSCALPropertiesDialog title={controlLabel />
);

File: packages/oscal-react-library/src/components/OSCALCatalogGroups.test.tsx [hunk 22]
```diff
@@ -52,6 +52,73 @@ const testGroups: ControlGroup[] = [
   },
 ];
 
+const testGroupswithProps: ControlGroup[] = [
+  {
```
Bronstrom, line 56: Not sure how I feel about this test data here (including the test data before this one). Maybe this could be moved within `test-data`?

Bronstrom:
Here's some screenshots of how the modal is looking:

generic_properties_modal_1

generic_properties_modal_2

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,179 @@
...
+  return (
+    <StyledTooltip title={property.remarks ?? ""}>
+      <StyledTableRow key={property.uuid}>
+        <TableCell>{property.name ?? NO_INFORMATION}</TableCell>
+        <TableCell>{property.class ?? NO_INFORMATION}</TableCell>
+        <TableCell>{property.value ?? NO_INFORMATION}</TableCell>
+      </StyledTableRow>
+    </StyledTooltip>
+  );
+};
```
kylelaker, line 63: I am concerned that our only option to represent a single Property is to have it as a table row... I suppose it's probably fine because I can't think of another place where we may need it...

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,179 @@
...
+              <StyledHeaderTableCell>Name</StyledHeaderTableCell>
+              <StyledHeaderTableCell>Class</StyledHeaderTableCell>
+              <StyledHeaderTableCell>Value</StyledHeaderTableCell>
+            </TableRow>
+          </StyledTableHead>
+          <TableBody>
+            {properties
+              ?.filter((property) => namespaceOf(property?.ns) === namespace)
+              .map((property) => (
+                <OSCALProperty property={property} key={`$namespaceOf(property?.ns)-properties`} />
```
kylelaker, line 95:
```suggestion
                <OSCALProperty property={property} key={`${namespaceOf(property?.ns)}-properties`} />
```

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const [open, setOpen] = React.useState(false);
+
+  if (!properties) {
+    return null;
+  }
+  const namespaceToProps: Record<string, Property[]> = {};
+  for (const prop of properties) {
+    namespaceToProps[namespaceOf(prop.ns)] ??= [];
+    namespaceToProps[namespaceOf(prop.ns)].push(prop);
+  }
```
kylelaker, line 128:
```suggestion
  for (const prop of properties) {
    const ns = namespaceOf(prop.ns);
    namespaceToProps[ns] ??= [];
    namespaceToProps[ns].push(prop);
  }
```

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,179 @@
...
+            properties={nist?.sort(byName)}
+            namespace={NIST_DEFAULT_NAMESPACE}
+            key={NIST_DEFAULT_NAMESPACE}
+          />
+          {
+            /* Handle 3rd party properties */
+            Object.entries(thirdParties)
+              .sort(([key1], [key2]) => key1.localeCompare(key2))
+              .map(([key, props]) => (
+                <OSCALProperties properties={props?.sort(byName)} namespace={key} key={key} />
```
kylelaker, line 172:
```suggestion
                <OSCALProperties properties={props.sort(byName)} namespace={key} key={key} />
```

packages/oscal-react-library/src/components/OSCALCatalogGroups.test.tsx
kylelaker: That's fine. I really dislike `test-data` but it exists and there's not much to do about that. If you move it, keep it in a TSX file.

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: Good catch!

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: From looking at the many places in the spec where properties appear, it seems like they appear in a very similar manner, which makes sense to have properties appear as a row in a table. Although, we could instead call this `OSCALPropertyRow` instead if that makes more sense?

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: Several of the tests will fair when I remove either this optional chaining operator or the one in `nist?.sort(byName)`

packages/oscal-react-library/src/components/OSCALProperties.tsx
kylelaker: I guess my thought is that it shouldn't be a separate component -- it cannot be displayed outside `OSCALProperties` which is a table; so it's very dependent on that representation. But I guess since it's not exported, then it's likely fine.

packages/oscal-react-library/src/components/OSCALBackMatter.tsx
Bronstrom: Sorry, I should have mentioned I corrected this to what you were suggesting (`<NotSpecifiedTypography component="span">Resource</NotSpecifiedTypography>`)

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: I'm wrong. It is possible to have properties defined and only NIST properties, or have properties defined and only 3rd party properties. `nist?.sort(byName)` must have the optional chaining operator, but it should be fine to remove it from the place you mentioned.  With the 3rd party properties filter, it will handle if there are no 3rd parties defined, so I should be able to remove the optional chain operator here: `props?.sort(byName)`. Although the type declaration should also account for there being no NIST properties present - which is a case where there would be passed no properties passed to `OSCALProperties`.

packages/oscal-react-library/src/components/OSCALCatalogGroups.test.tsx
Bronstrom: I think I'll just plan to leave it where it is. Would it make sense to move more of our data to the file where we are testing it in the future?

tuckerzp: I personally think having a huge list of "common" objects that we continuously add to makes testing a bit tricky. I think that testing the smallest unit possible with very minimum data makes our tests more understandable and makes it easier to test various edge cases. 

At the very least, we need to stop the pattern of missing data for tests, adding it to the already huge objects in `test-data` and making our tests more complex.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,183 @@
...
+  };
+
+  return (
+    <>
+      <OSCALPropertiesButton
+        size="small"
+        variant="outlined"
+        onClick={handleOpen}
+        aria-label={`${title} properties button`}
+        startIcon={<HomeIcon />}
```
kylelaker, line 149: Instead of `Home` let's do https://mui.com/material-ui/material-icons/?selected=Construction

And then we can set 
```
title="Open Properties"
```

or something like that so that we can get _something_ on hover?

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,183 @@
...
+    setOpen(false);
+  };
+
+  return (
+    <>
+      <OSCALPropertiesButton
+        size="small"
+        variant="outlined"
+        onClick={handleOpen}
+        aria-label={`${title} properties button`}
```
kylelaker, line 148: `title` may not be a string which won't work well in an `aria-label`. Also, I am not sure that "button" is appropriate here... We may want to use something like the other suggestion for the `title` attribute.

kylelaker (commented):
Small iconography change and then I think I am good to move forward with merging.

packages/oscal-react-library/src/components/OSCALProperties.tsx
```diff
@@ -0,0 +1,183 @@
...
+                />
+              ))}
+          </TableBody>
+        </Table>
+      </TableContainer>
+    </>
+  );
+};
+
+interface OSCALPropertiesDialogProps {
```
kylelaker, line 107:
```suggestion
export interface OSCALPropertiesDialogProps {
```

kylelaker (commented):
Oh! We also need a storybook file for this! At least for the `OSCALProperties` dialog itself

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: I much more favor that icon for the button instead. I was thinking more real estate in terms of property haha

`title` holds the control name, resource title, or name of document (for metadata) currently. I think this is useful for the modal to contain this info at the very least. Are you suggesting the button have the title "Open Properties"?

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: Good point!

packages/oscal-react-library/src/components/OSCALProperties.tsx
Bronstrom: I could just make this "open-properties" or should I be more specific? I was using `title` to differentiate them a little more, but I understand why that wouldn't work well.

packages/oscal-react-library/src/components/OSCALProperties.tsx
kylelaker: I am asking that the Button's `title` and `aria-label` attributes be set to "Open Properties"

kylelaker (approved):
Thanks for your work on this!

tuckerzp (approved):

PR #788, opened 2023-04-20
chore(metadata): extract metadata child components

tuckerzp:
This extracts Locations, Parties, & Roles to seperate files.
Hopefully this will help make Metadata a little more manageable.

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALLocations.tsx
+++ b/packages/oscal-react-library/src/components/OSCALLocations.tsx
@@ -29,7 +29,7 @@ import { OSCALMarkupLine } from "./OSCALMarkupProse";
 
 export interface OSCALMetadataLocationsProps {
   locations: Location[] | undefined;
-  urlFragment?: string | undefined;
+  urlFragment?: string;
 }
 
 export const OSCALMetadataLocations: React.FC<OSCALMetadataLocationsProps> = (props) => {
--- a/packages/oscal-react-library/src/components/OSCALLocations.tsx
+++ b/packages/oscal-react-library/src/components/OSCALLocations.tsx
@@ -1,0 +1,135 @@
+import { Location } from "@easydynamics/oscal-types";
+import {
+  DialogTitle,
+  Stack,
+  Grid,
+  Card,
+  Link,
+  Typography,
+  List,
+  DialogContent,
+} from "@mui/material";
+import React from "react";
+import {
+  OSCALMetadataContactTypeHeader,
+  TextWithIcon,
+  OSCALMetadataFieldArea,
+  OSCALMetadataAddress,
+  MetadataInfoList,
+  MetadataInfoType,
+  OSCALMetadataCard,
+  MetadataAvatar,
+} from "./OSCALMetadataCommon";
+import OpenInNewIcon from "@mui/icons-material/OpenInNew";
+import MapIcon from "@mui/icons-material/Map";
+import PlaceIcon from "@mui/icons-material/Place";
+import PhoneIcon from "@mui/icons-material/Phone";
+import EmailIcon from "@mui/icons-material/Email";
+import { OSCALMarkupLine } from "./OSCALMarkupProse";
+
+export interface OSCALMetadataLocationsProps {
+  locations: Location[] | undefined;
+  urlFragment?: string | undefined;
+}
+
+export const OSCALMetadataLocations: React.FC<OSCALMetadataLocationsProps> = (props) => {
+  const { locations, urlFragment } = props;
+
+  const cards = locations?.map((location) => (
+    <Grid item xs={12} md={4} key={location.uuid} component={Card}>
+      <OSCALMetadataLocation location={location} />
+    </Grid>
+  ));
+
+  return (
+    <OSCALMetadataFieldArea title="Locations" urlFragment={urlFragment}>
+      {cards}
+    </OSCALMetadataFieldArea>
+  );
+};
+
+interface OSCALMetadataLocationUrlsProps {
+  /**
+   * List of urls for a given location.
+   */
+  urls: string[] | undefined;
+}
+
+const OSCALMetadataLocationUrls: React.FC<OSCALMetadataLocationUrlsProps> = (props) => {
+  const { urls } = props;
+  return (
+    <Stack>
+      <OSCALMetadataContactTypeHeader title="URLs" />
+      {urls?.length ? (
+        urls.map((url) => (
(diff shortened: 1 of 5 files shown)


tuckerzp:
I left out keywords for now as that will likely need to be reworked at some point anyways.

packages/oscal-react-library/src/components/OSCALLocations.tsx
```diff
@@ -0,0 +1,135 @@
...
+import OpenInNewIcon from "@mui/icons-material/OpenInNew";
+import MapIcon from "@mui/icons-material/Map";
+import PlaceIcon from "@mui/icons-material/Place";
+import PhoneIcon from "@mui/icons-material/Phone";
+import EmailIcon from "@mui/icons-material/Email";
+import { OSCALMarkupLine } from "./OSCALMarkupProse";
+
+export interface OSCALMetadataLocationsProps {
+  locations: Location[] | undefined;
+  urlFragment?: string | undefined;
```
kylelaker, line 32:
```suggestion
  urlFragment?: string;
```

packages/oscal-react-library/src/components/OSCALMetadataCommon.tsx
```diff
@@ -1,12 +1,405 @@
...
+  Button,
+} from "@mui/material";
+import Accordion from "@mui/material/Accordion";
+import AccordionDetails from "@mui/material/AccordionDetails";
+import AccordionSummary from "@mui/material/AccordionSummary";
+import Grid from "@mui/material/Grid";
+import Link from "@mui/material/Link";
 import { styled } from "@mui/material/styles";
-import { Typography } from "@mui/material";
+import { default as React, ReactNode, useEffect } from "react";
```
kylelaker, line 25:
```suggestion
import React { ReactNode, useEffect } from "react";
```

packages/oscal-react-library/src/components/OSCALMetadataCommon.tsx
```diff
@@ -1,12 +1,405 @@
...
+   */
+  id: string;
+  /**
+   * Text to appear on avatar
+   */
+  text: string | undefined;
+  /**
+   * Icon to use if text is undefined
+   */
+  fallbackIcon: ReactNode;
```
kylelaker, line 60: Can you see if there's some MUI type that we can use here for an icon type?

packages/oscal-react-library/src/components/OSCALMetadataCommon.tsx
```diff
@@ -1,12 +1,405 @@
...
+  mobile: <SmartphoneIcon />,
+  office: <BusinessIcon />,
+};
+
+const ADDRESS_ICON_MAPPING = {
+  home: <HomeIcon />,
+  work: <BusinessIcon />,
+};
+
+const UNKNOWN_TYPE_ICON = <HelpOutlineIcon />;
```
kylelaker, line 158: Maybe move to the top?

packages/oscal-react-library/src/components/OSCALParty.tsx
```diff
@@ -0,0 +1,121 @@
...
+    props.party?.type === PartyType.ORGANIZATION ? <GroupIcon /> : <PersonIcon />;
+
+  const avatar = (
+    <MetadataAvatar id={props.party.uuid} text={props.party.name} fallbackIcon={fallbackIcon} />
+  );
+
+  return (
+    <OSCALMetadataCard
+      title={props.party.name}
+      subheader={props.partyRolesText?.map((role: Role) => role.title).join(", ")}
```
kylelaker, line 39:
```suggestion
      subheader={props.partyRolesText?.map((role) => role.title).join(", ")}
```

`tsc` should be able to infer the type here

packages/oscal-react-library/src/components/OSCALMetadataCommon.tsx
tuckerzp: I looked around and couldn't find anything too useful.

packages/oscal-react-library/src/components/OSCALMetadataCommon.tsx
```diff
@@ -45,6 +45,37 @@ const OSCALMetadataSectionInfoHeader = styled(Typography)`
...
+
+function getIconOfType(
+  mapping: Record<string, ReactNode> | undefined,
```
kylelaker, line 62: I think you can be more specific and use `SvgIcon`?

https://github.com/mui/material-ui/blob/master/packages/mui-material/src/utils/createSvgIcon.d.ts

kylelaker: Maybe with a `typeof` in here somewhere?

packages/oscal-react-library/src/components/OSCALMetadataCommon.tsx
```diff
@@ -1,12 +1,405 @@
...
+  }
+
+  return (
+    <>
+      {list.map((item, index) => {
+        return <ListItem key={index}>{mapContactType(item, infoType)}</ListItem>;
+      })}
+    </>
+  );
+};
```
kylelaker, line 282: I think I am starting to realize that this is a bit overly generic to expose as an interface in the library. And it requires passing two tightly coupled properties with the same value.


I'd rather see us do:

```tsx
export type Email = string;

interface LocationDisplayProps<T> {
  contact: T;
}

type LocationDisplay<T> = React.FC<LocationDisplayProps<T>>;

const OSCALMetadataAddress: LocationDisplay<Address> = ({ contact }) => {
  // The same code as now, just using `contact` instead of address
};
const OSCALMetadataEmail: LocationDisplay<Email> = ({ contact }) => {
};
const OSCALMetadataTelephoneNumber: LocationDisplay<TelephoneNumber> = ({ contact }) => {
};

interface MetadataLocationListProps<T> {
  list: T[] | undefined; // I do not like the name list.
  component: LocationDispay<T>;
  emptyMessage: string;
}

const MetadataLocationList<T>: React.FC<MetadataInfoListProps<T>> = (props) => {
  const { list, emptyMessage } = props;
  const LocationDisplayComponent = props.component;

  if (!list?.length) {
    return <NotSpecifiedTypography>{emptyMessage}</NotSpecifiedTypography>
  }

  return (
    <>
      {list.map((item, index) => {
        <ListItem key={index}><LocationDisplayComponent contact={item} /></ListItem>
      })}
    </>
  );
};
```

kylelaker:
There are too many conflicts and changes within Metadata at this point and too many remaining review comments.

PR #792, opened 2023-04-21
fix(catalog): catalog viewer only shows controls if they are under a group

### LINKED ISSUE #741: Catalog viewer only shows controls if they are under a group
### Description

The OSCAL specification allows controls to be defined as a list with a group (which the Viewer currently supports) and also at the root level of the Catalog (which the Viewer does not currently support).

### Environment

oscal-viewer.msd.easydynamics.com

### Steps to Reproduce

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/test-content/catalogs/basic-test-catalog.json

### Expected behavior

The controls in `/controls` should be displayed.

### Actual Behavior

Only items in `/groups/controls` are displayed.

### Additional Notes

An easy solution may be to have a "default" group where these controls are shown and the "Control Groups" can be renamed to to "Controls". Otherwise, a separate section for these "root"-level controls will need to be created.

### LINKED ISSUE #742: Prose is not displayed for groups in a catalog
### Description

Groups can have `parts` which have `prose`. This prose is not currently displayed in the viewer application but it should be, perhaps above the controls on the right-hand side of the tab selector?

### Environment

oscal-viewer.msd.easydynamics.com

### Steps to Reproduce

*Sample catalog link will be provided*

### Expected behavior

The prose for a group should be displayed, both for `/groups` and `//groups`.

### Actual Behavior

No prose is displayed for groups. In fact, it seems like `parts` is never considered at all.

### Additional Notes

_No response_

tuckerzp:
Previously OSCALCatalog ignored the `Controls` prop it was passed.
This creates an OSCALControls component to display controls outside
a group.

closes #741 
closes #742

### Testing

**No controls**

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/main/catalogs/NIST_SP-800-53_rev4_catalog.json

**Controls + Groups**

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/group-prose/catalogs/basic-test-catalog.json

**Just Controls**

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/group-prose/catalogs/basic-test-catalog-no-groups.json

---

Groups without controls is displayed the same

image

---

`Top Controls` is now shown as a group

image

---

Top level control groups with `prose` is displayed

image

---

Children groups with `prose` is displayed

image

---

No groups listed

image

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALCatalog.tsx
+++ b/packages/oscal-react-library/src/components/OSCALCatalog.tsx
@@ -1,10 +1,19 @@
+import { Catalog } from "@easydynamics/oscal-types";
 import React, { useEffect } from "react";
 import OSCALBackMatter from "./OSCALBackMatter";
 import OSCALCatalogGroups from "./OSCALCatalogGroups";
+import { EditableFieldProps } from "./OSCALEditableTextField";
 import { OSCALDocumentRoot } from "./OSCALLoaderStyles";
 import OSCALMetadata from "./OSCALMetadata";
 
-export default function OSCALCatalog(props) {
+export interface OSCALCatalogProps extends EditableFieldProps {
+  onResolutionComplete: React.EffectCallback;
+  catalog: Catalog;
+  urlFragment?: string | undefined;
+  parentUrl: string;
+}
+
+export const OSCALCatalog: React.FC<OSCALCatalogProps> = (props) => {
   const { onResolutionComplete, catalog, isEditable, onFieldSave, urlFragment, parentUrl } = props;
 
   useEffect(onResolutionComplete);
@@ -36,4 +45,6 @@ export default function OSCALCatalog(props) {
       />
     </OSCALDocumentRoot>
   );
-}
+};
+
+export default OSCALCatalog;
--- a/packages/oscal-react-library/src/components/OSCALCatalog.tsx
+++ b/packages/oscal-react-library/src/components/OSCALCatalog.tsx
@@ -1,12 +1,54 @@
 import { Catalog, ControlGroup } from "@easydynamics/oscal-types";
+import { Card, CardContent } from "@mui/material";
 import React, { useEffect } from "react";
+import { OSCALSection, OSCALSectionHeader } from "../styles/CommonPageStyles";
+import OSCALAnchorLinkHeader from "./OSCALAnchorLinkHeader";
 import OSCALBackMatter from "./OSCALBackMatter";
+import { OSCALCatalogControlListItem } from "./OSCALCatalogGroup";
 import OSCALCatalogGroups from "./OSCALCatalogGroups";
 import { EditableFieldProps } from "./OSCALEditableTextField";
 import { OSCALDocumentRoot } from "./OSCALLoaderStyles";
 import OSCALMetadata from "./OSCALMetadata";
 
-const UNGROUPED_CONTROLS_TITLE = "*Ungrouped Controls*";
+const UNGROUPED_CONTROLS_TITLE = "*Top*";
+
+interface OSCALCatalogControlsListProps {
+  readonly controls: Catalog["controls"];
+  readonly urlFragment?: string;
+}
+
+const OSCALCatalogControlsList: React.FC<OSCALCatalogControlsListProps> = ({
+  controls,
+  urlFragment,
+}) => {
+  const [isControlListItemOpened, setIsControlListItemOpened] = React.useState(false);
+  const [previousHandledFragment, setPreviousHandledFragment] = React.useState(undefined);
+
+  return (
+    <OSCALSection>
+      <Card>
+        <CardContent>
+          <OSCALAnchorLinkHeader>
+            <OSCALSectionHeader>Controls</OSCALSectionHeader>
+          </OSCALAnchorLinkHeader>
+          {controls?.map((control) => (
+            <OSCALCatalogControlListItem
+              key={control.id}
+              control={control}
+              urlFragment={urlFragment}
+              fragmentPrefix=""
+              fragmentSuffix=""
+              isControlListItemOpened={isControlListItemOpened}
+              setIsControlListItemOpened={setIsControlListItemOpened}
+              previousHandledFragment={previousHandledFragment}
+              setPreviousHandledFragment={setPreviousHandledFragment}
+            />
+          ))}
+        </CardContent>
+      </Card>
+    </OSCALSection>
+  );
+};
 
 export interface OSCALCatalogProps extends EditableFieldProps {
   readonly onResolutionComplete: React.EffectCallback;
@@ -15,9 +57,14 @@ export interface OSCALCatalogProps extends EditableFieldProps {
   readonly parentUrl: string;
 }
 
-export const OSCALCatalog: React.FC<OSCALCatalogProps> = (props) => {
-  const { onResolutionComplete, catalog, isEditable, onFieldSave, urlFragment, parentUrl } = props;
-
+export const OSCALCatalog: React.FC<OSCALCatalogProps> = ({
+  onResolutionComplete,
+  catalog,
+  isEditable,
+  onFieldSave,
+  urlFragment,
+  parentUrl,
+}) => {
   useEffect(onResolutionComplete);
 
   const partialRestData = {
@@ -41,10 +88,14 @@ export const OSCALCatalog: React.FC<OSCALCatalogProps> = (props) => {
         urlFragment={urlFragment}
       />
 
-      <OSCALCatalogGroups
-        groups={catalog.controls ? [defaultGroup, ...(catalog.groups ?? [])] : catalog.groups}
-        urlFragment={urlFragment}
-      />
+      {catalog.groups ? (
+        <OSCALCatalogGroups
+          groups={catalog.controls ? [defaultGroup, ...(catalog.groups ?? [])] : catalog.groups}
+          urlFragment={urlFragment}
+        />
+      ) : catalog.controls ? (
+        <OSCALCatalogControlsList controls={catalog.controls} urlFragment={urlFragment} />
+      ) : undefined}
 
       <OSCALBackMatter
         backMatter={catalog["back-matter"]}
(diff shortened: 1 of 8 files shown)


tuckerzp:
Note: I still need to add interface documentation

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -51,39 +58,38 @@ const StyledControlDescriptionWrapper = styled("div")`
...
+): boolean {
   // Ensure fragment exists and split by groupings
   if (!urlFragment || previousHandledFragment === urlFragment) {
-    return;
+    return false;
   }
   // Determine if current control/group list state matches ID
-  const currentList = fragmentSuffix.split("/")[0];
+  const currentList = fragmentSuffix?.split("/")[0];
   return currentList === listId;
```
kylelaker, line 73: Can we just simplify to:

```ts
return (
  urlFragment
  && previousHandledFragment !== urlFragment
  && fragmentSuffix?.split("/")?.[0] === listId
);

File: packages/oscal-react-library/src/components/OSCALCatalogGroups.tsx [hunk 2]
```diff
@@ -71,20 +78,28 @@ TabPanel.propTypes = {
...
  * @param {*} controlLayers An decending list of groups/controls
  * @param {*} rootLayer The top most layer
- * @returns This returns the lowest found control or null if not
+ * @returns This returns the lowest found control or undefined if not
  */
-function determineControlExists(groups, controlLayers, rootLayer) {
+function determineControlExists(
+  groups: ControlGroup[] | undefined,
+  controlLayers: any,
+  rootLayer: any
```
Bronstrom, line 86: `controlLayers` is a list of `strings`, and `rootLayer` is a `string`. Looks like I missed updating the type for the `@param`s in JSDoc for this function. If you can, please update those as well with the types.

Bronstrom (commented):
Looking good so far and thanks for adjusting this to typescript.

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -51,39 +58,38 @@ const StyledControlDescriptionWrapper = styled("div")`
...
-   * {string} fragmentSuffix - The end of a fragment which starts with the current group/control
-   *   being handled
-   * {boolean} listItemOpened - Tells whether or not the current list item is opened
-   * {function} setIsListItemOpened - Callback function to set list item opened
-   * {function} setIsListItemNavigatedTo - Callback function to set whether the list item has been
-   *   navigated to
-   * {string} previousHandledFragment - The fragment that was handled previously
-   * {function} setPreviousHandledFragment - Callback function to set the previously handled fragment
-   */
+const CollapsibleListItem: React.FC<CollapsibleListItemProps> = (props) => {
```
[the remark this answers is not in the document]
tuckerzp: Yeah! As I said above, I still need to update that.

packages/oscal-react-library/src/components/OSCALCatalog.tsx
```diff
@@ -15,6 +24,11 @@ export default function OSCALCatalog(props) {
     },
   };
 
+  const defaultGroup: ControlGroup = {
+    title: "*Ungrouped Controls*",
```
kylelaker, line 28: Can we make "*Ungrouped Controls*" a constant somewhere? I feel like we may want to change this at some point.

packages/oscal-react-library/src/components/OSCALCatalog.tsx
```diff
@@ -25,7 +39,10 @@ export default function OSCALCatalog(props) {
         urlFragment={urlFragment}
       />
 
-      <OSCALCatalogGroups groups={catalog.groups} urlFragment={urlFragment} />
+      <OSCALCatalogGroups
+        groups={catalog.controls ? [defaultGroup, ...(catalog.groups ?? [])] : catalog.groups}
+        urlFragment={urlFragment}
+      />
```
kylelaker, line 45: This will be awkward when there are no groups at all. Realizing that I don't know the names of all the components off the top of my head, what if we did:

```tsx
catalog.groups
  ? <OSCALCatalogGroups groups={catalog.controls ? [defaultGroup, ...(catalog.groups ?? [])] : catalog.groups} urlFragment={urlFragment} />
  : catalog.controls
  ? <OSCALCatalogControlsList controls={catalog.controls} />
  : undefined
```

kylelaker:
> Previously OSCALCatalog ignored the `Controls` prop it was passed. This creates an OSCALControls component to display controls outside a group.
> 
> closes #741 closes #742
> 
> ### Testing
> https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/group-prose/catalogs/basic-test-catalog.json
> 
> `Ungrouped Controls` is now shown as a group
> 
> image
> 
> Top level control groups with `prose` is displayed
> 
> image
> 
> Children groups with `prose` is displayed
> 
> image

Prose should be displayed _above_ the other children (like we do with controls)

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -324,6 +402,11 @@ const OSCALCatalogGroupList: React.FC<OSCALCatalogGroupListProps> = (props) => {
       previousHandledFragment={previousHandledFragment}
       setPreviousHandledFragment={setPreviousHandledFragment}
     >
+      {group.parts
+        ?.map((groupPart: Part) => groupPart.prose)
```
kylelaker, line 406: I am curious why we need to specify the type here -- `tsc` should be able to infer it right?

packages/oscal-react-library/src/components/OSCALCatalog.tsx
tuckerzp: My initial idea was similar to this. The list of controls displayed in the catalog is pretty tightly coupled to groups. Another option would be to display the controls like how we do in profile, however this also would not be super clean and would look a bit weird. 

I think that having a single group, while a bit strange, is the cleanest way to do this at the moment.

packages/oscal-react-library/src/components/OSCALCatalogGroups.tsx
```diff
@@ -68,23 +75,31 @@ TabPanel.propTypes = {
...
   }
+
   // Ensure lowest/deepest control exists within controls/groups
   for (let i = 1; i < controlLayers.length && upperLayer; i += 1) {
-    const subControl = upperLayer?.controls?.find((control) => control.id === controlLayers[i]);
-    const subGroup = upperLayer?.groups?.find(
+    const subControl: Control | undefined = upperLayer?.controls?.find(
+      (control) => control.id === controlLayers[i]
+    );
+    const subGroup: ControlGroup | undefined = upperLayer?.groups?.find(
```
kylelaker, line 102: I am confused about why this needs a specific type -- what part of this causes `tsc` to lose the ability to infer types?

packages/oscal-react-library/src/components/OSCALCatalogGroups.tsx
```diff
@@ -7,32 +7,39 @@ import { styled } from "@mui/material/styles";
...
-function TabPanel(props) {
-  const { children, groupId, value, ...other } = props;
+interface TabPanelProps {
+  children: ReactNode;
+  groupId: string;
+  value: string;
+}
+
+const TabPanel: React.FC<TabPanelProps> = (props) => {
+  const { children, groupId, value } = props;
```
kylelaker, line 31:
```suggestion
const TabPanel: React.FC<TabPanelProps> = ({ children, groupId, value }) => {
```

Just so you're aware, you can destructure `props` in the parameter list if you want. At some point, we should pick one or the other to do.

kylelaker (commented):
If the catalog specifies _only_ `controls` and not `groups`, how is it displayed?

packages/oscal-react-library/src/components/OSCALCatalogGroups.test.tsx
```diff
@@ -30,14 +31,30 @@ const testGroups = [
...
     ],
   },
 ];
 
-function getTextByChildren(text) {
-  function result(_content, node) {
-    const hasText = (checkNode) => checkNode.textContent === text;
+function getTextByChildren(text: string) {
+  function result(_content: any, node: any) {
+    const hasText = (checkNode: any) => checkNode.textContent === text;
```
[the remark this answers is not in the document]
tuckerzp: I *think* node is a bit of a misnomer here. I think this function is honestly a hacky way to test things because of Catalogs being a bit weird. The `node.children` part is really important and `ReactNode` does not have that attribute. 

I don't love it either, but I also think we may want to clean up catalog groups in the future and make this function unneeded.

packages/oscal-react-library/src/components/OSCALCatalogGroups.tsx
tuckerzp:
```
'subControl' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer. [7022]

'subGroup' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer. [7022]
```

kylelaker (commented):
Why does this use a different way to display the list of controls than a Profile does? It seems like when we're listing _just_ controls, those two should use the same code to display the controls?

tuckerzp:
> Why does this use a different way to display the list of controls than a Profile does? 

Profile has always displayed controls differently than the catalog does. That may be something to look into at some point.  I figured we would not want to remove the accordion like feature the controls in the catalog currently have. I can just have them displayed as we do in Profile.

kylelaker:
> Profile has always displayed lists of controls differently than the catalog does. That may be something to look into at some point

I am not sure why we'd go from having three ways of displaying controls (the way CDef/SSP does it, the way profile does it, and the way catalog does it) to four ways (CDef/SSP, Profile, Catalog w/ Groups, Catalog w/out Groups).

If it's something to look into later, we should open the ticket for that fix now.

tuckerzp:
> If it's something to look into later, we should open the ticket for that fix now.

#816 was created to track that. We may want to refactor it a bit / create a few tasks from it.

kylelaker (approved):
Thanks for the work on this @tuckerzp and for opening up #816.

Bronstrom (approved):

PR #790, opened 2023-04-21
chore(revisions): create style for unspecified data fields

tuckerzp:
image

PR #804, opened 2023-04-25
fix: nested `<p>` warnings in test and Viewer

tuckerzp:


packages/oscal-react-library/src/components/OSCALControlProse.js
```diff
@@ -25,9 +25,7 @@ const OSCALStatementEditControlsContainer = styled(Grid)`
   text-align: right;
 `;
 
-const NotImplementedStatement = styled(Typography)(
-  ({ theme }) => `color: ${theme.palette.grey[400]};`
-);
+const NotImplementedStatement = styled(Stack)(({ theme }) => `color: ${theme.palette.grey[400]};`);
```
kylelaker, line 28: This should probably not be a `Stack`... In fact, whatever is using this should probably just use the component you created in #790 right?

Bronstrom (approved):

packages/oscal-react-library/src/components/OSCALControlProse.js
```diff
@@ -309,10 +308,10 @@ export function OSCALReplacedProseWithParameterLabel(props) {
 
   if (!props.isImplemented) {
     return (
-      <NotImplementedStatement>
+      <NotSpecifiedTypography>
```
kylelaker, line 311: How does changing from a `Stack` to a `Typography` preserve the correctness of this? Are we sure this doesn't reintroduce the errors? And can you explain to me how/why?

tuckerzp: You're right. I was worried about that. Ideally I'd set the `Typography` component to something like a `<div>`, but that didn't seem to work.

packages/oscal-react-library/src/components/OSCALControlProse.js
```diff
@@ -315,27 +308,27 @@ export function OSCALReplacedProseWithParameterLabel(props) {
...
```
tuckerzp, line 311: Something like this does not work. 

```suggestion
      <NotSpecifiedTypography component={div}>
```

PR #805, opened 2023-04-25
chore(anchorlinks): Switch AnchorLinkHeader to .tsx

tuckerzp:
This adds an interface with two `any` values. We should eventually be
more explicit about the type, but the helper this function relies on
must be `any`.

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
+++ b/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
@@ -26,11 +26,11 @@ export interface OSCALAnchorLinkHeaderProps {
   /**
    * Value to be used in the id and fragment.
    */
-  value?: any;
+  name?: any;
 }
 
 export const OSCALAnchorLinkHeader: React.FC<OSCALAnchorLinkHeaderProps> = (props) => {
-  const { children, value } = props;
+  const { children, name } = props;
   const [isHover, setIsHover] = useState(false);
   const onEnter = () => {
     setIsHover(true);
@@ -39,7 +39,7 @@ export const OSCALAnchorLinkHeader: React.FC<OSCALAnchorLinkHeaderProps> = (prop
     setIsHover(false);
   };
 
-  const linkId = value || conformLinkIdText(children);
+  const linkId = name || conformLinkIdText(children);
 
   return (
     <Stack
--- a/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
+++ b/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { ReactNode, useState } from "react";
 import Stack from "@mui/material/Stack";
 import LinkIcon from "@mui/icons-material/Link";
 import { Link } from "react-router-dom";
@@ -22,9 +22,10 @@ export interface OSCALAnchorLinkHeaderProps {
    *
    * @example a <Typography />
    */
-  children: any;
+  children: ReactNode;
   /**
-   * Value to be used in the id and fragment.
+   * Value to be used in the id and fragment. If not specified,
+   * children will be used.
    */
   name?: string;
 }
--- a/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
+++ b/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
@@ -54,5 +54,3 @@ export const OSCALAnchorLinkHeader: React.FC<OSCALAnchorLinkHeaderProps> = (prop
     </Stack>
   );
 };
-
-export default OSCALAnchorLinkHeader;
--- a/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
+++ b/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
@@ -26,7 +26,7 @@ export interface OSCALAnchorLinkHeaderProps {
   /**
    * Value to be used in the id and fragment.
    */
-  name?: any;
+  name?: string;
 }
 
 export const OSCALAnchorLinkHeader: React.FC<OSCALAnchorLinkHeaderProps> = (props) => {
--- a/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
+++ b/packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
@@ -12,7 +12,16 @@ const AnchorLinkIcon = styled(LinkIcon)(({ theme }) => ({
   color: theme.palette.grey[500],
 }));
 
-export default function OSCALAnchorLinkHeader(props) {
+export interface AnchorLinkProps {
+  urlFragment?: string;
+}
+
+export interface OSCALAnchorLinkHeaderProps {
+  children: any;
+  value?: any;
+}
+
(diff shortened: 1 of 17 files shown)


packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
```diff
@@ -12,7 +12,16 @@ const AnchorLinkIcon = styled(LinkIcon)(({ theme }) => ({
   color: theme.palette.grey[500],
 }));
 
-export default function OSCALAnchorLinkHeader(props) {
+export interface AnchorLinkProps {
+  urlFragment?: string;
+}
+
+export interface OSCALAnchorLinkHeaderProps {
```
Bronstrom, line 19: Is there any reason for this to be exported? I think this interface is just used in this file?

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
```diff
@@ -44,4 +53,6 @@ export default function OSCALAnchorLinkHeader(props) {
       )}
     </Stack>
   );
-}
+};
+
+export default OSCALAnchorLinkHeader;
```
Bronstrom, line 58: I'm wondering if this is the best approach. We're taking a named export (`export const`) and then exporting it as the default export (which is typically used for functions). I would say at the very least remove the `export` in front of `const OSCALAnchorLinkHeader: . . .` above

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
```diff
@@ -12,7 +12,16 @@ const AnchorLinkIcon = styled(LinkIcon)(({ theme }) => ({
...
 }));
 
-export default function OSCALAnchorLinkHeader(props) {
+export interface AnchorLinkProps {
+  urlFragment?: string;
+}
+
+export interface OSCALAnchorLinkHeaderProps {
+  children: any;
+  value?: any;
```
Bronstrom, line 21: I would suggest adding some documentation to these variables. I can help you out if you need any.

Bronstrom (commented):
Good idea to turn this into TypeScript!

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
tuckerzp: @Bronstrom I would appreciate that! They are `any` and don't have documentation since it is not super clear what they should be :laughing:

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
tuckerzp: If a component is exported, its interface should also be exported.

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
tuckerzp: https://github.com/EasyDynamics/oscal-react-library/blob/6c1e060dae048064cb90cea7f587c79f14c2cdca/packages/oscal-react-library/src/components/OSCALMetadata.tsx#L797

This has already been set as a precedence

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
```diff
@@ -12,7 +12,16 @@ const AnchorLinkIcon = styled(LinkIcon)(({ theme }) => ({
   color: theme.palette.grey[500],
 }));
 
-export default function OSCALAnchorLinkHeader(props) {
+export interface AnchorLinkProps {
+  urlFragment?: string;
+}
+
+export interface OSCALAnchorLinkHeaderProps {
+  children: any;
```
kylelaker, line 20: Do we just take in `children` because that's what React elements do? I'll wait for the docs but it feels like we can do something smarter here.

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
kylelaker, line 21: Would love to see someone suffer the pain of renaming this to a _way_ more meaningful name once we know what it actually does lol

kylelaker (changes requested):
Is Metadata really the only place using anchor link header stuff that's been migrated to TS?

tuckerzp:
> Is Metadata really the only place using anchor link header stuff that's been migrated to TS?

Metadata is the only `.tsx` place using anchor links that I know of.

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
tuckerzp: `conformLinkIdText` is the reason that this function is hard to assign types to (I think). We are allowed to put pretty much anything as an anchor link, which I feel like is a bit weird. However I am unsure how to fix that helper function without rewriting a lot of how we do anchor links.

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
tuckerzp: It was removed

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
Bronstrom: Here is the general idea of the two:

`children` - (`any`, but would suggest trying `ReactNode`) This is the text or element, which is most commonly a styled `Typography`, surrounded by an `OSCALAnchorLinkHeader`.
`value` - (`string` or `null`) This is a given value/name to be used in the id and fragment. When one of these isn't provided the text within `children` is used for the id and fragment.

tuckerzp: `conformLinkIdText` makes it so that `children` can not be a `ReactNode` I thought, right?

Bronstrom: `conformLinkIdText` returns a `string`, but it's one parameter is `any`. I'm pretty sure `children` here is `any`, and its being passed to `conformLinkIdText`.

Bronstrom:
```suggestion
  name?: string;
```

`name` is probably more fitting. This is just an optional `string` that is provided for the id of the anchor link and fragment.

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
kylelaker: I am still confused why this can't be `ReactNode`

packages/oscal-react-library/src/components/OSCALAnchorLinkHeader.tsx
```diff
@@ -12,8 +12,25 @@ const AnchorLinkIcon = styled(LinkIcon)(({ theme }) => ({
...
+  /**
+   * Text or element to link to.
+   *
+   * @example a <Typography />
+   */
+  children: any;
+  /**
+   * Value to be used in the id and fragment.
+   */
+  name?: string;
```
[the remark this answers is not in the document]
Bronstrom: The text within `children` will be used within the id and fragment

Bronstrom (approved):

kylelaker (approved):

PR #826, opened 2023-04-28
feat(party): display metadata party logos

### LINKED ISSUE #411: Display Metadata Party Logos
### Description 
Certain parties represent organizations or companies. In some OSCAL examples, namely the [FedRAMP SSP Template](https://github.com/GSA/fedramp-automation/blob/master/dist/content/templates/ssp/json/FedRAMP-SSP-OSCAL-Template.json), a link to the logo is provided. As a user of such an OSCAL object, I want the logo to be displayed so that I can verify that it is included in the OSCAL object and that it is properly functioning.

### Acceptance Criteria
- The metadata React component displays the logo of any parties that have linked a logo
- The tests are updated to demonstrate the new functionality

### Implementation Notes
Party logos live in OSCAL Components, Catalogs, Profiles, and SSPs as links in `back-matter`, with a direct link to the logo image in its `rlinks`.
An object in `metadata/parties` can link this logo with an element in its `links` attribute, which is an object with:
- `href`, pointing to the `uuid` of the above object in `back-matter`
- and `rel`, set to `logo`

tuckerzp:
This adds the logo of a metadata party to the Details modal.

closes #411 

### Testing

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/display-logo/catalogs/basic-test-catalog.json

---

image

---

image

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALMetadata.tsx
+++ b/packages/oscal-react-library/src/components/OSCALMetadata.tsx
@@ -47,10 +47,12 @@ import {
   Role,
   PartyType,
   TelephoneNumber,
+  Link as OSCALLink,
 } from "@easydynamics/oscal-types";
 import { OSCALRevisionsButton } from "./OSCALRevision";
 import { OSCALMetadataLabel } from "./OSCALMetadataCommon";
 import { NotSpecifiedTypography } from "./StyledTypography";
+import resolveLinkHref, { ResolvableLinkHref } from "./oscal-utils/OSCALLinkUtils";
 
 const OSCALMetadataSectionInfoHeader = styled(Typography)`
   display: flex;
@@ -306,34 +308,72 @@ const MetadataInfoList: React.FC<MetadataInfoListProps> = (props) => {
   );
 };
 
-export interface OSCALMetadataPartyProps {
+interface OSCALMetadataPartyLogoProps extends ResolvableLinkHref {
+  link: OSCALLink;
+  partyName?: string;
+}
+
+const OSCALMetadataPartyLogo: React.FC<OSCALMetadataPartyLogoProps> = ({
+  link,
+  backMatter,
+  parentUrl,
+  partyName,
+}) => {
+  let logoUri = undefined;
+
+  if (backMatter) {
+    try {
+      logoUri = resolveLinkHref(backMatter, link.href, parentUrl, /^image\//, false);
+    } catch (err) {
+      // Silently fail on unresolved diagram resources
+    }
+  }
+
+  const imageSrc = logoUri ?? link.href;
+
+  return <img src={imageSrc} alt={`${partyName} logo`} style={{ maxWidth: "10em" }} />;
+};
+
+export interface OSCALMetadataPartyProps extends ResolvableLinkHref {
   party: PartyOrganizationOrPerson;
   partyRolesText: Role[] | undefined;
 }
 
-export const OSCALMetadataParty: React.FC<OSCALMetadataPartyProps> = (props) => {
-  const fallbackIcon =
-    props.party?.type === PartyType.ORGANIZATION ? <GroupIcon /> : <PersonIcon />;
+export const OSCALMetadataParty: React.FC<OSCALMetadataPartyProps> = ({
+  party,
+  partyRolesText,
+  backMatter,
+  parentUrl,
+}) => {
+  const fallbackIcon = party?.type === PartyType.ORGANIZATION ? <GroupIcon /> : <PersonIcon />;
 
-  const avatar = (
-    <MetadataAvatar id={props.party.uuid} text={props.party.name} fallbackIcon={fallbackIcon} />
-  );
+  const avatar = <MetadataAvatar id={party.uuid} text={party.name} fallbackIcon={fallbackIcon} />;
+
+  const logoLink = party.links?.find((item) => item.rel === "logo");
+  const logo = logoLink ? (
+    <OSCALMetadataPartyLogo
+      link={logoLink}
+      backMatter={backMatter}
+      parentUrl={parentUrl}
+      partyName={party.name}
+    />
+  ) : undefined;
 
   return (
     <OSCALMetadataCard
-      title={props.party.name}
-      subheader={props.partyRolesText?.map((role: Role) => role.title).join(", ")}
+      title={party.name}
+      subheader={partyRolesText?.map((role: Role) => role.title).join(", ")}
       avatar={avatar}
     >
       <DialogTitle id="scroll-dialog-title">
-        <Stack direction="row" alignItems="center" gap={1}>
-          {avatar}
+        <Stack direction="row" alignItems="center" justifyContent="space-between">
           <Stack direction="column">
-            {props.party.name}
-            {props.partyRolesText?.map((role) => (
+            {party.name}
+            {partyRolesText?.map((role) => (
               <Typography key={role.title}> {role.title} </Typography>
             ))}
           </Stack>
+          {logo}
         </Stack>
       </DialogTitle>
       <DialogContent dividers>
@@ -342,7 +382,7 @@ export const OSCALMetadataParty: React.FC<OSCALMetadataPartyProps> = (props) => 
             <OSCALMetadataContactTypeHeader icon={<MapIcon fontSize="small" />} title="Address" />
             <List>
               <MetadataInfoList
-                list={props.party.addresses}
+                list={party.addresses}
                 infoType={MetadataInfoType.address}
                 emptyMessage="No address information provided"
               />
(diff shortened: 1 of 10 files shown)


packages/oscal-react-library/src/components/OSCALMetadata.tsx
```diff
@@ -306,34 +308,72 @@ const MetadataInfoList: React.FC<MetadataInfoListProps> = (props) => {
...
+
+  if (backMatter) {
+    try {
+      logoUri = resolveLinkHref(backMatter, link.href, parentUrl, /^image\//, false);
+    } catch (err) {
+      // Silently fail on unresolved diagram resources
+    }
+  }
+
+  const imageSrc = logoUri ?? link.href;
```
tuckerzp, line 332: Im not sure if this is the cleanest way to deal with `resolveLinkHref` failing and the fact that `backMatter` doesn't need to exisit.

I was thinking that it could be `resolveLinkHref` that deals with an undefined `backMatter` but just did this for now.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
```diff
@@ -13,10 +16,21 @@ export function getAbsoluteUrl(href: string | undefined, parentUrl: string): str
   return new URL(href ?? "", parentUrl).toString();
 }
 
+export interface ResolvableLinkHref {
```
tuckerzp, line 19: Not sure if there is a better name for this interface....

kylelaker, line 19:
```suggestion
export interface UriReferenceLookup {
```

Is it better for the various components to _extend_ this interface? Or to have it as a field?

```ts
export interface MyComponentProps extends UriReferenceLookup {
}
```

or

```ts
export interface MyComponentProps {
  uriReferenceLookup: UriReferenceLookup;
}
```

packages/oscal-react-library/src/components/OSCALMetadata.tsx
Bronstrom, line 332:
```suggestion
  let imageSrc;
  if (backMatter) {
    try {
      imageSrc = resolveLinkHref(backMatter, link.href, parentUrl, /^image\//, false);
    } catch (err) {
      imageSrc = link.href;
    }
  }
```

An idea that may not be entirely correct.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
```diff
@@ -1,7 +1,10 @@
 import type { BackMatter } from "@easydynamics/oscal-types";
 import { BackMatterLookup } from "./OSCALBackMatterUtils";
 
-export function getAbsoluteUrl(href: string | undefined, parentUrl: string): string | undefined {
+export function getAbsoluteUrl(
+  href: string | undefined,
+  parentUrl: string | undefined
```
Bronstrom, line 6: This could be an interface as well.

packages/oscal-react-library/src/components/OSCALMetadata.tsx
Bronstrom: You can write the catch as: `catch(_)` if unused param gives trouble or maybe even use `finally` instead.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
Bronstrom: I'm also wondering do we favor using `parentUrl?: string` over `parentUrl: string | undefined`? It may make sense to write it as the former in the case that we receive an empty string.

packages/oscal-react-library/src/components/OSCALMetadata.tsx
tuckerzp: An issue with that is if `backMatter` is falsy, `imageSrc` is never initialized. I had considered doing what you did and then also initializing it on line 323, but that seemed uncessary. I think its kinda a gross chain of logic no matter what, so I chose what I felt was the best. 

I do however need to update that comment...

Bronstrom:
```suggestion
  let imageSrc;
  let logoUri;
  
  if (backMatter) {
    try {
      logoUri = resolveLinkHref(backMatter, link.href, parentUrl, /^image\//, false);
    } finally {
      imageSrc = logoUri ?? link.href;
    }
  }
```
Another idea . . .

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
tuckerzp: > This could be an interface as well.

Do we gain much by doing so? The one below was created as components that call this function are passed props that match its arguments. This is not the case where we call `getAbsoluteUrl`. Not saying that I won't do it, I just am wondering why we would. 

> I'm also wondering do we favor using parentUrl?: string over parentUrl: string | undefined? It may make sense to write it as the former in the case that we receive an empty string.

I personally think that *something* should always be passed to the function. Even if it is `""`. All places we call this currently we put something there. 

I think calling a function called `getAbsoluteUrl` somewhat implies there is likely a `parentUrl` to care about. That is different from `resolveLinkHref` where you don't need a parent url to resolve a link.

packages/oscal-react-library/src/components/OSCALMetadata.tsx
Bronstrom:
```suggestion
  let imageSrc;
  let logoUri;
  
  try {
    if (backMatter) {
      logoUri = resolveLinkHref(backMatter, link.href, parentUrl, /^image\//, false);
    }
  } finally {
    imageSrc = logoUri ?? link.href;
  }
```
I'm trying to avoid us having to use half the functionality of a try-catch block and I think we can do that here. This would work in the case of `backMatter` being false.

tuckerzp: I am not convinced removing the catch is inherently better. I think the fact that we are actively choosing to ignore an error is not something to hide with more complexity. Especially if that means we have to add another `let`, which is also not ideal.

I realize that basing decisions on old code isn't great, but we do this in `OSCALDiagram`. If we decide to do something when `resolveLinkHref` fails, we would go back to the catch in both places.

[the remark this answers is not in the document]
tuckerzp: This is a lot cleaner but also has the same issue that `resolveLinkHref` requires `backMatter` to be defined. I  am unsure where to put the check that makes the most sense.

Since `resolveLinkHref` more or less provides a nice wrapper around using the `BackMatterLookup` class, I thought that it doesnt make sense for you to not have `backmatter` when calling `resolveLinkHref`.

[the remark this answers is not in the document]
kylelaker: Just add `if (!backMatter) return undefined` to the top of `maybeResolveUri`?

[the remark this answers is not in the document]
tuckerzp: I did that, you just cant do `const maybeResolveUri: typeof resolveLinkHref`

kylelaker (commented):
Good stuff so far! Two small questions before I approve.

packages/oscal-react-library/src/components/OSCALMetadata.tsx
```diff
@@ -842,7 +906,7 @@ export const OSCALMetadata: React.FC<OSCALMetadataProps> = (props) => {
             <OSCALMetadataParties
               parties={props.metadata?.parties}
               metadata={props.metadata}
-              urlFragment={props.urlFragment}
+              uriReferenceLookup={{ backMatter: props.backMatter, parentUrl: props.parentUrl }}
```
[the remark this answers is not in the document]
tuckerzp: Looking at the live Viewer, we can't navigate to an individual party. Linking to the #parties section also still works. With that said, that was an accident deleting that.

packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
```diff
@@ -13,12 +16,23 @@ export function getAbsoluteUrl(href: string | undefined, parentUrl: string): str
...
+}
+
 export default function resolveLinkHref(
   backMatter: BackMatter,
   href: string,
-  parentUrl: string,
   mediaType: RegExp,
-  preferBase64 = false
+  preferBase64 = false,
+  parentUrl?: string
```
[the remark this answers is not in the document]
kylelaker: Yep! Exactly! And then if the variable names are the same you can use the shorthand of

```ts
resolveLinkHref({ backMatter, href, ... });
```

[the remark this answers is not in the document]
tuckerzp: I will update this then

Bronstrom (approved):
Looks good @tuckerzp!

PR #827, opened 2023-04-28
fix: withdrawn controls are not displayed correctly

kylelaker:
This makes it so that withdrawn controls can be expanded. It also makes it so that nested controls/groups display as withdrawn if a parent is withdrawn.

- Withdrawn Control 
image

- Withdrawn Control with children 
image


packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -159,96 +145,47 @@ const CollapsibleListItem: React.FC<CollapsibleListItemProps> = (props) => {
...
-    </StyledListItemPaper>
+    <Accordion
+      expanded={isOpen}
+      onChange={handleClick}
+      TransitionProps={{
+        onEntered: () => !group && setListItemOpened(true),
+        unmountOnExit: true,
+        timeout: 0,
+      }}
+      disabled={false && props.isEmpty}
```
kylelaker, line 157: Thinking of just never disabling... the styling is weird (see https://mui.com/material-ui/react-accordion/#basic-accordion). If you want this, please provide a good style for disabled accordions. Otherwise, I think having a small expand isn't that bad.

I do like that this makes `disabled` totally distinct from `withdrawn`

packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -1,32 +1,58 @@
 import React, { useEffect } from "react";
```
tuckerzp, line 1: Thank you for switching this to `.tsx`!

packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -1,32 +1,58 @@
...
-    !["childLevel", "withdrawn", "ownerState", "theme", "sx", "as"].includes(prop),
-})`
-  margin-top: 1em;
-  margin-bottom: 1em;
-  margin-left: ${(props) => (props.childLevel > 0 ? "1.5em" : "0")};
-  margin-right: ${(props) => (props.childLevel > 0 ? "1.5em" : "0")};
-  ${(props) => props.childLevel > 0 && "background-color: #fffcf0;"}
-  ${(props) =>
-    props.withdrawn && `text-decoration: line-through; color: ${props.theme.palette.grey[400]};`}
-`;
```
kylelaker, line 27: Removing this has caused _more_ complaints about `withdrawn` and `childLevel` and the DOM and all that in the browser console. `shouldForwardProp` probably needs to be added back?

packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -1,32 +1,58 @@
...
+  control: Control;
+  componentId?: string;
+  implementedRequirement?: ImplementedRequirementElement;
+  includeControlIds?: string[];
+  excludeControlIds?: string[];
+  onRestError?: (...args: any[]) => void;
+  onRestSuccess?: (...args: any[]) => void;
+  modificationAlters?: Alteration[];
+  modificationSetParameters?: SetParameterValue[];
+  withdrawn: boolean;
```
kylelaker, line 52: `onRestSuccess` and `onRestError` probably need to be in another interface that this one extends.

packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -43,6 +69,8 @@ function ControlsList(props) {
     fragmentPrefix,
   } = props;
   return (
+    // TODO: I think that in this `control.parts` map is where we need to stuff to make the
+    // last bit of text use the `WithdrawnControlText` component? But I am not 100% sure.
```
kylelaker, line 73: This also might be wrong? Not entirely sure. `OSCALControlCard` just has the text decoration applied to the whole card which helps (and actually may have implicitly marked children as withdrawn). But in the list view, that's not there to help us (no card) so I *think* this is correct.

I _really_ don't want to just apply this styling to the `Accordion` because that's too focused on the implementation and means we wouldn't be able to display withdrawn controls in other contexts outside of the accordion. So I think the right fix is to make control parts wiser about withdrawn stuff?

packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -74,17 +102,25 @@ function ControlsList(props) {
...
     </div>
   );
+};
+
+export interface OSCALControlProps extends ControlsListProps, AnchorLinkProps {
+  listItemOpened?: boolean;
+  isItemNavigatedTo?: boolean;
+  showInList?: boolean;
+  previousHandledFragment?: string;
+  setPreviousHandledFragment?: (fragment: string) => void;
```
kylelaker, line 120: I think this stuff about being open and about fragments probably should be an interface if it gets used in any other components ever...

packages/oscal-react-library/src/components/oscal-utils/OSCALCatalogUtils.ts
```diff
@@ -7,10 +7,10 @@ import matchingProp from "./OSCALPropUtils";
  * @param control A catalog control
  * @returns true if control is withdrawn
  */
-export default function isWithdrawn(control: Control): boolean {
+export default function isWithdrawn(item: Control | ControlGroup): boolean {
```
kylelaker, line 10:
```suggestion
interface Propable {
  props?: Property[];
}
export default function isWithdrawn(item: Propable): boolean {
```

Not sure if this is good or dumb but it's an idea I had while implementing this. I'd have to read and see what besides these two could be withdrawn.

packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -135,9 +172,9 @@ export default function OSCALControl(props) {
   const label = propWithName(control.props, "label")?.value;
 
   return showInList ? (
-    <ControlsList {...props} />
+    <ControlsList {...props} withdrawn={withdrawn || isWithdrawn(control)} />
   ) : (
-    <OSCALControlCard childLevel={childLevel ?? 0} withdrawn={isWithdrawn(control)}>
+    <OSCALControlCard childLevel={childLevel ?? 0} withdrawn={withdrawn || isWithdrawn(control)}>
```
kylelaker, line 177: This recurring `withdrawn || isWithdrawn(control)` needs to become a variable.

packages/oscal-react-library/src/components/OSCALControl.tsx
kylelaker: Going to do this in #830 instead...

packages/oscal-react-library/src/components/OSCALControl.tsx
kylelaker: This should probably be done in a separate PR.

kylelaker:
> First of all, I notice a few styling differences. There is no gap between out top-level controls, the control's are white (was a very light gray), and the corners are no longer rounded.

This is intentional as part of the switch from a `Collapse` to an `Accordion` to match the other collapsable fields in the application. Is this an observation or a requested change?

kylelaker:
> Also, I see no click animation occurring on the top-level controls, and this may just be an issue on my end but when I open a control, the opening animation seems more sluggish.

So are you seeing an animation or not? I am confused about what exactly you're reporting here.

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -159,96 +143,29 @@ const CollapsibleListItem: React.FC<CollapsibleListItemProps> = (props) => {
...
-    <StyledListItemPaper>
-      <StyledListItem>{children}</StyledListItem>
-    </StyledListItemPaper>
+    <Accordion
+      expanded={isOpen}
+      onChange={handleClick}
+      TransitionProps={{
+        onEntered: () => !group && setListItemOpened(true),
+        unmountOnExit: true,
+        timeout: 0,
```
kylelaker, line 153: This is what we're using to prevent the long transitions. If you want to have animations but faster, then increase `timeout`. It is `0` because as-is on `develop` today, if you open AC-3 and then close it, you have to wait for the animation to close things _below your viewport_ before you see any feedback that you clicked the button.

Maybe 0 is too jarring but suggest a value.

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -112,6 +89,13 @@ interface CollapsibleListItemProps extends CatalogGroupFragmentProps {
    * Callback function to set list item opened
    */
   readonly setListItemOpened: (...args: any[]) => void;
+
+  /**
```
tuckerzp, line 93: Not sure if this space was intentional or not
```suggestion
  readonly setListItemOpened: (...args: any[]) => void;
  /**
```

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
tuckerzp: I think 0 is fine for now. It is a pretty simple thing to change later if we want to.

tuckerzp (approved):
After playing with the accordion more, I do think this looks and works better than what we previously had.

Thanks for your work on this @kylelaker

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
kylelaker: Absolutely intention.

Bronstrom (approved):
Looks good @kylelaker!

kylelaker:
@tuckerzp and @Bronstrom I have updated the styling to be closer to the old styling but also slightly closer to the Accordion examples, which should make this easier to _later_ reuse in Metadata (I will open a PR for that after this one has been merged)

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
kylelaker: I changed this to `500`

tuckerzp (approved):

Bronstrom (approved):

PR #829, opened 2023-05-01
chore: bump timeout on profile tests

kylelaker:
These tests perform profile resolution and reach out to do HTTP requests
and all sorts of things. We have not set specific performance
thresholds; the goal of this test is to ensure the underlying code
works, not necessarily that it works *quickly*.

This increases the timeout to 30s because we've been bumping it by a
second or two and I'm a bit tired of that. This should give us enough
padding to know whether we're slow or broken.


tuckerzp (approved):

Bronstrom (approved):
Definitely a needed change, at least for now.

PR #830, opened 2023-05-01
chore: add types to the loader

kylelaker:
This takes a first swing at adding type information *making as few
changes to the code as possible* barring very explicit bugs or things
required to get types to work.

There is a non-exported and deprecated `Mutable` type added in the REST
Utils. It should not find any additional use and all existing uses
should eventually be replaced. This is an example of a bad thing done to
avoid breaking code.

Additionally, I don't understand how `populatePartialRestData` ever
worked. There is all kinds of stuff with checking `.length` and then
trying to use that list (or single-character string?) to index a list.
That is the biggest functional change to the code here.

I would appreciate knowing this has been tested against various editing
paths in the editor. If you reproduce a bug, be explicit about what you
edited and include the built path, request, and any response data (and
logs from the server) and post it as a comment on the PR.


packages/oscal-react-library/src/components/OSCALProfile.js
```diff
@@ -17,7 +17,7 @@ import { OSCALAnchorLinkHeader } from "./OSCALAnchorLinkHeader";
  * represents the baseline of selected controls from one or more control catalogs.
  * For more information see: https://pages.nist.gov/OSCAL/concepts/layer/control/profile/
  *
- * @param {Object} props
+ * @param {*} props
```
kylelaker, line 20: `profile` is not an attribute of `Object` and I am _not_ diving into typing `OSCALProfile` right now lol

packages/oscal-react-library/src/components/oscal-utils/OSCALRestUtils.ts
```diff
@@ -1,50 +1,34 @@
...
-  POST: "POST",
-  PUT: "PUT",
-};
+export enum RestMethod {
+  DELETE = "DELETE",
+  GET = "GET",
+  PATCH = "PATCH",
+  POST = "POST",
+  PUT = "PUT",
+}
```
Bronstrom, line 20: Great use of `enum` here!

Bronstrom (commented):
This is looking great @kylelaker! This is definitely a step in the right direction to improve our higher-level components. It would be nice to have some of these types nailed down a little more, but it's definitely difficult to determine some of these. At the very least it would be nice to have some documentation behind more of what these attributes are and have interfaces for them.

packages/oscal-react-library/src/components/OSCALLoader.tsx
```diff
@@ -54,37 +64,68 @@ export function getRequestedUrl() {
...
+  handleRestError: (e: Error) => void;
+  handleFieldSave: (...args: any[]) => any;
+}
+
+export default function OSCALLoader(props: OSCALLoaderProps): ReactElement {
   const [isLoaded, setIsLoaded] = useState(false);
   const [isResolutionComplete, setIsResolutionComplete] = useState(false);
   const [hasDefaultUrl, setHasDefaultUrl] = useState(props.hasDefaultUrl);
-  const [oscalData, setOscalData] = useState([]);
+  const [oscalData, setOscalData] = useState(null as unknown as OscalWithSource);
```
tuckerzp, line 101: Just for my understanding, this is unideal right?

tuckerzp (approved):
This is an awesome first step to refactoring how we do loading / editing!

I tried to beat the loader up quite a bit, and things seem to be working correctly! end-to-end tests pass locally, our unit tests pass, editing functionality works, and I haven't had issues loading any documents!

Something I did notice (which we don't need to fix here) is that the description field when we edit control implementations on an SSP doesn't seem to do anything anymore. I thought we could hover and view it before? This may be related to #592, but I thought it was worth mentioning.

packages/oscal-react-library/src/components/OSCALLoader.tsx
kylelaker: Definitely -- and a holdover from before I defined `OscalWithSource`. `{} as OscalWithSource` is much better and what it is defined as now.

Bronstrom (approved):

Issue #364, opened 2022-04-19
Update to React 18


kylelaker:
[React 18 was released last month](https://reactjs.org/blog/2022/03/29/react-v18.html) and it brings some features for concurrent rendering that might be cool. It's also generally good to keep our dependencies up-to-date. React 18 brings some new concurrent rendering capabilities.

### Implementation Notes

 - We have a lot of dependencies tightly coupled to React, we need to make sure they all work together (including `@testing-library/react`, `react-scripts`, and `react-dom-router`)
 - There is a blog post on [How to Upgrade to React 18](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html)

### Acceptance Criteria

 - React 18 is used without any known regressions
 - The [new Root API](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis) has been adopted
 - Ensure current testing is sufficient to validate successful upgrade. Adjust testing if necessary.
 - Ensure no tests fail related to components use of new version.

kylelaker:
`react-ga` was another one that caused dependency conflicts but that's resolved in #449

kylelaker:
We have started to see test failures on the `oscal-editor-deployment` repository that coincide with the merging of the React 18 upgrade. I don't want to move forward with closing this task until we're sure those tests are passing correctly.

Bronstrom:
I haven't been watching the test failures closely on `oscal-editor-deployment` lately. Are we still thinking they could be failing as a result of us updating to React 18 @kylelaker?

PR #832, opened 2023-05-02
chore: update to react 18

kylelaker:
This is a first attempt at updating to React 18. Notably, the page
content never displays. Is this possibly due to the new batching of
requests? Maybe.


kylelaker:
This should be used as the starting point for #364.

Important to note that when resolving conflicts in `package-lock.json` and other files, you may end up with multiple versions of React installed which will cause some _crazy_ errors. Run `npm ls react` in the root of the repo to see what versions are being pulled in. If it's inconsistent, then `npm update` may help.

tuckerzp:
Taking this out of draft for now. I think that things are working. Editing & Viewing seem to work the same and the tests are passing. 

I would love a fairly thorough review to make sure I haven't missed anything.  

I also am unsure why tests are not running...

packages/oscal-react-library/package.json
kylelaker, line 1: I think we need to bump the peer dependencies to `^18.2.0` for React/react-dom too. Pretty sure we no longer support <18.0.

kylelaker:
@tuckerzp reopen so I can review

PR #843, opened 2023-05-05
feat: Add properties story

### LINKED ISSUE #718: Display properties for metadata, back matter, and controls

Bronstrom:
This adds a properties story in storybook.

Closes #718

### DIFF
--- a/packages/oscal-react-library/src/stories/OSCALProperties.stories.tsx
+++ b/packages/oscal-react-library/src/stories/OSCALProperties.stories.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+import { StoryFn } from "@storybook/react";
+import { Property } from "@easydynamics/oscal-types";
+import { OSCALPropertiesDialog, OSCALPropertiesDialogProps } from "../components/OSCALProperties";
+
+const exampleProperties: Property[] = [
+  {
+    name: "Example Property 1",
+    value: "Example Value",
+    ns: "http://easydynamics.com/ns/oscal/example",
+  },
+  {
+    name: "Example Property 2",
+    value: "Example Value",
+  },
+  {
+    name: "Example Property 3",
+    value: "Example Value",
+  },
+];
+
+export default {
+  title: "Components/Properties Dialog",
+  component: OSCALPropertiesDialog,
+};
+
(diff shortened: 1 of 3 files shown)


packages/oscal-react-library/src/stories/OSCALProperties.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+  },
+  {
+    name: "Example Property 2",
+    value: "Example Value",
+  },
+  {
+    name: "Example Property 3",
+    value: "Example Value",
+  },
+];
```
Bronstrom, line 20: For our storybook content, would we rather include our data within where we're defining the story or keep all data to a seperate file?

kylelaker: i like this better

tuckerzp:
> However, looks like this poses an issue to the linter:

Not sure if this is correct, but what happens if you give the linter what it wants?

packages/oscal-react-library/src/stories/OSCALProperties.stories.tsx
tuckerzp: I think for something so small and specific to the `Properties` storybook page, it is ok to put it in the file itself.

kylelaker:
> Not sure if this is correct, but what happens if you give the linter what it wants?

To explain why this is bad idea, Storybook is not a run time dependency so should be listed in `devDependencies`. The linter doesn't realize Stories are special files that aren't in the production build.

kylelaker (approved):

kylelaker (approved):

tuckerzp (approved):

PR #848, opened 2023-05-08
chore: implement global styling components

### LINKED ISSUE #775: Implement Styles at a Global Scale
### Description

As a developer of the OSCAL Viewer, I would like styles to be implemented at a global level for easy reusability among components.

### Acceptance Criteria

* Following a propsed strategy, an implementation of styles at a global scale is made
* A directory structure is implemented
* Documentation should be provided for other developers to understand how it relates to our style guide
  * Include details in README
  * Include comments with styling 

### Proposed Solution

_No response_

### Additional Notes

* No refactoring of code should be completed in this issue
* This issue depends on https://github.com/EasyDynamics/oscal-react-library/issues/774

### Implementation

Bronstrom:
For better reusability of styles, common styling of our application is moved to the top level of our application and passed to lower level components with the use of `ThemeProvider` and `GlobalStyles`. This issue also works to add the initial implementation of @emcathow's style guide.

Resolves #775


packages/oscal-viewer/src/themes/AppTheme.js
```diff
@@ -0,0 +1,37 @@
...
+    secondary: {
+      main: "#00BDE3",
+    },
+  },
+  typography: {
+    h1: {
+      fontWeight: "700",
+      fontSize: "50px",
+    },
+    fontFamily: ["Source Sans Pro", "sans-serif"].join(","),
```
kylelaker, line 17: Can we do the same thing we do for `oscal.io` and use the Roboto font? I started to do that in #809. I am fine with using that same import from Google Fonts.

Generally though, please try to align things with
 - https://github.com/EasyDynamics/oscal.io/blob/main/assets/css/main.scss
 - https://github.com/EasyDynamics/oscal.io/blob/main/_sass/customize.scss
 - https://github.com/EasyDynamics/oscal.io/blob/main/_sass/colors.scss
 - https://github.com/EasyDynamics/oscal.io/blob/main/_sass/base.scss

packages/oscal-viewer/src/GlobalStyles.jsx
```diff
@@ -0,0 +1,94 @@
...
+  },
+  ".App-logo": {
+    height: "40vmin",
+    pointerEvents: "none",
+  },
+  "@media (prefers-reduced-motion: no-preference)": {
+    ".App-logo": {
+      animation: "App-logo-spin infinite 20s linear",
+    },
+  },
```
kylelaker, line 28: Delete this too

packages/oscal-viewer/src/GlobalStyles.jsx
```diff
@@ -0,0 +1,94 @@
...
+});
+
+export function OSCALGlobalStyles(props) {
+  const { theme } = props;
+  console.log(theme.palette.primary.main);
+  return (
+    <>
+      <GlobalStyles styles={globalStyles(theme)} />
+    </>
+  );
```
kylelaker, line 93:
```suggestion
  return <GlobalStyles styles={globalStyles(theme)} />;
```

When the time comes to remove the log, ditch the Fragment and the destructuring to simplify this.

packages/oscal-viewer/src/App.js
```diff
@@ -231,8 +220,14 @@ function App() {
 
   return (
     <StyledEngineProvider injectFirst>
+      <link rel="preconnect" href="https://fonts.googleapis.com" />
+      <link rel="preconnect" href="https://fonts.gstatic.com" />
+      <link
+        href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,200;0,300;0,400;0,600;0,700;0,900;1,200;1,300;1,400;1,600;1,700;1,900&display=swap"
+        rel="stylesheet"
+      />
```
kylelaker, line 228: Is this really how we should do this? I don't love `<link>` tags in the body like this; in fact, I don't think the `preconnect` will even be honored. Perhaps it needs to go in `public/index.html`? It's static HTML that doesn't need React (and that React doesn't need to care about).

Bronstrom: My idea here was to keep the top-level styling together (the font, theme, and global styles all being declared right here), but I think your right on it making more sense to have static HTML `link` be in an HTML file. Will move it to the place you suggested.

packages/oscal-viewer/src/GlobalStyles.jsx
```diff
@@ -0,0 +1,94 @@
...
+    color: "#00BDE3",
+  },
+  "@keyframes App-logo-spin": {
+    from: {
+      transform: "rotate(0deg)",
+    },
+    to: {
+      transform: "rotate(360deg)",
+    },
+  },
```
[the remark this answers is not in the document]
Bronstrom: I noticed a couple of the ".App" styles looked like they weren't being used. Will remove.

packages/oscal-viewer/src/themes/AppTheme.js
Bronstrom: The choice to use Source Sans Pro here was to match Emily's style guide. We may want to have a larger discussion on this as a team before we decide on making this change. I've been hearing some opinions on keeping the tools completely separate including styling - as `oscal.io` is made for the community and will be separate from our own branding, while this tool will fit our own branding.

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -16,7 +17,7 @@ import { AnchorLinkProps, OSCALAnchorLinkHeader } from "./OSCALAnchorLinkHeader"
 import OSCALControl from "./OSCALControl";
 import OSCALControlLabel from "./OSCALControlLabel";
 import OSCALControlPart from "./OSCALControlPart";
-import { Accordion, AccordionDetails, AccordionSummary } from "./StyedAccordion";
+import { Accordion, AccordionSummary } from "./StyledAccordion";
```
kylelaker, line 20: Good catch on the rename here. I saw it the other day and it drove me crazy

packages/oscal-react-library/src/components/OSCALCatalogGroup.tsx
```diff
@@ -154,7 +155,7 @@ const CollapsibleListItem: React.FC<CollapsibleListItemProps> = (props) => {
       }}
     >
       <AccordionSummary>{itemText}</AccordionSummary>
-      <AccordionDetails>{children}</AccordionDetails>
+      <MuiAccordionDetails className="AccordionDetails">{children}</MuiAccordionDetails>
```
kylelaker, line 158: I don't love this... I don't like that three very tightly coupled components get used very differently. Even if we end up just having:

```ts
export const AccordionDetails = (props) => <MuiAccordionDetails {...props} className="AccordionDetails" />;
```

I like that better because it gives consistent usage between them. Not saying I _love_ that idea either but I prefer consistent usage.

packages/oscal-react-library/src/components/StyledAccordion.tsx
```diff
@@ -0,0 +1,12 @@
+import ArrowForwardSharp from "@mui/icons-material/ArrowForwardIosSharp";
+import MuiAccordion from "@mui/material/Accordion";
+import MuiAccordionSummary from "@mui/material/AccordionSummary";
```
kylelaker, line 3:
```suggestion
import MuiAccordion, { AccordionProps} from "@mui/material/Accordion";
import MuiAccordionSummary, { AccordionSummaryProps} from "@mui/material/AccordionSummary";
```

packages/oscal-viewer/src/App.js
Bronstrom: I think this is a good idea, will go ahead and make it a dependency in the project through this method.

packages/oscal-react-library/src/components/StyledAccordion.tsx
```diff
@@ -0,0 +1,12 @@
...
+import React from "react";
+
+export const Accordion = (props: any) => (
+  <MuiAccordion className="Accordion" disableGutters elevation={0} square {...props} />
+);
+
+export const AccordionSummary = (props: any) => (
+  <MuiAccordionSummary className="AccordionSummary" expandIcon={<ArrowForwardSharp />} {...props} />
+);
```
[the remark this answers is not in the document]
Bronstrom: I want to ask a clarifying question here. Are you asking for me to consider the difference between the following?:
```
export const Accordion = (props: any) => (
  <MuiAccordion {...props} className="Accordion" disableGutters elevation={0} square />
);
```

and

```
export const Accordion = (props: any) => (
  <MuiAccordion className="Accordion" disableGutters elevation={0} square {...props} />
);
```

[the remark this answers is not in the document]
Bronstrom: Receiving an error when attempting to add `as typeof MuiAccordion`:

```
Conversion of type 'Element' to type 'OverridableComponent<AccordionTypeMap<{}, "div">>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.ts(2352)
```

I don't think we need to include the `typeof` for this to be valid TypeScript. Is providing the type for `props` not enough?

packages/oscal-viewer/src/GlobalStyles.jsx
```diff
@@ -0,0 +1,107 @@
+import React from "react";
```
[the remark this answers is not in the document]
Bronstrom: The reason for this was to easily align these files in our other OSCAL repositories following the same styling structure. Some of our other repositories don't currently support TypeScript, and it seemed easier to start with JS. Furthermore, the POF was also written in JS.

Although, before this PR is marked "Ready for Review" these files will be in TypeScript.

Bronstrom:
I want to note that any files that are written in JS in this PR will be converted to TS before it is marked "Ready for Review". Mainly the reason for this is to better share progress here while at the same time keep styling in other repositories aligned.

I have created a `styles` folder as had been discussed in the ADR and started to place the more complex styled components there, grouping them under a common name. Instead of using the term "Styled<Component name>", which the ADR says to stray away from, I'm thinking they could be name "OSCAL<Component name>" instead. This follows the naming scheme with our other component files, however being in the "styles" folder will separate them from there more complex component counterparts. I'm thinking another reason for giving these files this "OSCAL…" prefix is to separate them from MUI components, allowing us to easily identify our own components.

Bronstrom:
Within `App.js` you can now toggle a view of the Style guide components on a single page by setting the flag `isStyleMode` to `true`.

image

packages/oscal-react-library/src/components/styles/OSCALButtons.js
mpemy, line 1: @Bronstrom Excellent job!
Just a small suggestion when defining  styled components  as this one
export const OSCALTertiaryGrayscaleButton = (props) => {
  return (
    <Button
      {...props}
      className="DestructiveButton"
      variant="contained"
      variant="text"
      disableElevation
      sx={{
        backgroundColor: "#B31515", //(theme) => theme.palette.destructive.main,
        paddingX: "0.25rem",
        marginX: 0,
        color: (theme) => theme.palette.black.main,
        ":hover": {
          backgroundColor: "#9B1616", //(theme) => theme.palette.destructive.dark,
          backgroundColor: (theme) => theme.palette.gray.main,
        },
      }}
    />
it is preferable to use
const OSCALTertiaryGrayscaleButton = styled(Button)`
...
`;
This will allow the user of the library to adjust the style of the component according. This is more relevant when one wants to adjust the size of styled components.
This applies to all similar components.

Bronstrom: Sounds good @mpemy, I can understand the difficulties in trying to adjust these components due to there being only one `sx` allowed for a component. Ideally, we don't tweak these components too much, but I can understand the desire to want the flexibility to tweak them.

I would say to make a new component if they are more major visually styling changes - like color or sizing if it is going to be reused throughout the application. For example, if we need a button that looks like an `OSCALTertiaryButton` and is used for deletes (uses red coloring over the standard blue), we probably want to add an `OSCALTertiaryDestructiveButton` to the `OSCALButtons` file. However, it is minor tweaks where we will make one or two small changes to one of these styling components, I can see the desire we want to still have the `sx` property available to the developer when importing one of these components.

I'll go ahead and follow the approach suggested and shift places where I used `sx` in these styling files to `styled(<componentName>)...`

mpemy: @Bronstrom Thanks!

Bronstrom: After going through this approach, some of the styling needed to be pulled out from the actual components. It appears that some of the properties being passed to styling components weren't being accounted for when attempting to only use `styled(<componentName>)` for the component in the PKI tool (`{…props}` and some other attributes are needed to keep this functioning correctly). I've ended up more opting for the approach of using `styled(<componentName>)` as a helper method for the component that is being styled. I've decided to keep these "helper" `const`s private and call these as `OSCALHelper<componentName>`. Definitely open for other suggestions on this process or the naming of these "helpers" 😆

I've also left some of the `sx` properties in these styling files for helper related methods, or areas where MUI components are stacked within each other to make a single styled component - this shouldn't have an effect on giving the `sx` property to the styled component.

PR #849, opened 2023-05-09
feat(control): uniform display for control parts

kylelaker:
This makes it so that Discussion, Assessment Methods, and Objectives are
laid out similarly. Overviews and Statements always appear at the top
(in that order); all other parts are grouped within the drop downs. This
gates them behind a click (Discussion already was) but it makes the
controls a bit easier to read without a wall of Assessment Objective
text.

Component Definition: 
image

Catalog: 
image


Bronstrom (commented):
This is looking good so far! I think this is a major improvement over the previous design by breaking up the expansive text. I think having sections to divide the content up is especially helpful. I would suggest that we have this content opened at the start though to reduce the user having to click several accordions to view a control in full, but I still like them having the freedom to close the accordions when they desire.

packages/oscal-react-library/src/components/OSCALControlPart.js
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { styled } from "@mui/material/styles";
-import OSCALControlGuidance from "./OSCALControlGuidance";
+// import OSCALControlGuidance from "./OSCALControlGuidance";
```
Bronstrom, line 3: This line can just be removed.

kylelaker:
> I would suggest that we have this content opened at the start though to reduce the user having to click several accordions to view a control in full, but I still like them having the freedom to close the accordions when they desire.

I agree; however, that doesn't solve the wall-of-text issue that this PR is trying to solve. We previously hid some portion of this behind a click (Discussion) and the 800-53A content (or assessment-focused content) is less likely to be the desired text for a user (usually you want the overview, statement, and possibly discussion).

Because we plan to do a larger evaluation of the design for controls soon, I'd like to keep this as-is for now and we can see about auto-expanding in the follow-on designs.

Bronstrom (approved):
> Because we plan to do a larger evaluation of the design for controls soon, I'd like to keep this as-is for now and we can see about auto-expanding in the follow-on designs.

That's fair. Looks good to me!

PR #853, opened 2023-05-10
feat: create hoverable popover

### LINKED ISSUE #592: Support OSCAL Markup (Line and Multi-Line) in Hover-Over (Tooltips-like functionality)

tuckerzp:
This creates a  [popover](https://mui.com/material-ui/react-popover/) component that is displayed on hover over. The main component is intentionally fairly generic. You can imagine many different components you would want to hover over and see text for. 

There are also two wrapper components created for the common use cases today of displaying `MarkupLine` and `MarkupMultiLine` in a popover. This has been implemented in `OSCALSystemImplementationComponents` as an example. 

closes #592 

### Testing

- The default SSP and our old default SSP contain components with a description to display on hover. 

---

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/main/system-security-plans/ssp-example.json

image


packages/oscal-react-library/src/components/HoverablePopover.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { OSCALMarkupLine, OSCALMarkupMultiLine } from "./OSCALMarkupProse";
+
+export interface HoverablePopoverProps {
+  /**
+   * The content of the popover.
+   */
+  popoverContent: React.ReactNode;
+  /**
+   * Content to to hover over. This must fit within a span.
+   */
```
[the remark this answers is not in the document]
tuckerzp: Fair enough. I had thought it worth mentioning that it is placed within a `span` since I couldn't think of a better way of restricting it. I will remove that part of the comment.

packages/oscal-react-library/src/components/HoverablePopover.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const markup = <OSCALMarkupMultiLine>{popoverContent}</OSCALMarkupMultiLine>;
+
+  return <HoverablePopover popoverContent={markup}>{children}</HoverablePopover>;
+};
+
+export const MarkupLinePopover: React.FC<MarkupPopoverProps> = ({ popoverContent, children }) => {
+  const markup = <OSCALMarkupLine>{popoverContent}</OSCALMarkupLine>;
+
+  return <HoverablePopover popoverContent={markup}>{children}</HoverablePopover>;
+};
```
kylelaker, line 90: I don't understand how/why this gives much benefit over the callers doing this themselves. It feels like an awkward API. The benefit is a `string` instead of a `ReactNode`? But if they need to pass additional props, they have to fallback to `HoverablePopover` anyway?

packages/oscal-react-library/src/test-data/SystemData.ts
```diff
@@ -58,13 +59,13 @@ export const usersTestData = [
...
 
-export const componentsTestData = [
+export const componentsTestData: AssessmentAssetsComponent[] = [
   {
     title: "Example Component",
     uuid: "component-1",
     description: "An example component.",
     status: {
-      state: "operational",
+      state: PurpleState.OPERATIONAL,
```
kylelaker, line 68: Use the alias we export from `index.ts`

packages/oscal-react-library/src/components/HoverablePopover.tsx
tuckerzp: I don't _love_ having to pass a component as a prop and this seemed like such a common use case (maybe the only use case) that the wrapper felt beneficial. 

I had considered making `HoverablePopover` private and just using those wrappers since that was the main case the component would be used.

I can remove it if we feel strongly they don't provide much value.

Bronstrom: I would almost suggest just checking the `popover` type to see if it's a `string` and format it differently with `OSCALMarkupLine`. This would avoid the need to create a second `interface` and `const`.

tuckerzp: > I would almost suggest just checking the `popover` type to see if it's a `string` and format it differently with `OSCALMarkupLine`. This would avoid the need to create a second `interface` and `const`.

If a string is being dispalyed, it is not always going to be `markup`. Even if it was, it may be a `markup line` or a `markup multiline`.

Bronstrom: Sorry, I understand now. Maybe I'm just not understanding all the cases of how a multiline will be provided, but couldn't we just check the `string` to have a newline character in it to determine if it should be `markup multiline`?

tuckerzp: `"a *cool* string"` is valid both as markup line and markup multi line. There are places in the spec that very specifically state which we should choose. So I don't think we can do that as nice as it would be. 

I think the issue Kyle is mentioning is the inclusion of those wrapper components at all. As you _can_ just put the correct mark up component in the prop.

kylelaker (approved):

PR #855, opened 2023-05-11
fix: large control display "squishes" tabs for Component controls

kylelaker:
Historically, our display of controls for components has been a bit
fragile, where if the control side got too large, it would squish into
the component tabs and cause overflow, truncation, or wrapping. This is
suboptimal and instead we should use a behavior more like what we do in
OSCALControlGroups where we set a specific side on Grid items. The same
numbers are used in both places.

Because of the complexity around handling fragments, I did not rework
this into a shared component. No other styles for the component have
changes to align it with Controls (though we should do that).

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALControlImplementationImplReq.js
+++ b/packages/oscal-react-library/src/components/OSCALControlImplementationImplReq.js
@@ -6,6 +6,7 @@ import PropTypes from "prop-types";
 import Tabs from "@mui/material/Tabs";
 import Tab from "@mui/material/Tab";
 import Box from "@mui/material/Box";
+import Grid from "@mui/material/Grid";
 import OSCALControl from "./OSCALControl";
 import getControlOrSubControl from "./oscal-utils/OSCALControlResolver";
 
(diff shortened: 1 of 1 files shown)


packages/oscal-react-library/src/components/OSCALControlImplementationImplReq.js
```diff
@@ -92,7 +93,7 @@ function a11yProps(index) {
  * Creates the internal elements of Control Implementation.
  *
  * @param {object} props SSP properties
- * @returns The corresponding Control Implementation Request
+ * @returns The corresponding Control Implementation Reques
```
kylelaker, line 96:
```suggestion
 * @returns The corresponding Control Implementation
```

packages/oscal-react-library/src/components/OSCALControlImplementationImplReq.js
```diff
@@ -111,41 +112,46 @@ export default function OSCALControlImplementationImplReq(props) {
...
+          </Grid>
+          <Grid item lg={9.5} sm={12}>
             {Object.values(props.components).map((component, index) => (
-              <ComponentTabButton
-                label={component.title}
-                {...a11yProps(index)}
-                key={component.uuid}
-              />
+              <ComponentTabPanelScrollable value={value} index={index} key={component.uuid}>
+                <p>What?!</p>
```
kylelaker, line 136:
```suggestion
```

tuckerzp (approved):

Issue #82, opened 2021-07-02
Display Control Modifications According to Position Value


rgauss:
As a user of the OSCAL viewer or underlying React components, I want to see the indicator of modifications displayed in a way that reflects the [position value](https://pages.nist.gov/OSCAL/reference/latest/profile/json-reference/#/profile/modify/alters/adds/position) that has been defined for them so that I can understand how they apply to the control.

### Acceptance Criteria
* The visual indicator of modifications is displayed in a position relative to the control or `id-ref` that reflects the defined position value for that modification

brian-easyd:
It sounds like of the four possibilities, we are only accomplishing "after" at this point and need to accomplish the other three.
- **before**: Preceding the id-ref target
- **after**: Following the id-ref target
- **starting**: Inside the control or id-ref target, at the start
- **ending**: Inside the control or id-ref target, at the end

PR #856, opened 2023-05-12
feat: display control modification icon by position

mpemy:
Closes: #82


packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -241,9 +241,139 @@ const OSCALControl: React.FC<OSCALControlProps> = (props) => {
...
 
+  const position = props?.modificationAlters
+    ?.find((item) => item["control-id"] === props.control.id)
+    ?.adds?.flatMap((item) => item["position"])
+    .filter((part) => part)[0];
+
   const label = propWithName(control.props, "label")?.value;
   const controlOrParentWithdrawn = withdrawn || isWithdrawn(control);
 
+  if (position === "after") {
```
kylelaker, line 252: Can you add `import { Position } from "@easydynamics/oscal-types";` to the top of this file and then use the enum variants in these conditions (`if (position === Position.AFTER) {`)?

I do think we need to find a slightly less-repetitive way to handle the code within the conditions but I am not quite sure how to go about that refactoring. I think @tuckerzp is pretty familiar with this so maybe he can provide an opinion

mpemy: Thanks @kylelaker for this suggestion, I will do that.

tuckerzp: > a slightly less-repetitive way to handle the code within the conditions

Briefly looking at this, the only difference I see here for each if is the position of `modifcationDisplay` in the component being created.

At the very least, could we not do something like:

```TypeScript
<OSCALControlCard>
  <position1>
    {position === Position.AFTER) ? modificationDisplay : null}
  </position1>
  <position2>
    {position === Position.Before) ? modificationDisplay : null}
  </position2>
</OSCALControlCard>
```

I don't think we need to create a completely different declaration of the component if we are just switching 1 thing around based on an if statement.

mpemy: Thanks @tuckerzp, I refactored the code. 
This is the profile I am using to test the implementation. https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/test-content/profiles/NIST_SP-800_rev4TestProfile.json
These are few screenshots
image

image

packages/oscal-react-library/src/components/OSCALControlPart.js
```diff
@@ -44,35 +50,73 @@ export default function OSCALControlPart(props) {
...
-        componentId={props.componentId}
-        modificationSetParameters={props.modificationSetParameters}
-        modificationDisplay={modificationDisplay}
-        isEditable={props.isEditable}
-        onRestSuccess={props.onRestSuccess}
-        onRestError={props.onRestError}
-        partialRestData={props.partialRestData}
-      />
+      <Typography>
+        {modificationDisplayBefore}
```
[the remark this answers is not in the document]
mpemy: @tuckerzp I tried that approach first, but you can see that when I defined modificationDisplayBefore, I have to turn off all the other modificationDisplay so I have more than one instruction to be executed, that is why I set all other modificationDisplay to undefined. I first try to set them to null, but I got some linting issues.

tuckerzp:
Looking at the screenshot now, I also am wondering about the position of the modification icon. Why are they now on a completely new line? 

For example, `ac-2_smt.a` has modifications on the line above. I wonder if it would look better like.

```
a. {icon} identifies...
````

It may be helpful to describe where you think modifications should appear for each `position` value. It can even be a  quick code block like the one above.

File: packages/oscal-react-library/src/components/OSCALControl.tsx [hunk 3]
```diff
@@ -241,9 +242,50 @@ const OSCALControl: React.FC<OSCALControlProps> = (props) => {
...
+
+  switch (position) {
+    case Position.AFTER:
+      // eslint-disable-next-line no-lone-blocks
+      {
+        modificationDisplayBefore = undefined;
+        modificationDisplayEnding = undefined;
+        modificationDisplayStarting = undefined;
+      }
+      break;
```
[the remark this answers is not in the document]
mpemy: I kept the switch block and added the break clause within the curly brackets.

packages/oscal-react-library/src/components/OSCALControl.tsx
```diff
@@ -241,9 +242,44 @@ const OSCALControl: React.FC<OSCALControlProps> = (props) => {
...
+    ?.adds?.flatMap((item) => item["position"])
+    .filter((part) => part)[0];
+
   const label = propWithName(control.props, "label")?.value;
   const controlOrParentWithdrawn = withdrawn || isWithdrawn(control);
 
+  let modificationDisplayBefore = modificationDisplay;
+  let modificationDisplayAfter = modificationDisplay;
+  let modificationDisplayEnding = modificationDisplay;
+  let modificationDisplayStarting = modificationDisplay;
```
tuckerzp, line 256: > but you can see that when I defined modificationDisplayBefore, I have to turn off all the other modificationDisplay

```suggestion
  const modificationDisplayBefore = position === Position.BEFORE ? modificationDisplay : null;
  const modificationDisplayAfter = position === Position.AFTER ? modificationDisplay : null;;
  const modificationDisplayEnding = position === Position.ENDING ? modificationDisplay : null;;
  const modificationDisplayStarting = position === Position.STARTING ? modificationDisplay : null;;
```

I am pretty sure this is equivelent to what you are doing right? If `position == BEFORE`, all other the other variables will be `null`. You also could just do this check everywhere you put these variables, but that is up to personal preference. 

This makes it so we do not have to use `let` and it makes the code a lot clearer. I really think there must be a cleaner way to deal with this other than the `switch`.

packages/oscal-react-library/src/components/OSCALControlPart.js
```diff
@@ -117,7 +155,31 @@ export default function OSCALControlPart(props) {
...
+            modificationAlters={props.modificationAlters}
+            modificationSetParameters={props.modificationSetParameters}
+            key={part.id ?? part.name}
+            isEditable={props.isEditable}
+            onRestSuccess={props.onRestSuccess}
+            onRestError={props.onRestError}
+            partialRestData={props.partialRestData}
+          />
+        ))}
+        {modificationDisplay}
```
tuckerzp, line 179: Why is there not a `modificationDisplayEnding` in this file?

tuckerzp (changes requested):

packages/oscal-react-library/src/components/OSCALControlPart.js
mpemy: This block of codes in under the condition that position == Position.ENDING, that is why at first, I did not use modificationDisplayEnding there. But for code clarity, I just replaced modificationDisplay by modificationDisplayEnding.
Thanks!

tuckerzp:
I made a (hopefully) simpler emaple Profile to test with. Feel free to make any changes you feel you'd want.

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/profile-mods/profiles/test-profile.json

image

---

@mpemy is this loading like intended? It is still a bit unclear to me where you intend modifications to be for each position.

mpemy:
@tuckerzp I have addressed the positioning issue. The implementation follows acceptance criteria: 
before: Preceding the id-ref target
after: Following the id-ref target
starting: Inside the control or id-ref target, at the start
ending: Inside the control or id-ref target, at the end.

Thanks!

mpemy:
@tuckerzp  This is a screenshot using your profile.
image

kylelaker (approved):
Thanks for your work on this and for collaborating with @tuckerzp. I think this is a clean implementation that meets the ticket. We might need to more largely re-evaluate how modifications are handled in the future.

mpemy:
You are welcome @kylelaker.

tuckerzp (approved):

PR #858, opened 2023-05-12
feat: display params in catalogs, groups, controls

### LINKED ISSUE #822: Support viewing parameters for catalogs, groups & controls
### Description

As a user of the OSCAL Viewer/Editor, I would like to be able to view `params `for catalogs, groups, and controls.

### Acceptance Criteria

* [`params`](https://pages.nist.gov/OSCAL/reference/1.0.5/complete/json-reference/#/catalog/controls/params) are properly displayed at the catalog, group, and control levels
  * This includes presentation of the list of parameters, to include: `label`, `usage`, `constraint`, `guideline`, `value`/`select`, and `remarks`

### Proposed Solution

Ideally params is made into it's own component that is passed to catalogs, groups, and controls to be displayed.

### Additional Notes

_No response_

### Implementation

tuckerzp:
Creates a dialog box with accordions to display parameters. In theory, there can be quite a bit of information put here so accordions were chosen over using a table like we do for properties. 

closes #822 

### Testing

Based on example catalogs today, often most fields are undefined and only a few things are actually displayed. This example shows a "worst" case parameter definition. 

https://raw.githubusercontent.com/EasyDynamics/oscal-demo-content/tuckerzp/params/catalogs/basic-test-catalog.json

---

image

---

image

image

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALParam.tsx
+++ b/packages/oscal-react-library/src/components/OSCALParam.tsx
@@ -6,6 +6,7 @@ import { OSCALPropertiesDialog } from "./OSCALProperties";
 import { Stack } from "@mui/system";
 import { Accordion, AccordionDetails, AccordionSummary } from "./StyedAccordion";
 import { ButtonLaunchedDialog } from "./ButtonLaunchedDialog";
+import { OSCALMarkupMultiLine } from "./OSCALMarkupProse";
 
 export interface OSCALParamProps {
   param: Parameter;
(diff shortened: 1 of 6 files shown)


packages/oscal-react-library/src/components/OSCALParam.tsx
```diff
@@ -0,0 +1,267 @@
...
+interface OSCALSelectProps {
+  choice?: string[];
+}
+
+const OSCALSingleSelect: React.FC<OSCALSelectProps> = ({ choice }) => {
+  const [selectValue, setSelectValue] = React.useState("");
+
+  const handleChange = (event: SelectChangeEvent) => {
+    setSelectValue(event.target.value as string);
+  };
```
tuckerzp, line 158: I initially wanted `OSCALSingleSelect` and `OSCALMultiSelect` to be a single `OSCALSelect` that has a `multiple` flag.

This led to a lot of ternary nonsense that I could not quite get to work. If someone has a cleaner way of dealing with a single component, I will be happy to update this.

packages/oscal-react-library/src/components/OSCALParam.tsx
```diff
@@ -0,0 +1,267 @@
...
+      {links?.map((link) => (
+        <Link key={link.href} href={link.href}>
+          {link.text ?? link.href}
+        </Link>
+      ))}
+    </>
+  );
+};
+
+interface OSCALSelectProps {
```
tuckerzp, line 149: For [select](https://pages.nist.gov/OSCAL/reference/latest/catalog/json-reference/#/catalog/params/select), I thought it made the most sense to actually select things with the [MUI Select component](https://mui.com/material-ui/react-select/).

I realized that there is not really a way to do anything with that selection other than displaying it currently, but it seemed better than a table of what you could select.

tuckerzp:
Marking this as `Ready for Review` so that I can get comments on the design here.

kylelaker (approved):
This looks good. I am not worried about the duplication in the select element right now. I think this is smoother. I like that this helps Props and Params work well together.

Bronstrom (approved):
Nice work, looks good!

PR #859, opened 2023-05-12
fix: rendering components before loading

tuckerzp:
OSCAL documents were being rendered before loading was complete.

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALLoader.tsx
+++ b/packages/oscal-react-library/src/components/OSCALLoader.tsx
@@ -64,7 +64,7 @@ export function getRequestedUrl() {
   return new URLSearchParams(window.location.search).get("url");
 }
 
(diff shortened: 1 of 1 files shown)


packages/oscal-react-library/src/components/OSCALLoader.tsx
```diff
@@ -323,6 +323,7 @@ export default function OSCALLoader(props: OSCALLoaderProps): ReactElement {
       </Grid>
     );
   } else if (oscalUrl) {
+    console.log("WTF?!", JSON.stringify(oscalData));
```
kylelaker, line 326:
```suggestion
```

kylelaker (approved):
I hate this fix but I guess it makes sense. We really ought to look into better loading handling.

Bronstrom (approved):

Issue #111, opened 2021-07-08
Display Back Matter media types in a human-readable format


kylelaker:
As a user involved in the governance, risk, and compliance of systems and components exploring controls, I want the media type of the back matter resources to be displayed in a human readable format when possible.

### Acceptance Criteria

 - Back matter media types have human-readable names for the common types (OSCAL documents, PDFs, HTML document, etc)
 - If the media type doesn't match to something well-known, the exact value from the `media-type` field is displayed
 - If the media type isn't specified, whichever heuristic is used to determine a media type results in a human readable name (such as using the file extension)
 - Product Owner agrees with the list of recognized types.
 
### Implementation Notes

 - Per [the specification](https://pages.nist.gov/OSCAL/reference/latest/complete/json-reference/#/catalog/back-matter/resources/rlinks/media-type), this is supposed to be an IANA registered media type; however, it seems that some things don't comply with this
   - Our [default catalog](https://raw.githubusercontent.com/usnistgov/oscal-content/master/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_catalog.json) seems to not specify `media-type` for many documents; however we seem to detect many of them as PDFs or "Unknown"
   - Some of the media types, such as `application/oscal.content+json` is not formally registered as an IANA media type

brian-easyd:
Find an (the?) authoritative source for media types and recognize those.
Enumerate the official list of recognized types and associate with human-friendly names. Consider icons (not required)

brian-easyd:
Core list plus OSCAL extensions.

kylelaker:
@brian-easyd We have an awkward issue with this because of the _very_ small number of back matter resources in existing documents that specify a `media-type`. So in most cases, the viewer today falls back to `Unknown` or a guess based off the file extension.

## Determining Media Type

```mermaid
graph TD
    A([Displaying Media Type]) --> B{Is `media-type` present?}
    B -->|Present| C
    B -->|Absent| D[Use file extension to guess media type]
    D --> E{Can media type\n be determine from\nextension}
    E -->|Media Type Found| C[Convert to friendly-name\nusing lookup]
    E -->|Not Found| Fallback([Display 'Unknown Content'])
    C --> Z([Display friendly name for\nmedia-type])
```

## Friendly names for MIME types

The IANA registrations do not necessarily contain a friendly name. This lists how we will represent specific media types and classes of media types in the viewer.

### `image/`
| Name | Display in Viewer |
[...]

PR #861, opened 2023-05-15
feat: human-readable display for back matter media types

laurelmay:
This updates how we display resources in Back Matter to try to use a more human-readable name. When `media-type` is provided, we prefer that; otherwise we fallback to guessing based on the file extension using the `mime-db` package's database (which includes registrations with IANA as well as the Nginx and Apache databases).

The media types are largely those defined in #111 though with some additional ones found during the implementation that may be useful. The lookup algorithm is not significantly modified besides the feedback given in the issue comments.

The UI is also updated slightly to give more visual indication that it's a button and can be clicked to open the link.

Closes: #111

### Screenshots

General Display:
image

OSCAL Media Types:
image


packages/oscal-react-library/src/components/oscal-utils/OSCALLinkUtils.ts
```diff
@@ -56,6 +57,10 @@ function getFileExtension(url: string): string {
   }
 }
 
+export function getFallbackMediaTypeFromExtension(link: ResourceLink): string | false {
+  return mimeTypeLookup(link.href);
+}
+
```
tuckerzp, line 63: I assume we want to get rid of this?

```suggestion
```

kylelaker: Oh yeah I have totally rewrite that.

kylelaker:
Todo:

- [ ] Better document how the lookups are performed
- [ ] Write tests for the new code

tuckerzp (commented):
This is looking really good so far. After seeing your implementation, I now agree `mime-db` is the better choice here.

packages/oscal-react-library/src/components/oscal-utils/OSCALMediaTypeUtils.ts
```diff
@@ -0,0 +1,334 @@
...
+        case "bmp":
+          return "Image (BMP)";
+        case "webp":
+          return "Image (WebP)";
+        case "x-icon":
+          return "Image (Icon)";
+        case "vnd.microsoft.icon":
+          return "Image (Icon)";
+        case "vnd.adobe.photoshop":
+          return "Image (Photoshop)";
```
Bronstrom, line 246: I have a refactoring question/suggestion related to capturing the type and subtype, which I like as `type (subtype)`. I'm noticing you're writing the type a lot without much change for example here. Would it be a lot of extra work to capture the type first (could be captured a variable or even a separate helper function with a switch statement) and then provide the subtype? I understand there are some cases where you want to only provide the subtype or suffix to be returned so this might add some complication.

Bronstrom (commented):
This is looking good so far @kylelaker! I'm liking the choices you've made with capturing and formatting the media types are great and I think "Unrecognized" is a good fallback for when the media type isn't found.

packages/oscal-react-library/src/components/oscal-utils/OSCALMediaTypeUtils.ts
kylelaker: I don't understand what you're asking for... Unfortunately, we can't just assume that we can programatically pull a useful name consistently. Consider SVG or Photoshop above. This code might be repetitive but it is explicit and it is correct. There won't be subtle bugs where _some_ media types display incorrectly.

Bronstrom: What I'm asking is instead of writing the type ("Image"/"Audio"/"Video"/etc.) before each subtype, capturing this elsewhere and concatenating this with the subtype.

> Unfortunately, we can't just assume that we can programatically pull a useful name consistently. 

Although, I think this answers my question.

kylelaker: Even if we do something like the following, it will be wrong more often that it's worth:

function toTitleCase(s: string): string {
  return s.charAt(0).toUpperCase() + s.slice(1);
}
function normalizeSubtype(m: MediaType): string {
  // Guess how the fuck the media type might look and hope to god we guess
  // correctly when it comes to brand names and that they don't capitalize
  // the whole thing or a word part-way through.
  // We could maintain a list of exceptions but is that really any better
  // than where we are now?
  if (m.subtype.startsWith("vnd.")) {
    const lastBit = s.split(".").pop();
    return toTitleCase(lastBit);
  }
  // Hopefully this file type is all uppercase! We're not maintaining any
  // overrides. So we'll definitely get MP3 wrong.
  return m.subtype.split(".").pop().toUpperCase();
}
function display(m: MediaType): string {
  return toTitleCase(m.type) + " " + "(" + normalizeSubtype(m) + ")";
}
```

Comment (kylelaker) [reply packages/oscal-react-library/src/components/oscal-utils/OSCALMediaTypeUtils.ts:246]: Even if we do something like the following, it will be wrong more often that it's worth:

```
function toTitleCase(s: string): string {
  return s.charAt(0).toUpperCase() + s.slice(1);
}
function normalizeSubtype(m: MediaType): string {
  // Guess how the media type might look and hope we guess
  // correctly when it comes to brand names and that they don't capitalize
  // the whole thing or a word part-way through.
  // We could maintain a list of exceptions but is that really any better
  // than where we are now?
  if (m.subtype.startsWith("vnd.")) {
    const lastBit = s.split(".").pop();
    return toTitleCase(lastBit);
  }
  // Hopefully this file type is all uppercase! We're not maintaining any
  // overrides. So we'll definitely get MP3 wrong.
  return m.subtype.split(".").pop().toUpperCase();
}
function display(m: MediaType): string {
  return toTitleCase(m.type) + " " + "(" + normalizeSubtype(m) + ")";
}
```

Bronstrom (approved):
Thanks for your work on this @kylelaker! I appreciate the documentation providing more reasoning on the less clear parts of the code.

tuckerzp (approved):

PR #877, opened 2023-05-24
chore: implement hoverable popover

### LINKED ISSUE #873: Implement hoverable popover
### Description

#592 added a hover-able popover to replace where we put `OSCAL Markup` inside a `ToolTip`. We need to actually find and replace everywhere we do that. 

### Acceptance Criteria

- Places where we put `OSCALMarkup` inside a `ToolTip` are identified & replaced

### Proposed Solution

_No response_

### Additional Notes

There may be a few places where we should be displaying `OSCAL Markup` and instead use a `Typography`, those places should also be identified and updated.  

### Implementation

tuckerzp:
`StyledTooltip` is unable to have Markup prose as a title. 
This replaces everywhere we use a styled tooltip that should 
have Markup prose to use a HoverablePopover.

closes #873 

Continuation of #853


packages/oscal-react-library/src/components/HoverablePopover.tsx
```diff
@@ -26,6 +26,29 @@ export const HoverablePopover: React.FC<HoverablePopoverProps> = ({ popoverConte
...
+      transformOrigin={{
+        vertical: "top",
+        horizontal: "left",
+      }}
+      onClose={handlePopoverClose}
+      disableRestoreFocus
+    >
+      <Box sx={{ padding: "1em" }}>{popoverContent}</Box>
+    </Popover>
+  ) : null;
```
tuckerzp, line 50: I was thinking about cases in which  `popoverContent` is not defined and was unsure of the best functionality. I would think that the popover should not appear if there is no content to display.

kylelaker: Then maybe to simplify the whole thing...

```ts
if (!popoverContent) {
  return (<>{children}</>);
}
```

kylelaker: But also it's remarkably unlikely `popoverContent` will be falsey since it's like to be React component with a falsey child. So in any case, _this_ doesn't feel like the right place to do the check?

tuckerzp: Yeah that is the issue I am running into. I had originally thought it would be falsy if the `MarkupProse` returned `null`. I may just add a check like:

```TypeScript
<HoverablePopover popoverContent={item && <MarkupProse>{item}</MarkupProse>} />
```

mpemy (commented):
@tuckerzp, please post some screenshots so we can see how this looks like.

tuckerzp:
> see how this looks like.

@mpemy this is a continuation of #853. The component has not changed how it looks

Bronstrom (changes requested):
Overall this is looking good. I'm noticing two more places where `StyledTooltip` should be replaced with `HoverablePopover`. Once in `OSCALBackmatter` and the other in `OSCALControlProse` where a `StyledTootip` is passed `statementByComponentDescriptionMarkup` as the title.

Bronstrom (approved):

mpemy (commented):
It looks OK.

kylelaker (approved):

kylelaker:
Thank you @tuckerzp for the work and @Bronstrom for supporting getting this over the finish line.

PR #878, opened 2023-05-26
chore: removed profile part unit test

mpemy:
Closes https://github.com/EasyDynamics/easygrc/issues/180, Closes https://github.com/EasyDynamics/easygrc/issues/181


packages/oscal-react-library/src/components/OSCALControlPart.test.js
```diff
@@ -0,0 +1,42 @@
...
+  controlProseDecimalTestData,
+  exampleParams,
+  exampleDecimalParams,
+} from "../test-data/ControlsData";
+import { sspRestData } from "../test-data/SystemData";
+
+const labelTestData = "label-1";
+const labelTestDataDecimal = "label-1.1";
+
+describe("OSCALControlProse", () => {
```
tuckerzp, line 25: I think this test already exists? 

https://github.com/EasyDynamics/oscal-react-library/blob/64fed6bdc4b7592790479698f02faf98aeba62ad/packages/oscal-react-library/src/components/OSCALControlProse.test.js#L24

mpemy: Yeah, this test already exists, I am modifying this test so that when the part is removed the prose should appear with strike-through line.

tuckerzp: Hm ok. Maybe we should change the `describe` then?

In general, we have been moving to components testing their own functionality (rather than the parent or other components). So `OSCALControlPart` tests render an `OSCALControlPart` and test only its functionality and `OSCALControlProse` tests render an `OSCALControlProse` and test only its functionality. 

I realize this is a draft, but I want to make sure to be on the same page before we go too far in writing tests. It is confusing when we make a change to a very low level component and tests begin to fail in it's parent. In theory, a component should be usable on it's own.

mpemy: You are right, as I said, I adjusted that unit test to test control part functionalities. I just based my initial implementation on tests that have similar functionalities, given that I was not able to run test locally, I first used a test that should passed, as an initial starting point.

Bronstrom (commented):
@mpemy this is looking great! A couple structural things stood out at me, and I've commented below about them.

packages/oscal-react-library/src/components/OSCALProfile.test.js
```diff
@@ -1,37 +1,41 @@
...
+  profileWithIncludeAllAndExcludeAC3,
+} from "../test-data/ProfileData";
 
 describe("The OSCAL Profile", () => {
   it("displays controls", async () => {
     render(
       <OSCALProfile
         profile={profileTestData}
-        parentUrl={parentUrlTestData}
+        parentUrl={urlSmallCatalogUrl}
```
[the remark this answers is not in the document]
kylelaker: I am fine with 100,000. What I would like us to do though is define a constant around line 9-11.

```ts
const PROFILE_RESOLUTION_TEST_TIMEOUT_MS = 100_000;
```

Then I'd like for all the places where we bumped the `timeout` to be `timeout: PROFILE_RESOLUTION_TEST_TIMEOUT_MS`

kylelaker:
I agree overall with @Bronstrom's review and will provide approval once those items are addressed.

packages/oscal-react-library/src/components/OSCALProfile.test.js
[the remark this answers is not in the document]
mpemy: Thanks @Bronstrom for your suggestions, I will decrease the timeouts.

packages/oscal-react-library/src/components/OSCALProfile.test.js
```diff
@@ -41,19 +45,128 @@ describe("The OSCAL Profile", () => {
...
+
+    const resultAC2 = await screen.findByText("AC-2", {
+      timeout: 100000,
+    });
+    const resultAC1 = await screen.findByText("AC-1", {
+      timeout: 100000,
+    });
+
+    expect(resultAC2).toBeVisible();
+    expect(resultAC1).toBeVisible();
```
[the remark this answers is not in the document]
mpemy: Thanks for your suggestions, I will implement them.

Bronstrom (approved):
Nice work @mpemy! This looks good to me.

PR #883, opened 2023-06-08
chore(tests): always pass `onResolutionComplete` to documents

kylelaker:
We have a bunch of odd failures because `onResolutionComplete` is
required but we don't pass it in the tests. This (hopefully) fixes that.
CDef's tests got reworked in this because it helped to identify the
issue but I did more minimal changes in the other locations.

### DIFF
--- a/packages/oscal-react-library/src/components/OSCALComponentDefinition.test.js
+++ b/packages/oscal-react-library/src/components/OSCALComponentDefinition.test.js
@@ -9,22 +9,33 @@ test("OSCALComponentDefinition loads", () => {
   render(<OSCALComponentLoader />);
 });
 
-function componentDefinitionRenderer() {
-  render(<OSCALComponentDefinition componentDefinition={componentDefinitionTestData} />);
-}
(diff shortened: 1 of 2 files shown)


Bronstrom (approved):

packages/oscal-react-library/src/components/OSCALComponentDefinition.test.js
```diff
@@ -9,22 +9,33 @@ test("OSCALComponentDefinition loads", () => {
...
+      />
+    );
 
-function testOSCALComponentDefinitionComponent(parentElementName, renderer) {
-  test(`${parentElementName} shows component title`, () => {
-    renderer();
-    const result = screen.getByText("Example Component");
-    expect(result).toBeVisible();
+    // THEN
+    expect(screen.getByText("Example Content")).toBeVisible();
```
kylelaker, line 23:
```suggestion
    expect(screen.getByText("Example Component")).toBeVisible();
```

Bronstrom (approved):

Issue #885, opened 2023-06-08
Update navigation menus


Bronstrom:
### Description

As a consumer of the Editor, I would like to view a visual of how the navigation will appear.

### Acceptance Criteria

* In the Editor, an `AppBar` appears at the top of the window, including:
  * A search bar
  * An `Avatar` for the user and place for the user's name
  * Notification and Settings buttons
* In the Editor, a Permanent Drawer component displays on the left side of the window, including:
  * The company logo in the top left corner
  * A hierarchical navigation tab menu
  * "Logout" and "Support" tabs

### Proposed Solution

The implementation should fall in line with the style guide and mock-ups made for the tool.

### Additional Notes

This implementation does not need link/routing support and should be purely handled on the front-end.

### Implementation

- [ ] I can (or plan to) submit a pull request to implement this
- [ ] Implementing this may result in a breaking change

Bronstrom:
Capturing the Drawer component and vertical menu in a more concrete issue: https://github.com/EasyDynamics/oscal-react-library/issues/885. The design has also been decided in mock-ups.

brian-easyd:
@Bronstrom the mock-ups are for the editor. We have not considered changes to the viewer in our UI design work. Please revise this to be about the editor.

Bronstrom:
Sounds good, will swap out the word "Viewer" for "Editor" in the issue description.

Bronstrom:
6/22/23 Sprint Planning - Moved from Sprint 49 to Sprint 50

Bronstrom:
7/6/23 Sprint Planning - Moved from Sprint 50 to Sprint 51

Bronstrom:
We decided to add a dev flag to the application to toggle development features on and off, which currently is just the navigation menus in this issue. In order to view this feature, the Editor must be running with `isDevMode` set to `true` in `packages/oscal-viewer/src/App.js`.

PR #901, opened 2023-06-16
feat: add feature flags

kylelaker:
This adds basic support for feature flags. Only two-state (true/false)
feature flags are enabled where `true` enables the feature and `false`
disables. Usage of the components would look a bit like this:

```tsx
<FeatureFlagsProvider flags={{ isEditingEnabled: true, isEnterpriseSupportEnabled: false }}>
  <SomeParentThing>
    <Feature flag="isEditingEnabled">
      <MyEditingComponent />
    </Feature>
  </SomeParentThing>
</FeatureFlagsProvider>
```

Specifically in our application, `FeatureFlagsProvider` would probably
be defined in the `App` component and then `Feature` used throughout the
library. An `App` may define additional features; however, it's
important that the _library_ only use features defined in
`FeatureFlags.tsx`.

### DIFF
--- a/packages/oscal-react-library/src/components/Feature.tsx
+++ b/packages/oscal-react-library/src/components/Feature.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { FeatureFlagConfiguration, FeatureFlags } from "../contexts/FeatureFlags";
+
+export type FeatureFlagKey<T extends FeatureFlagConfiguration> = keyof T;
+
+export interface FeatureProps<T extends FeatureFlagConfiguration = FeatureFlagConfiguration> {
+  feature: FeatureFlagKey<T>;
+  children: React.ReactNode;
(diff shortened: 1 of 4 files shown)


Bronstrom:
> Specifically in our application, `FeatureFlagsProvider` would probably be defined in the `App` component and then `Feature` used throughout the library. An `App` may define additional features; however, it's important that the _library_ only use features defined in `FeatureFlags.tsx`.

This approach makes the most sense to me. Would we want to have `FeatureFlagsProvider` placed in `App` now or wait till we have something better to test it out on? https://github.com/EasyDynamics/oscal-react-library/issues/885 would probably be a good test for feature flags, as we had discussed only having some items display in the menu based on the flags being set.

mpemy (approved):
This implementation looks fine.

PR #904, opened 2023-06-21
feat: update navigation menus

Bronstrom:
This adds a more updated Drawer component and `AppBar` to the application implemented in TypeScript. This also adds svg icons within an images folder.

## Testing
In order to view the drawers within the application, the Editor must be running with `isDevMode` enabled (set to `true`) within `packages/oscal-viewer/src/App.js`.

Closes https://github.com/EasyDynamics/oscal-react-library/issues/885


Bronstrom:
I've been working to convert these files currently to TypeScript and have run into a couple common errors I'm working to resolve - it seems TypeScript is not a fan of custom colors (`primaryAccent`) as well as currently will not resolve our svg icons.

The other issue I've run is that I can't seem to export components from `index.ts` in oscal-react-library to be used in `App.js` oscal-viewer. I'm wondering if this is because `rollup` doesn't work the same way with `lerna` - would you have any clue on this one @kylelaker?  I believe `npm start` would automatically run `rollup` previously.

Bronstrom:
Worked to resolve the above problems by working out build failures with module augmentation. There are still a few more styling improvements that need to be made, but the AppBar and Permanent Drawer are functioning similar to what is described in the issue. For now, I've made a "flag" constant `isDevMode` which can be just toggled on and off in the code. When `isDevMode` is `true`, the new navigation is show, otherwise the Drawer Selector is shown. I also have the navigation tree view from the `OSCALDrawerSelector` also displaying within the Permanent Drawer. This is a temporary feature for testing purposes - so we can still view OSCAL documents with the new Permanent Drawer and AppBar displaying on the top and left sides.

mpemy (approved):
@Bronstrom, you did an excellent job. Thanks!

PR #920, opened 2023-08-07
Hide navigation dev features on `viewer.oscal.io`

Bronstrom:


mpemy (approved):
Thanks @Bronstrom for your prompt implementation of a fix for this bug.

PR #927, opened 2023-08-17
manage-group

mpemy:


packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    let sibling = orphans[0];
+    for (let i = 1; i < orphans.length; i++) {
+      orphans[i].rightSibling = sibling;
+      sibling = orphans[i];
+    }
+  }
+
+  console.log("fin groups after parents :", groupsAndSubs);
+  console.log("ordered groups  after parents", orderGroups);
+  console.log("orphans", orphans);
```
Bronstrom, line 202: Before making this final, can we remove all occurrences of `console.logs` in this PR that is used just for testing? I'm okay with keeping the success and failure messages for now.

packages/oscal-react-library/src/components/OSCALCatalogBaseline.tsx
```diff
@@ -36,19 +36,21 @@ import { FormatBold, FormatItalic, FormatQuote, Subscript, Superscript } from "@
...
 
+import GroupDrawer from "./OSCALCatalogManageGroup";
+
 const Hug = styled(Container)`
   position: absolute;
   width: 348px;
   height: 36px;
   top: 120px;
-  left: 810px;
+  left: 840px;
```
Bronstrom, line 46: You don't have to make the change here, but in the future can we start using `rem` instead of `px` for our style sizing? Emily made this choice in the styleguide and I would like to follow this method. This is useful if we were to ever change our default font size, everything else will scale correctly.

`1rem` is equivalent to our font size and our font size is set to `16px`. I mostly try to think in divisible of `4px` when sizing (what is the unit of measure Emily mentioned in the grid design of the application). Here's a quick idea of how `px` sizes translate to `rem`:
- `4px` - `0.25rem`
- `8px` - `0.5rem`
- `12px` - `0.75rem`
- `16px` - `1rem`
- `24px` - `1.5rem`
- `32px` - `2rem`
- `64px` - `4rem`

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+      <Container
+        onClick={handleNewGroupClick}
+        sx={{ height: 73, border: "1px solid #D2D2D2", background: "#EAEAEA", width: 312 }}
+      >
+        <Container
+          sx={{
+            height: 73,
+            left: 20,
+            position: "absolute",
+            background: "#F6F6F6",
```
Bronstrom, line 464: This is `backgroundGray` in the theme

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+            border: "1px solid #D2D2D2",
```
Bronstrom, line 465: I would suggest adding `#D2D2D2` and any new occuring colors to our palette theming (in `AppTheme`) if this they are a commonly occurring color (using this color more than 2 times).

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+                onChange={handleEditIDChange}
+                sx={{
+                  left: 32,
+                  height: 35,
+                  width: 52,
+                  top: 5,
+                  position: "absolute",
+                  fontSize: 10,
+                  fontWeight: 200,
+                  background: "#ffffff",
```
Bronstrom, line 487: This is `white` in the palette theming

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+                  left: 88,
+                  top: 5,
+                  position: "absolute",
+                  height: 32,
+                  width: 198,
+                  background: "#ffffff",
+                  "& .MuiOutlinedInput-root": {
+                    "& > fieldset": {
+                      borderColor: (theme: { palette: { secondary: { main: any } } }) =>
+                        theme.palette.secondary.main,
```
Bronstrom, line 511:
```suggestion
                      borderColor: (theme) => theme.palette.secondary.main,
```

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+          }}
+        ></Box>
+      </Container>
+    );
+  };
+  const RootLevel: React.FC = () => {
+    function handleRootLevelClick() {
+      setSelectedItemName("Root");
+    }
+    const rootSelected = selectedItemName === "Root" ? true : false;
```
Bronstrom, line 602: I don't think the ternary operator is needed here. I think `const rootSelected = selectedItemName === "Root"` will return the right value.

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    };
+    function addNewGroupSuccess(response: any) {
+      console.log("successful addition of a new group", response);
+    }
+    function addNewGroupFail(e: any) {
+      console.log("Fail to create a new group", e.statusText);
+    }
+    fetchTransaction("/edit_group_title", request_json, addNewGroupSuccess, addNewGroupFail);
+  }
+  function DeleteGroup(ID: string) {
```
Bronstrom, line 219: To keep our function naming aligned I would just use camel casing here: `deleteGroup`.

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
```diff
@@ -0,0 +1,1052 @@
...
+            Delete this Group?
+        </Typography> */}
+
+        {/* <OSCALTextField
+          label={text}
+          id={"delete"}
+          onChange={}
+          small
+          sx={{ width: 180 }}
+        ></OSCALTextField> */}
```
Bronstrom, line 347: I would suggest deleting this unused code unless you think this will be used in a future iteration of catalog groups. If so, I would add a `TODO` comment and specify when this will be added.

packages/oscal-react-library/src/components/images/icons/delete.svg:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
+ZoneId=3
+HostUrl=https://www.figma.com/
```
Bronstrom, line 3: For all custom `svg` files - we should just be grabbing the `svg`s directly (with their path code) and placing them into this images folder rather than importing this through a link on figma.

packages/oscal-react-library/src/components/styles/OSCALButtons.js
```diff
@@ -87,6 +87,26 @@ export const OSCALSmallSecondaryButton = (props) => {
...
+      sx={{
+        height: 20,
+        ":hover": {
+          backgroundColor: "#023E88",
+          color: "#FFFFFF",
+        },
+      }}
+    />
+    //backgroundColor: (theme) => theme.palette.secondary.main,
+    //color: (theme) => theme.palette.white.main,
```
Bronstrom, line 106: This commented code can be removed

Bronstrom (commented):
@mpemy, this is looking great so far! I have some review comments with suggestions for code tidying, making the styling process in the future simpler, as well as a note on importing new custom `svg`s.

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
mpemy: @Bronstrom Thanks for the reminder. I removed all debug console.logs

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
mpemy: @Bronstrom, when I define a new theme color in AppTheme.js I ran into an error when I compile the application.
image
Typescipt do not recognize the newly added theme colors. I do not know if these new colors have to define in another file besides AppTheme.js

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
mpemy: The same happened when I directly use a palette already defined 
image
It looks like typescript needs to extra directive to accept certain palettes defined in AppTheme.js

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
mpemy: You are right. Thanks for pointing that out. Thanks!

packages/oscal-react-library/src/components/images/icons/delete.svg:Zone.Identifier
mpemy: Done.

packages/oscal-react-library/src/components/styles/OSCALButtons.js
mpemy: Done.

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
Bronstrom: Oh, I completely forgot. Everytime we define a new `theme` in `AppTheme.js` TypeScript requires us to also define this as a type option for the palette. I created a file called `createPalatte.d.ts` which currently has `backgroundGray` and `primaryAccent` defined currently. We will need to add `lightGray` to the two interfaces there.

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
Bronstrom: You're correct on this. You can follow the same method as I had mentioned with the `lightGray` palette option above.

packages/oscal-react-library/src/components/OSCALCatalogManageGroup.tsx
Bronstrom: Was this issue able to be resolved without needing to update `createPalatte.d.ts`?

mpemy: @Bronstrom, Thanks for this suggestion. I added few more palettes to cover all colors we have in the application so far.

Bronstrom:
I appreciate the changes you have made here @mpemy. This PR and #944 touch a few of the same files. Once #944 is merged, let's make sure this merges cleanly with main.

Bronstrom:
Closing PR due to `comply0-open` now handling new work.

Issue #888, opened 2023-06-09
Upload OSCAL Files


brian-easyd:
### Description

As a security practitioner I want to upload existing OSCAL files into the tool so I can view and edit them.

The upload capability needs to send the file to the back end for validation and placement into the datastore.

The user interface should offer both a drag-and-drop capability as well as a file selection capability.

### Acceptance Criteria

- Aligned with the Spring 2023 styling strategy, guidance, and standards
- A valid OSCAL file is dropped into the drop zone, and ends up within the server's datastore in accordance with the datastore specification. 
- A valid OSCAL file is selected, and ends up within the server's datastore in accordance with the datastore specification. 
- An invalid OSCAL file is dropped into the drop zone and is rejected as invalid

### Proposed Solution

- file validation and placement into the datastore should be handled by the back end.
- An OTSL transaction and helper may be necessary 

### Implementation

- [ ] I can (or plan to) submit a pull request to implement this
- [ ] Implementing this may result in a breaking change

emcathow:
No Catalogs/Baselines Uploaded: 
https://www.figma.com/file/ntIfIHIhMYbiqS9Nuc7EQG/Catalog-Mockups?type=design&node-id=8-977&t=iqrjELBQeVi4uCKQ-0 

Upload: 
https://www.figma.com/file/ntIfIHIhMYbiqS9Nuc7EQG/Catalog-Mockups?type=design&node-id=70-702&t=iqrjELBQeVi4uCKQ-0

Uploading: 
https://www.figma.com/file/ntIfIHIhMYbiqS9Nuc7EQG/Catalog-Mockups?type=design&node-id=70-703&t=iqrjELBQeVi4uCKQ-0

Success: 
https://www.figma.com/file/ntIfIHIhMYbiqS9Nuc7EQG/Catalog-Mockups?type=design&node-id=70-704&t=iqrjELBQeVi4uCKQ-0

Fail: 
https://www.figma.com/file/ntIfIHIhMYbiqS9Nuc7EQG/Catalog-Mockups?type=design&node-id=70-705&t=iqrjELBQeVi4uCKQ-0

Bronstrom:
6/22/23 Sprint Planning - Moved from Sprint 49 to Sprint 50

Bronstrom:
7/6/23 Sprint Planning - Moved from Sprint 50 to Sprint 51

Bronstrom:
[...]

PR #944, opened 2023-09-15
feat (catalog): Support uploading OSCAL files

mpemy:
Closes https://github.com/EasyDynamics/oscal-react-library/issues/888


packages/oscal-react-library/src/components/OSCALCatalogBaseline.tsx
```diff
@@ -466,7 +420,11 @@ export default function OSCALCatalogBaseline() {
   let contact: ContactInfo = {};
   address = authorAddress;
   const createdModel = newOSCALModel === undefined ? Data : newOSCALModel;
-  const LabelText = openCatalogBaseline ? createdModel.title : " Catalogs & Baselines";
+  const LabelText = openCatalogBaseline
+    ? createdModel.title
+    : uploadNewCatalogBaseline
+    ? " Upload a Catalog or Baseline "
+    : " Catalogs & Baselines";
```
Bronstrom, line 427: I'm having a little difficulties reading this. To keep the code clear and concise, I would suggest breaking this into two constants, each handling one of the ternary operations.

packages/oscal-react-library/src/components/OSCALCatalogBaseline.tsx
```diff
@@ -520,6 +478,472 @@ export default function OSCALCatalogBaseline() {
...
+        {upload && (
+          <Box
+            sx={{
+              top: 180,
+              left: 40,
+              height: 400,
+              width: 1030,
+              background: "#ffffff",
+              position: "absolute",
+              border: "1xp solid #0028675E",
```
Bronstrom, line 812: Just wanted to just mention it once more, in the future we should try to keep the units of measure to `rem`, where we are sizing/positioning components.

packages/oscal-react-library/src/components/OSCALCatalogBaseline.tsx
```diff
@@ -1154,10 +1624,58 @@ export default function OSCALCatalogBaseline() {
...
-              </Select>
+                <option value="AL">Alabama</option>
+                <option value="AK">Alaska</option>
+                <option value="AZ">Arizona</option>
+                <option value="AR">Arkansas</option>
+                <option value="CA">California</option>
+                <option value="CO">Colorado</option>
+                <option value="CT">Connecticut</option>
+                <option value="DE">Delaware</option>
+                <option value="DC">District of Columbia</option>
```
Bronstrom, line 1635: Is this the same `Select` field and options used in #927? I think we could refactor this code and reuse it, once one of these PRs are merged in.

packages/oscal-react-library/src/components/OSCALLoader.tsx
```diff
@@ -28,6 +29,7 @@ import {
 import { ReactElement } from "react-markdown/lib/react-markdown";
 import { AnchorLinkProps } from "./OSCALAnchorLinkHeader";
 
+const isDevMode = false; // This boolean is used to switch to dev mode
```
Bronstrom, line 32: Do we need a second `isDevMode` flag in the loader? Is there an issue with just using the one in the App? Maybe it can be placed in a context or be passed down as `props` to the loader.

packages/oscal-react-library/src/components/images/icons/fileUpload.svg:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
+ZoneId=3
+HostUrl=https://www.figma.com/
```
Bronstrom, line 3: Same comment here on the `svg`s as #927: we want these to be `svg`s with path code instead of figma links.

packages/oscal-react-library/src/components/OSCALCatalogBaseline.tsx
```diff
@@ -520,6 +478,472 @@ export default function OSCALCatalogBaseline() {
...
+            ":hover": {
+              color: "#ff6600",
+            },
+            fontFamily: "Source Sans Pro",
+          }}
+        >
+          CATALOGS & BASELINES
+        </Typography>
+      </OSCALBreadCrumbs>
+    );
```
Bronstrom, line 622: I'm trying to pick out the differences between these three conditions here, because it looks like a lot of reusable code. Just an idea, but can you just validate these conditions when attempting to render parts of this code? For example, with the first condition it looks like you're trying to figure out whether to include this breadcrumb item:
```
        {(item?.model.title?.length > 0) &&
          <CatalogTooltip title={usedTitle.toUpperCase()}>
            <Typography
              color="primary"
              sx={{
                fontWeight: (theme) => theme.typography.fontWeightBold,
                textTransform: "uppercase",
                ":hover": {
                  color: "#ff6600",
                },
                fontFamily: "Source Sans Pro",
              }}
            >
              {trunckatedTitle.toUpperCase()}
            </Typography>
          </CatalogTooltip>
         }

```

packages/oscal-react-library/src/components/OSCALCatalogBaseline.tsx
mpemy: @Bronstrom Thanks for the insight, I broke it down to ternary operations.

mpemy:
@Bronstrom, I addressed code review comments and suggestions. As for the flag isDevMode, I tried to add it as a global constant but the issue it creates is that the package oscal-viewer should be set as a dependency for the other packages. But those packages are already dependencies for oscal-viewer.
I also tried OSCALContext but it did not work. Given that we will soon have a dev branch, we will not have this issue once that branch is established.

Bronstrom (commented):
Thanks for working hard on the palette updates! One small change before we merge this PR in:

packages/oscal-react-library/src/components/OSCALCatalogBaseline.tsx
```diff
@@ -1259,9 +1934,12 @@ export default function OSCALCatalogBaseline() {
                 </StepperBar>
               </Grid>
               <Grid item xs={12}>
-                <AddInformationLabel>
+                {/* <AddInformationLabel>
                   Add information about the {Model.toLowerCase()}:
-                </AddInformationLabel>
+                </AddInformationLabel> */}
```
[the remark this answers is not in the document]
mpemy: @Bronstrom, Sorry I just forgot to delete that line before checking the code.
I just addressed this issue. Please also review the PR feat (catalog): Manage catalog groups #927. I also made similar changes on styling on Manage Group.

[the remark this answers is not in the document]
Bronstrom: No worries, appreciate you addressing this quickly!

Bronstrom (approved):

PR #960, opened 2023-11-15
fix(loader): Adjust `OSCALLoaderForm` button styling

Bronstrom:
This PR adjusts the layout of the text field and buttons in the loader form. The buttons are now made a consistent size and the field and buttons will span the whole screen depending on the window change. Below you can see examples of how the loader form looks at various window widths:

image

image

image


brian-easyd (approved):
LGTM

